Issue #45, opened 2017-06-30
Deal with send_all for ReplyAction


ericnewcomer:
..or come up with a migration story for this crazy feature

Relevant RP test: `temba.flows.FlowsTest.test_send_all_replies`

rowanseymour:
Closed via #49

PR #49, opened 2017-07-07
Send to all urns

ericnewcomer:
Addresses #45

### DIFF
--- a/cmd/flowrunner/testdata/flows/all_actions.json
+++ b/cmd/flowrunner/testdata/flows/all_actions.json
@@ -55,7 +55,7 @@
                             "twitter:benhaggerty"
                         ],
                         "all_urns": true,
-                        "text": "Hi @contact.name, are you ready?"
+                        "text": "I'm a message to both of @contact.name's urns."
                     },
                     {
                         "uuid": "6d1346c0-48d8-4108-9c58-e45a1eb0ff7a",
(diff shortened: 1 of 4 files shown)


flows/actions/reply.go
```diff
@@ -51,6 +54,15 @@ func (a *ReplyAction) Execute(run flows.FlowRun, step flows.Step) error {
 		return nil
 	}
 
-	run.AddEvent(step, events.NewSendMsgToContact(run.Contact().UUID(), text, a.Attachments))
+	urns := run.Contact().URNs()
+	if a.AllURNs && len(urns) > 0 {
+		for i := range urns {
```
nicpottier: This may be a question of how we expect this to work for contacts that have no URNs but I think the more go-ing way of doing this would probably be:

```go
if a.AllURNS {
  for _, urn := range run.Contact().URNS() {
     run.AddEvent(step, events.NewSendMsgToURN(urn, text, a.Attachments))
  }
} else {
...
}
```

cmd/flowrunner/testdata/flows/all_actions.json
```diff
@@ -48,6 +48,16 @@
                         "text": "Hi @contact.name, are you ready?"
                     },
                     {
+                        "uuid": "ac110f56-a66c-4462-921c-b2c6d1c6dadb",
+                        "type": "send_msg",
```
nicpottier: I don't think this means what you think it means. :)

PR #52, opened 2017-07-10
Update localization to cascade thru contact languages

rowanseymour:
Gets rid FlowRun.flowTranslations altogether because we can't pick a single language translation for the run - that translation might not have a specific text, so need to always cascade thru all the language translations.

### DIFF
--- a/cmd/flowserver/flow.go
+++ b/cmd/flowserver/flow.go
@@ -44,10 +44,11 @@ func (r *flowResponse) MarshalJSON() ([]byte, error) {
 }
 
 type startRequest struct {
-	Flows   json.RawMessage      `json:"flows"    validate:"required"`
-	Contact json.RawMessage      `json:"contact"  validate:"required"`
-	Extra   json.RawMessage      `json:"extra,omitempty"`
-	Input   *utils.TypedEnvelope `json:"input"`
+	Environment *json.RawMessage     `json:"env"`
+	Flows       json.RawMessage      `json:"flows"    validate:"required"`
+	Contact     json.RawMessage      `json:"contact"  validate:"required"`
+	Extra       json.RawMessage      `json:"extra,omitempty"`
+	Input       *utils.TypedEnvelope `json:"input"`
 }
 
 func handleStart(w http.ResponseWriter, r *http.Request) (interface{}, error) {
@@ -67,6 +68,16 @@ func handleStart(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 	err = utils.ValidateAll(start)
 	if err != nil {
 		return nil, err
+	}
+
+	var env utils.Environment
+	if start.Environment != nil {
+		env, err = utils.ReadEnvironment(start.Environment)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		env = utils.NewDefaultEnvironment()
 	}
 
 	// read our flows
@@ -93,11 +104,11 @@ func handleStart(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 		}
 	}
 
-	// build our environment
-	env := engine.NewFlowEnvironment(utils.NewDefaultEnvironment(), startFlows, []flows.FlowRun{}, []*flows.Contact{contact})
+	// build our flow environment
+	flowEnv := engine.NewFlowEnvironment(env, startFlows, []flows.FlowRun{}, []*flows.Contact{contact})
 
 	// start our flow
-	session, err := engine.StartFlow(env, startFlows[0], contact, nil, input, start.Extra)
+	session, err := engine.StartFlow(flowEnv, startFlows[0], contact, nil, input, start.Extra)
 	if err != nil {
 		return nil, fmt.Errorf("error starting flow: %s", err)
 	}
@@ -106,10 +117,11 @@ func handleStart(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 }
 
 type resumeRequest struct {
-	Flows   json.RawMessage      `json:"flows"       validate:"required,min=1"`
-	Contact json.RawMessage      `json:"contact"     validate:"required"`
-	Session json.RawMessage      `json:"session"     validate:"required"`
-	Event   *utils.TypedEnvelope `json:"event"       validate:"required"`
+	Environment *json.RawMessage     `json:"env"`
+	Flows       json.RawMessage      `json:"flows"       validate:"required,min=1"`
+	Contact     json.RawMessage      `json:"contact"     validate:"required"`
+	Session     json.RawMessage      `json:"session"     validate:"required"`
+	Event       *utils.TypedEnvelope `json:"event"       validate:"required"`
 }
 
 func handleResume(w http.ResponseWriter, r *http.Request) (interface{}, error) {
@@ -129,6 +141,16 @@ func handleResume(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 	err = utils.ValidateAll(resume)
 	if err != nil {
 		return nil, err
+	}
+
+	var env utils.Environment
+	if resume.Environment != nil {
+		env, err = utils.ReadEnvironment(resume.Environment)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		env = utils.NewDefaultEnvironment()
 	}
 
 	// read our flows
@@ -165,11 +187,11 @@ func handleResume(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 	}
 
 	// build our environment
-	env := engine.NewFlowEnvironment(utils.NewDefaultEnvironment(), flowList, session.Runs(), []*flows.Contact{contact})
+	flowEnv := engine.NewFlowEnvironment(env, flowList, session.Runs(), []*flows.Contact{contact})
 
 	// hydrate all our runs
 	for _, run := range session.Runs() {
-		err = run.Hydrate(env)
+		err = run.Hydrate(flowEnv)
 		if err != nil {
 			return nil, utils.NewValidationError(err.Error())
 		}
@@ -182,7 +204,7 @@ func handleResume(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 	}
 
 	// resume our flow
-	session, err = engine.ResumeFlow(env, activeRun, event)
+	session, err = engine.ResumeFlow(flowEnv, activeRun, event)
 	if err != nil {
 		return nil, fmt.Errorf("error resuming flow: %s", err)
 	}
--- a/cmd/flowserver/flow.go
+++ b/cmd/flowserver/flow.go
@@ -44,7 +44,7 @@ func (r *flowResponse) MarshalJSON() ([]byte, error) {
 }
 
 type startRequest struct {
-	Environment *json.RawMessage     `json:"env"`
+	Environment *json.RawMessage     `json:"environment"`
 	Flows       json.RawMessage      `json:"flows"    validate:"required"`
 	Contact     json.RawMessage      `json:"contact"  validate:"required"`
 	Extra       json.RawMessage      `json:"extra,omitempty"`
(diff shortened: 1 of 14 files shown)


flows/actions/save_contact_field.go
```diff
@@ -55,26 +55,7 @@ func (a *SaveContactField) Execute(run flows.FlowRun, step flows.Step) error {
 		run.AddError(step, err)
 	}
 
-	// if this is either name or language, we save directly to the contact
```
rowanseymour: This is unrelated - am assuming this stuff isn't meant to be here since update_contact is the action that updates language

flows/definition/localization.go
```diff
@@ -11,35 +11,21 @@ type itemTranslations map[string][]string
 // languageTranslations map a node uuid to item_translations - say "node1-asdf" to { "text": "je suis francais!" }
 type languageTranslations map[flows.UUID]itemTranslations
 
-func (t *languageTranslations) GetTranslations(uuid flows.UUID, key string, backdown []string) []string {
+func (t *languageTranslations) GetTextArray(uuid flows.UUID, key string) ([]string, bool) {
```
rowanseymour: "translations" felt a bit overloaded

flows/engine/engine.go
```diff
@@ -95,20 +97,6 @@ func ResumeFlow(env flows.FlowEnvironment, run flows.FlowRun, event flows.Event)
 	return run.Session(), nil
 }
 
-// initializes our context based on our flow and current context
-func initTranslations(run flows.FlowRun) {
-	// set our language based on our contact if we have one
-	contact := run.Contact()
-	if contact != nil {
```
rowanseymour: I'm not seeing where else we support runs with no contact so am assuming this isn't actually possible

flows/runs/run.go
```diff
@@ -197,33 +196,29 @@ func (r *flowRun) ExpiresOn() *time.Time       { return r.expiresOn }
...
+	return textArray[0]
 }
 
-func (r *flowRun) GetTranslations(uuid flows.UUID, key string, backdown []string) []string {
-	if r.translations == nil {
-		return backdown
+func (r *flowRun) GetTextArray(uuid flows.UUID, key string, native []string) []string {
+	for _, lang := range r.languages {
+		if lang == r.Flow().Language() {
+			return native
```
rowanseymour: I've renamed `backdown` to `native` because the native text in the flow will often be the preferred language - i.e. not a backdown.

flows/actions/save_contact_field.go
nicpottier: Good catch, ya, we moved things over to `update_contact` and never removed it.

flows/definition/localization.go
nicpottier: Should add comment here as to what the return params are

nicpottier: Is it useful to have a distinction between `nil` here? (which I'm assuming is the purpose of the bool return value?) Just wondering when a caller would ever act differently when receiving a nil value and true vs a nil value and false. (if never, then lets just return a single value and nil when not found)

flows/engine/engine.go
nicpottier: I don't think there's anything in the engine that currently requires a contact, at least I've tried to make that not a requirement. Seems that might be a nice thing to maintain no?

flows/contact.go
```diff
@@ -11,17 +11,32 @@ import (
...
-	name     string
-	language utils.Language
-	timezone *time.Location
-	urns     URNList
-	groups   GroupList
-	fields   *Fields
+	uuid           ContactUUID
+	name           string
+	language       utils.Language
+	otherLanguages utils.LanguageList
```
nicpottier: Can you talk about why you went for two fields here instead of just one list?

flows/engine/engine.go
rowanseymour: What is the usecase for a run without a contact? If there's a good usecase then it's worth it. Otherwise it's a nuisance to have to keep checking `run.Contact() != nil` and from a glance it looks like we don't always do that, e.g. https://github.com/nyaruka/goflow/blob/master/flows/engine/engine.go#L14

flows/contact.go
rowanseymour: I started off with one list this morning (https://github.com/nyaruka/rapidpro/pull/1347) and then switch it to two because one list doesn't let a user determine whether a contact has a language stored, i.e. we need to differentiate between the explicitly set language and implicit backdown languages.

flows/definition/localization.go
rowanseymour: I guess as long as we never request translations of empty things... then no. Can change

flows/engine/engine.go
nicpottier: Flows sure seem like they could be useful even without a contact no?

flows/contact.go
nicpottier: Ah right.. ok, good call. `otherLanguages` seems like a punt on a name though.. would `backdownLanguages` make more sense? `envLanguages`?

flows/contact.go
```diff
@@ -11,17 +11,32 @@ import (
...
+
+// PreferredLanguages gets all languages for this contact in order of preference
+func (c *Contact) PreferredLanguages() utils.LanguageList {
+	var languages utils.LanguageList
+
+	if c.language != utils.NilLanguage {
+		languages = append(languages, c.language)
+	}
+
+	return append(languages, c.otherLanguages...)
```
nicpottier: Do we expect any kind of contract that languages won't be repeated here? Seems that could happen through a `setLanguage` within a flow.

flows/engine/engine.go
rowanseymour: I guess so. I'll restore these checks and fix the other ones

flows/contact.go
rowanseymour: Renamed to `backdownLanguages` tho I'm not in love with that

flows/contact.go
rowanseymour: RapidPro is responsible for not duplicating between contact.language and contact.backdown_languages, and I've updated this to only include flow language if it's not already included

flows/engine/engine.go
```diff
@@ -55,8 +53,6 @@ func ResumeFlow(env flows.FlowEnvironment, run flows.FlowRun, event flows.Event)
 		return run.Session(), nil
 	}
 
-	initTranslations(run)
```
rowanseymour: I've moved this initialization into flowrun as it seems like it belongs in there

rowanseymour:
@nicpottier please review

flows/runs/run.go
```diff
@@ -197,33 +207,40 @@ func (r *flowRun) ExpiresOn() *time.Time       { return r.expiresOn }
...
 
-func (r *flowRun) updateTranslations() {
-	if r.flowTranslations != nil {
-		r.translations = r.flowTranslations.GetTranslations(r.language)
-	} else {
-		r.translations = nil
+func (r *flowRun) SetLanguages(langs utils.LanguageList) {
+	r.languages = langs
+
+	// if languge list doesn't include the flow's native language, add that last
```
nicpottier: typo

flows/contact.go
```diff
@@ -11,17 +11,32 @@ import (
...
+// PreferredLanguages gets all languages for this contact in order of preference
+func (c *Contact) PreferredLanguages() utils.LanguageList {
+	var languages utils.LanguageList
+
+	if c.language != utils.NilLanguage {
+		languages = append(languages, c.language)
+	}
+
+	return append(languages, c.backdownLanguages...)
+}
```
nicpottier: Say a contact starts with `'eng'` as a language, but backdown languages is set to `['fra', 'spa']`, if an update contact action setting the language to 'spa' happens, won't preferred languages return `['spa', 'fra', 'spa']`?

rowanseymour: Come to think of it, RapidPro shouldn't remove duplicates from backdown. If a user has language 'eng' and backdown 'eng', we should provide both of those to goflow, because even if the contact language changes, we still need to know that 'eng' is a valid backdown language.

So I think...
  * contact.language and contact.backdownLanguages are allowed to overlap
  * contact.PreferredLanguages should return with duplicates removed
  * run.SetLanguages shouldn't add flow language if it's already included

nicpottier: I think you are right about all the above.

But maybe the reason we are getting confused here is that `backdownLanguages` really isn't an attribute of the contact, but rather an attribute of the environment (or perhaps even the flow).

Maybe these attributes really belong in some kind of environment that is  passed to the flow? (and maybe that's where channels live too?)

rowanseymour: Yeah and if you want to run flows without contacts then that's definitely necessary. How about...

```
 "env": {"timezone": "UTC", "languages": ["eng", "spa"]}
```

What would be the argument for having channels inside env instead of at the top level?

cc @ericnewcomer

nicpottier: I like TZ on contact though, but ya that would be the backdown for a null TZ on a contact, which is more correct.

To be clear, I was thinking maybe the channels available would be passed in `env`, not the preferred channel for a URN or anything. (those would just be channel uuids?) That kinda feels like the environment that the flow is running in no? Other things that might be in env:
 * resthooks?
 * transfer to integrations?
 * admin boundary stuff?
 * email account settings?

rowanseymour: Yeah was assuming `env.timezone` was a backdown. I could do channel list in env or toplevel. They feel like dependencies in the same way dependent flows are.

Admin boundaries might warrant being their own microservice if we don't want to send all of the Nigeria in our flow server requests.

Why would goflow need email settings? Won't it just be sending back send_email events?

nicpottier: Ya, I think you are right on the email side, we don't need that at the goflow level.

For whatever reason it feels cleaner to me to have a required `env` parameter to the flow engine which as optional fields than having lots of optional fields, but realize that's personal. Env is pretty overloaded in goflow, but I can't think of anything better nor am I sure it is inappropriate where it is used.

rowanseymour:
@nicpottier @ericnewcomer updated so that backdown languages are passed in via `env`, and this is read into a `utils.Environment`. To handle contact overriding of timezone and languages, we have `runEnvironment` which implements `flows.FlowEnvironment` but overrides `Languages()` and `Timezone()`

cmd/flowserver/flow.go
```diff
@@ -44,10 +44,11 @@ func (r *flowResponse) MarshalJSON() ([]byte, error) {
 }
 
 type startRequest struct {
-	Flows   json.RawMessage      `json:"flows"    validate:"required"`
-	Contact json.RawMessage      `json:"contact"  validate:"required"`
-	Extra   json.RawMessage      `json:"extra,omitempty"`
-	Input   *utils.TypedEnvelope `json:"input"`
+	Environment *json.RawMessage     `json:"env"`
```
nicpottier: More of a discussion point but do we want to go down the line of abbreviating some things or keep everything verbose? My bias would be keep things verbose apart from acronyms but what do you guys think? @ericnewcomer

flows/runs/run.go
```diff
@@ -235,15 +276,16 @@ func NewRun(env flows.FlowEnvironment, flow flows.Flow, contact *flows.Contact,
...
-		environment: env,
 		contact:     contact,
 		flow:        flow,
 		status:      flows.StatusActive,
 		createdOn:   now,
 		modifiedOn:  now,
 	}
 
-	// create our new context
+	r.environment = newRunEnvironment(env, r)
```
nicpottier: Are there any implications with run environments being stacked here? IE, subflows will be creating runs with an environment that is a runEnvironment.

utils/environment.go
```diff
@@ -52,36 +55,42 @@ func (e *environment) SetTimezone(timezone *time.Location) {
...
 //------------------------------------------------------------------------------------------
 // JSON Encoding / Decoding
 //------------------------------------------------------------------------------------------
 
 type envEnvelope struct {
-	DateFormat DateFormat `json:"date_format"`
-	TimeFormat TimeFormat `json:"time_format"`
-	Timezone   string     `json:"timezone"`
+	DateFormat DateFormat   `json:"date_format"`
+	TimeFormat TimeFormat   `json:"time_format"`
```
nicpottier: Forgot about these guys, which we were going to have pass into the flow runner anyways, so yay for adding env!

cmd/flowserver/flow.go
ericnewcomer: For the json format, i'd like to stick with verbose. Those names can all compress down to nothing if we want at some point, so keeping them as easily human readable as possible sounds ideal.

rowanseymour: Updated

flows/runs/run.go
rowanseymour: Gonna go with "don't think so, but we should test that"

PR #53, opened 2017-07-13
Inputs as separate thing

rowanseymour:


flows/inputs/base.go
```diff
@@ -0,0 +1,30 @@
...
+func (i *BaseInput) CreatedOn() time.Time        { return i.createdOn }
+func (i *BaseInput) SetCreatedOn(time time.Time) { i.createdOn = time }
+
+// Resolve resolves the passed in key to a value, returning an error if the key is unknown
+func (i *BaseInput) Resolve(key string) interface{} {
+	switch key {
+	case "time":
+		return i.createdOn
+	case "channel_uuid":
+		return i.channelUUID
```
rowanseymour: This will become a channel object in a subsequent pr

flows/runs/run.go
```diff
@@ -549,11 +541,6 @@ func (r *flowRun) MarshalJSON() ([]byte, error) {
 		re.Child = r.child.UUID()
 	}
 
-	re.Input, err = utils.EnvelopeFromTyped(r.input)
```
rowanseymour: Can't load the input here because we haven't yet loaded the channels. We don't actually need the input since we should have new input if we're reloading runs. Anyway in the next pr I'm reworking how we load stuff so it's easier to resolve references to other objects.

rowanseymour:
Closed in favor of #55

PR #54, opened 2017-07-13
Migration Tweaks

ericnewcomer:
Create update_contact action as appropriate, minor migration tweaks


flows/definition/legacy.go
```diff
@@ -369,7 +380,8 @@ func createCase(baseLanguage utils.Language, exitMap map[string]flows.Exit, r le
 
 	case "regex":
 		fallthrough
-
+	case "contains":
```
rowanseymour: you can write this `case "contains", "contains", "regex":`

PR #55, opened 2017-07-17
Event refactor

rowanseymour:
Requires https://github.com/nyaruka/rapidpro/pull/1354 to use with rapidpro unit tests


cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -72,16 +72,6 @@
...
-          "channel_uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d",
-          "contact_uuid": "ba96bf7f-bc2a-4873-a7c7-254d1927c4e3",
-          "created_on": "2000-01-01T00:00:00.000000000-00:00",
-          "id": 0,
-          "step_uuid": "",
-          "text": "Ryan Lewis",
-          "type": "msg_received",
-          "urn": "tel:+12065551212"
-        },
-        {
```
rowanseymour: We no longer repeat caller events back to the caller

cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -129,7 +119,6 @@
               "channel_uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d",
               "contact_uuid": "ba96bf7f-bc2a-4873-a7c7-254d1927c4e3",
               "created_on": "2000-01-01T00:00:00.000000000-00:00",
-              "id": 0,
```
rowanseymour: We don't use this anywhere so I think we can drop it for now

flows/events/flow_entered.go
```diff
@@ -30,38 +28,15 @@ type FlowEnteredEvent struct {
...
+		FlowUUID:    flow,
+		ContactUUID: contact,
+	}
 }
 
 // Type returns the type of this event
 func (e *FlowEnteredEvent) Type() string { return TypeFlowEntered }
 
-// Resolve resolves the passed in key for this event
-func (e *FlowEnteredEvent) Resolve(key string) interface{} {
```
rowanseymour: These kind of events are no longer become input

flows/engine/engine.go
```diff
@@ -143,24 +139,30 @@ func resumeNode(run flows.FlowRun, node flows.Node, step flows.Step, event flows
...
 	if err != nil {
 		run.AddError(step, err)
 		run.Exit(flows.StatusErrored)
 		return noDestination, step, nil
 	}
 
+	// we can now apply the caller events
+	for _, event := range callerEvents {
+		run.ApplyEvent(step, event)
+	}
```
nicpottier: I don't disagree, but mostly interested in maybe documenting here why we do the wait and apply in this order.. perhaps having a longer comment here would help. I imagine there is a good reason around flow exits and such..

cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -72,16 +72,6 @@
     {
       "events": [
         {
-          "channel_uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d",
-          "contact_uuid": "ba96bf7f-bc2a-4873-a7c7-254d1927c4e3",
-          "created_on": "2000-01-01T00:00:00.000000000-00:00",
-          "id": 0,
-          "step_uuid": "",
```
rowanseymour: At some point I think we can stop repeating "step_uuid" here as events are also nested under the steps in the path... but whilst we're generating flow step objects in rapidpro, it's a bit awkward to figure out which event occured where without these

flows/interfaces.go
```diff
@@ -163,12 +163,17 @@ type Context interface {
...
+	CreatedOn() time.Time
 	SetCreatedOn(time.Time)
 
 	Step() StepUUID
 	SetStep(StepUUID)
 
+	FromCaller() bool
+	SetFromCaller(bool)
+
+	Apply(FlowRun)
```
nicpottier: I'm surprised this doesn't return an `error`.. we don't have any case where something can go wrong when applying?

flows/interfaces.go
```diff
@@ -163,12 +163,17 @@ type Context interface {
...
 type Event interface {
-	CreatedOn() *time.Time
+	CreatedOn() time.Time
 	SetCreatedOn(time.Time)
 
 	Step() StepUUID
 	SetStep(StepUUID)
 
+	FromCaller() bool
+	SetFromCaller(bool)
```
nicpottier: I know this is because we are trying not to echo this back out, but man, smells a bit having it on our interface.

flows/engine/engine.go
```diff
@@ -143,24 +139,30 @@ func resumeNode(run flows.FlowRun, node flows.Node, step flows.Step, event flows
 		return noDestination, step, nil
 	}
 
-	err := wait.End(run, step, event)
+	// try to resume our wait with the first caller event
+	err := wait.End(run, step, callerEvents[0])
```
nicpottier: Maybe waits should take a list of events? This seems like a very specific and subtle contract with our callers.

rowanseymour: If order is important then the wait is waiting for the _next_ event, i.e. the first in the list. This feels right IMO

flows/interfaces.go
rowanseymour: You'd prefer setting the flag directly in ReadEvents?

flows/interfaces.go
rowanseymour: Currently the actions catch the error cases and turn them into error events, but yeah seems ok for events to generate error events - certainly if they come from the caller.

flows/engine/engine.go
rowanseymour: The wait is the gate keeper than decides if flow execution can proceed - so I think it makes sense that caller events are applied after that. I'll document that somewhere..

flows/interfaces.go
nicpottier: Wonder if this, step and create_on are other signs that maybe we need something wrapping an Event that is really the StepEvent or the like.. would that make sense?

So while we currently have stuff like:

```json
{ "events": "events": [
                  {
                    "groups": [
                      {
                        "name": "Survey Audience",
                        "uuid": "2aad21f6-30b7-42c5-bd7f-1b720c154817"
                      }
                    ],
                    "step_uuid": "",
                    "type": "add_to_group"
                  },
                  {
                    "field_name": "Activation Token",
                    "field_uuid": "f06c5b73-eb0d-4417-b7e0-4f650ed30dc8",
                    "step_uuid": "",
                    "type": "save_contact_field",
                    "value": "XXX-YYY-ZZZ"
                  }
]
}
```

Maybe it is more right to have something like:

```json
{ "events": "events": [
                  {
                    "step_uuid": "",
                    "event": {
                    "groups": [
                      {
                        "name": "Survey Audience",
                        "uuid": "2aad21f6-30b7-42c5-bd7f-1b720c154817"
                      }
                    ],
                    "type": "add_to_group"
                    }
                  },
                  {
                    "step_uuid": "",
                    "event": {
                    "field_name": "Activation Token",
                    "field_uuid": "f06c5b73-eb0d-4417-b7e0-4f650ed30dc8",
                    
                    "type": "save_contact_field",
                    "value": "XXX-YYY-ZZZ"
                  }
   }
]
}
```

It definitely is less than ideal that we have `step_uuid` on event for example, when this is really only an attribute that is there because of the passed back `events` list, not really an attribute of event at all. Basically maybe we should be putting events in containers that contain timestamps / step uuids instead of making them all deal with that. (and then having the confusion as to whether that is a required attribute by the caller)

Any ideas there?

rowanseymour: Well now that I've embraced the "events as real history" view I'm cool with `created_on` being the real world time. Regarding step_uuid I think we could ditch it and infer it from the nested events in the path.. but for what we're trying to do currently creating flowsteps, it's kinda awkward to figure out which events in the session are which events in the path.

I was wondering about changing `events` to something like:

```
"new_events": [
  {
    "step_uuid": "1234..."
    "events": [
       ...
    ],
  }
]
```

That makes it easy to get stuff like step time, node uuid etc for each of your new events that you need to apply in the caller.

flows/interfaces.go
nicpottier: I guess we can always add this if/when we see a need, so no need to change it now.

flows/interfaces.go
nicpottier: I like that.. at least not having `step_uuid` on event would be a move in the right direction. Should we have an `applied_on`?

rowanseymour: Is that ever going to be meaningfully different to the step time?

nicpottier: True, true, probably not.

flows/engine/engine.go
nicpottier: Ya, I can't currently really think of a case when we would resume with more than one event so I'm fine with it as is for now.

flows/interfaces.go
rowanseymour: K so then I'll update it as above - do like having caller events have all the same fields as server events

ericnewcomer: Also, from a parallel discussion, we also need to know the action uuid that generated an event, so can easily see this as a sibling of step_uuid above.

rowanseymour: We don't currently know/need that but is that something you think we should log?

nicpottier: Eric kind of has a use case for that actually in the broadcast batching.. being able to map which action created a `send_msg` event would actually be useful there I think.

rowanseymour: Ok am merging this as is for now because I need to take care of some other stuff, and have created #56

PR #57, opened 2017-07-18
Support for legacy form_field rulesets

ericnewcomer:
* Add migration for between test
* Add migration for form_field ruleset to create @(field(...)) operand
* Change @(field(...)) to be zero-based with related migration
* Change field splitting to consider space as whitespace delimiter


excellent/functions.go
```diff
@@ -939,10 +941,22 @@ func Field(env utils.Environment, args ...interface{}) interface{} {
...
 
 	fields := strings.Split(source, sep)
-	if field-1 >= len(fields) {
+	if field >= len(fields) {
 		return ""
 	}
-	return strings.TrimSpace(fields[field-1])
+
+	// when using spaces as a delimiter, we consider it splitting on whitespace, so remove empty values
+	if len(strings.TrimSpace(sep)) == 0 {
```
nicpottier: Should we only be special casing `" "` maybe? Seems slightly weird to be doing this automatically for tabs and multiple spaces.

PR #58, opened 2017-07-19
Wrap output events in container with step and action UUIDs

rowanseymour:


cmd/docgen/run.go
```diff
@@ -91,10 +91,10 @@ func eventsForAction(actionJSON []byte) (json.RawMessage, error) {
 		return nil, err
 	}
 
-	events := session.Events()
-	eventJSON := make([]json.RawMessage, len(events))
-	for i, event := range events {
-		typed, err := utils.EnvelopeFromTyped(event)
+	eventLog := session.EventLog()
```
rowanseymour: Definitely open to naming discussion - we need terminology that makes it clear that these are not events but something else that contains an event and information about how it was generated

cmd/flowrunner/main.go
```diff
@@ -21,8 +21,8 @@ import (
 )
 
 type Output struct {
-	Session json.RawMessage        `json:"session"`
-	Events  []*utils.TypedEnvelope `json:"events"`
+	Session  json.RawMessage   `json:"session"`
+	EventLog []json.RawMessage `json:"events"`
```
rowanseymour: Haven't done it yet but I think this should now be something other than just `events` for clarity - e.g. `event_log` or `log`

cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -1,4 +1,5 @@
 {
+  "caller_events": [],
```
rowanseymour: Randomness has decided that "caller_events" should always come first in the test files now

cmd/docgen/run.go
ericnewcomer: How about just output?

cmd/flowrunner/main.go
ericnewcomer: Hmmm.. guess we have an Output already. Maybe event_log makes sense then.

cmd/docgen/run.go
rowanseymour: I like output but what would be the name of each item in output?

ericnewcomer: Could be OutputLog, but Output having output with output logs? That's a bit nutty. Event log at least removes confusion with top level Output.

cmd/flowrunner/testdata/flows/all_actions_test.json
ericnewcomer: or the alphabet?

rowanseymour: ah i forgot they were previously called "resume_events" before I renamed them to "caller_events"

flows/runs/event_log.go
```diff
@@ -0,0 +1,56 @@
+package runs
+
+import (
+	"encoding/json"
+
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/utils"
+)
+
+type eventLogEntry struct {
```
nicpottier: This naming seems a bit tortured. It is a log, of events, but actually no it is a log of entries which are of events? 

What about it is just a log and it has log entries? We don't need to describe that the log entries are of events do we?

cmd/flowrunner/testdata/flows/all_actions_test.json
nicpottier: Looks like while golang does have random map iteration, json encoding of maps is sorted by key.
https://golang.org/src/encoding/json/encode.go

flows/runs/event_log.go
rowanseymour: so `session.Log() []flows.LogEntry`  ? It's a bit vague but I like the brevity

how do we feel about using `log` in response, e.g. 

```json
{
  "session": {},
  "log": [{}, {}]
}
```

nicpottier: I like log in response. And ya I like the brevity as well.

PR #60, opened 2017-07-20
Self-describing constants for run and webhook statuses

rowanseymour:
We use self-describing constants everywhere else for things like event types so not seeing a reason to keep these as single chars, e.g. "expired" vs "errored" instead of "E" vs "X"


nicpottier:
I hate it but FINE!

PR #61, opened 2017-07-21
Sessions and runs refactor

rowanseymour:
This wasn't supposed to turn into a refactor but you know how things go. Currently the engine starts up something like:

1. flows + contacts loaded
2. FlowEnvironment created from the flows + runs + contacts
3. Session created from that
4. Runs created and assigned a session, also added to the FlowEnvironment

But this PR reworks things so that:

1. flows + channels + contacts loaded
2. SessionEnvironment created from the flows + channels + contacts (i.e. dependencies)
3. Session created from that
4. Runs created from the session (i.e. no such thing as a session-less run)

This means things in the session don't have to store UUIDs for flows, contacts and channels, and resolve those later to real objects - they can resolve them as they're being unmarshaled.

Arguably we need to go one step further and have SessionEnvironment do the loading of flows + channels + contacts in the order channels &rarr; contacts &rarr; flows so that we can resolve the dependencies between those as they are loaded (e.g. contact will have a channel)

### DIFF
--- a/cmd/flowrunner/testdata/channels/default.json
+++ b/cmd/flowrunner/testdata/channels/default.json
@@ -1,5 +1,6 @@
 {
-        "uuid": "channel1-test-4b7f-a34b-e37e31e86451",
+        "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d",
         "name": "Twilio Channel",
+        "address": "+12345671111",
         "type": "T"
-    }
+}
--- a/cmd/flowrunner/testdata/flows/all_actions_test.json
+++ b/cmd/flowrunner/testdata/flows/all_actions_test.json
@@ -183,7 +183,6 @@
       "session": {
         "runs": [
           {
-            "channel_uuid": "",
             "child_uuid": "",
             "contact_uuid": "ba96bf7f-bc2a-4873-a7c7-254d1927c4e3",
             "created_on": "2000-01-01T00:00:00.000000000-00:00",
@@ -339,7 +338,6 @@
             }
           },
           {
-            "channel_uuid": "",
             "contact_uuid": "ba96bf7f-bc2a-4873-a7c7-254d1927c4e3",
             "created_on": "2000-01-01T00:00:00.000000000-00:00",
             "exited_on": "2000-01-01T00:00:00.000000000-00:00",
--- a/cmd/flowserver/flow.go
+++ b/cmd/flowserver/flow.go
@@ -104,7 +104,7 @@ func handleStart(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 	}
 
 	// start our flow
-	session, err := engine.StartFlow(sessionEnv, flowlist[0], contact, nil, callerEvents, start.Extra)
+	session, err := engine.StartFlow(sessionEnv, flowlist[0], contact, nil, callerEvents, start.Extra, nil)
 	if err != nil {
 		return nil, fmt.Errorf("error starting flow: %s", err)
 	}
--- a/cmd/flowserver/flow.go
+++ b/cmd/flowserver/flow.go
@@ -94,8 +94,8 @@ func handleStart(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 		return nil, err
 	}
 
-	// build our flow environment
-	flowEnv := engine.NewFlowEnvironment(env, flowlist, channelList, []*flows.Contact{contact})
+	// build our session environment
+	sessionEnv := engine.NewSessionEnvironment(env, flowlist, channelList, []*flows.Contact{contact})
 
 	// read our caller events
 	callerEvents, err := events.ReadEvents(start.Events)
@@ -104,7 +104,7 @@ func handleStart(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 	}
 
 	// start our flow
-	session, err := engine.StartFlow(flowEnv, flowlist[0], contact, nil, callerEvents, start.Extra)
+	session, err := engine.StartFlow(sessionEnv, flowlist[0], contact, nil, callerEvents, start.Extra)
 	if err != nil {
 		return nil, fmt.Errorf("error starting flow: %s", err)
 	}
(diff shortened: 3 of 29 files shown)


cmd/flowrunner/testdata/channels/default.json
```diff
@@ -1,5 +1,6 @@
 {
-        "uuid": "channel1-test-4b7f-a34b-e37e31e86451",
+        "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d",
```
rowanseymour: This has to be correct channel uuid used in the test files now

cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -183,7 +183,6 @@
       "session": {
         "runs": [
           {
-            "channel_uuid": "",
```
rowanseymour: Runs now longer have a channel - channel is on input (received on) and contact (preferred)

cmd/flowserver/flow.go
```diff
@@ -151,39 +159,37 @@ func handleResume(w http.ResponseWriter, r *http.Request) (interface{}, error) {
...
+	channelList, err := flows.ReadChannels(resume.Channels)
 	if err != nil {
 		return nil, err
 	}
-	if len(session.Runs()) == 0 {
-		return nil, utils.NewValidationError("session: must include at least one run")
-	}
 
-	// clear the event log if it was passed in
-	session.ClearLog()
```
rowanseymour: We don't load the log

flows/definition/flow.go
```diff
@@ -28,10 +27,6 @@ func (f *flow) Nodes() []flows.Node                    { return f.nodes }
 func (f *flow) Translations() flows.FlowTranslations   { return f.translations }
 func (f *flow) GetNode(uuid flows.NodeUUID) flows.Node { return f.nodeMap[uuid] }
 
-func (f *flow) CreateRun(env flows.FlowEnvironment, contact *flows.Contact, parent flows.FlowRun) flows.FlowRun {
```
rowanseymour: Sessions are responsible for creating runs now

flows/engine/engine.go
```diff
@@ -6,12 +6,17 @@ import (
...
 	"github.com/nyaruka/goflow/flows/events"
+	"github.com/nyaruka/goflow/flows/runs"
 )
 
 // StartFlow starts the flow for the passed in contact, returning the created FlowRun
-func StartFlow(env flows.FlowEnvironment, flow flows.Flow, contact *flows.Contact, parent flows.FlowRun, callerEvents []flows.Event, extra json.RawMessage) (flows.Session, error) {
-	// build our run
-	run := flow.CreateRun(env, contact, parent)
+func StartFlow(env flows.SessionEnvironment, flow flows.Flow, contact *flows.Contact, parent flows.FlowRun, callerEvents []flows.Event, extra json.RawMessage, session flows.Session) (flows.Session, error) {
+	if session == nil {
```
rowanseymour: Session is passed in by start_flow actions to ensure runs are created in same session

flows/events/msg_received.go
```diff
@@ -27,9 +26,9 @@ const TypeMsgReceived string = "msg_received"
 // @event msg_received
 type MsgReceivedEvent struct {
 	BaseEvent
-	ChannelUUID flows.ChannelUUID `json:"channel_uuid"     validate:"required,uuid4"`
-	URN         flows.URN         `json:"urn"              validate:"required"`
-	ContactUUID flows.ContactUUID `json:"contact_uuid"     validate:"required,uuid4"`
+	ChannelUUID flows.ChannelUUID `json:"channel_uuid" validate:"uuid4"`
```
rowanseymour: Making channel optional since we have channel-less messages in rapidpro

flows/runs/run.go
```diff
@@ -68,14 +68,8 @@ func (e *runEnvironment) refreshLanguagesCache() {
 type flowRun struct {
 	uuid flows.RunUUID
 
-	contact     *flows.Contact
-	contactUUID flows.ContactUUID
```
rowanseymour: Runs no longer need to store uuids for flows and contacts because they're loaded after those things are loaded

cmd/flowserver/flow.go
```diff
@@ -38,6 +38,7 @@ func (r *flowResponse) MarshalJSON() ([]byte, error) {
 type startRequest struct {
 	Environment *json.RawMessage       `json:"environment"`
 	Flows       json.RawMessage        `json:"flows"    validate:"required"`
+	Channels    []json.RawMessage      `json:"channels,omit_empty"`
```
nicpottier: No need to change this now, but food for thought, should we just change to pass in a `SessionEnvironment` here instead of breaking out all the things we use to build said `SessionEnvironment`.. seems we are breaking this apart and then basically having two formats for what is a `SessionEnvironment`

rowanseymour: I think that makes sense - tho I think then contacts would be a list, and we'd a contact uuid somewhere else to say which contact is being started

flows/definition/flow.go
nicpottier: Sweet, way nicer.

flows/engine/engine.go
nicpottier: From an ordering perspective it feels to me like Session should be earlier in this list, perhaps before Flow. I always try to think of parameter order being from big to small or broad to specific.

rowanseymour: Generally agree but any exceptions when a param is usually null? In Python it would likely go last because it has a default value of none

nicpottier: Hrmm, ya, do agree with that, though this method has a lot of nilable things as well.

Is there ever a time when the both session and parent will be nil at the same time? Or where the session of the parent is not the same as the passed in session?

IE, do we really need session given that we have parent?

rowanseymour: oh good call yeah we don't need it - either this method is creating a new session or we have a parent run

PR #62, opened 2017-07-24
Session refactor part 2

rowanseymour:
* **assets** is **flows** + **channels**, i.e. things which are immutable from the pov of the engine, and potentially cacheable
* **session**:
   - is the top-level session with the engine
   - is created from **assets** + **environment** + **contact**
* **output** is **session** + **log** - although at runtime the session holds the log, it's useful to split these two up in the output as the caller is expected to pass back the **session** but not the log

Thus a start request becomes: 

```json
{ 
  "assets": {
    "channels": [...],
    "flows": [...]
  },
  "environment": {...},
  "contact": {...},
  "flow_uuid": "13242...",
  "events": [...]
}
```

The response becomes:

```json
{ 
  "session": {
    "runs": [...],
    "contact": {...}
  },
  "log": [...]
}
```

And a resume request becomes:

```json
{ 
  "assets": {
    "channels": [...],
    "flows": [...]
  },
  "environment": {...},
  "session": {
    "runs": [...],
    "contact": {...}
  },
  "events": [...]
}
```

Notes
 * The flow to start is now explicit and not just the first flow in the list
 * RapidPro will always want to delete the contact from the session in a response and rebuild it for each request - but a different caller might not have to worry about outside changes to the contact and might always pass the session json back to the engine exactly as it received it
 * Arguably environment should be also passed back (like contact) on the session as it might be changed by an event


cmd/flowserver/flow.go
```diff
@@ -37,9 +32,9 @@ func (r *flowResponse) MarshalJSON() ([]byte, error) {
 
 type startRequest struct {
 	Environment *json.RawMessage       `json:"environment"`
-	Flows       json.RawMessage        `json:"flows"    validate:"required"`
-	Channels    []json.RawMessage      `json:"channels,omit_empty"`
-	Contact     json.RawMessage        `json:"contact"  validate:"required"`
+	Assets      json.RawMessage        `json:"assets"           validate:"required"`
+	Flow        flows.FlowUUID         `json:"flow_uuid"        validate:"required"`
```
nicpottier: Why the change from this being the first? It being the first has the nice property that it removes a whole class of possible errors.

rowanseymour: Not following - this is just the flow uuid to start...

rowanseymour: Should rename this to `FlowUUID`

nicpottier:
So some random thoughts:
 * it seems slightly dirty that we require/expect rebuilding/replacing the contact on a session when they resume. Sure would be nice to have that contract be simple in that you just pass back the session. Wonder if maybe contact should be an asset as well?
 * Commented elsewhere on this, but why pass in the flow id vs first flow?
 * Are `assets` just part of the environment? To me that seems kinda right, but I agree that maybe assets are special in a way in that they could be lazy loaded. (IE, we might standardize on something like https://spacetelescope.github.io/understanding-json-schema/structuring.html

ericnewcomer:
Was just going to say the same thing re: contact. I really don't like the idea of having to crack open a session and muck with it at the caller end. Seems there should be a bright line between things that could reasonably be expected to change outside the engine (contacts) and things only the engine should mess with (runs).

flows/engine/assets.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/flows/definition"
+)
+
+type assets struct {
```
nicpottier: I am warming to this name, though fear that it might be confusing to others due to assets commonly being used for graphics / css etc.. Not sure I have a better suggestion though so that doesn't make a very powerful argument.

ericnewcomer: I think I don't mind assets too much. These play a similar role for the engine as sprites for a game, in a way. I think my expectation though is that they are static. While these are static within a single execution cycle, they are likely to change quite a bit. That's probably okay though.

cmd/flowserver/flow.go
nicpottier: I was saying we now have the possible error of that flow uuid not being present in the flow list, which isn't possible when you just say the first flow will be started.

That said, it does feel weird to have that behavior of implicitly starting the first flow in `assets` so I agree this is the right change.

rowanseymour:
* My thinking on contact (and maybe environment) is that they are part of the session (i.e. state that can change) and the simplest caller won't want to refresh the contact part of the session because they won't be operating in a world where the contact changes outside of the engine - they will just pass the session back as they got it
 * Assets are supposed to represent a clear boundary between things that are immutable and cacheable from the pov of the engine (flows, channels) and things which carry session state (runs, contact, input, etc). So I don't see a contact as an asset.

rowanseymour:
Also I made assets separate from environment because I think the latter belongs on session and is set by an event like 'set_env', i.e. we might want to pass back environment so callers know what the current environment is and whether they need to change it.

ericnewcomer:
Is contact both? Can it be provided as an asset that would override a session contact? I'm really not okay with splicing the session. I don't want to optimize for the simplest caller, I want to optimize for the first (and possibly only) caller.

nicpottier:
I think the distinction of what is an "asset" and what isn't is going to be a really difficult line to explain. If we instead just say hey, no outside objects in `Session` only references to those, then that feels slightly clearer to me.

On the environment is feeling more and more to me like something that is an event and an attribute on the session. Callers can choose to set the environment on every resume by generating a new `update_env` event, but they don't have to. Environments don't have ids, so they can't be assets, nor are they loaded, so I think they are indeed a different thing.

rowanseymour:
I'm not getting why refreshing a single key in the session json is such a big deal but it's not a hill I feel the need to die on. 

Conceptually for me, a session is a contact's interaction with the flow engine, and there's a logic to it being embedded in the session for requests. It makes a clear boundary between immutable/cacheable assets which an engine uses (`assets`), and mutable session state (`session`).

nicpottier:
IE, I prefer contact as an asset, and perhaps even having a `set_contact` event leading a start flow?

nicpottier:
The only asset that is immutable right now (and then is more of a RapidPro thing) is Channel, but I don't think even that is a contract the engine necessarily cares about is it? So do your feelings change since Flows are clearly mutable and so are Channels in a generic world?

nicpottier:
On having `environment` and `contact` being set by events, maybe there is something nice to that, in that for the simplest flows (think biz logic), you now no longer need either a contact or environment. (the former can be nil, the latter can just be a default) That's kinda neat no?

nicpottier:
```
I'm not getting why refreshing a single key in the session json is such a big deal but it's not a hill I feel the need to die on.
```

rowanseymour:
From the pov of the engine - flows and channels are immutable - i.e. the engine can't change them. Neither of those are actually immutable in rapidpro so if we did cache them on the flowserver side we'd need a way to let flowserver know that they've changed.

The immutability on the flowserver side manifests itself as not including them (`assets`) in the flowserver response.

`contact`, `runs`, `input` etc hold mutable session state which is why we return them so the caller can see what has changed. 

That to me is the fundamental difference and why I don't see a contact as an asset. I agree that setting environment and contact on the session using events (just like `input`) could be neat, but I don't see how that requires either of those things to be assets.

nicpottier:
Ok I think I see the distinction you are making, it is more about the response having `contact` in it (in your case in the Session) so that the caller can see the new state. And if you stick that in `assets` then that seems weird to return in a different thing.

What about a hybrid then that is consistent?

If we are saying we like the idea of `env` being passed in via an `update_env` event, and if we then also say that if the caller wants to change that, then it should pass in a new `update_env` on the resume, then why treat contacts differently?

IE, maybe `contact` is indeed on the session, but it is set and updated in the same way for start and resume, via a `update_contact` event, same as `env`. Both `env` and `contact` are then attributes on the session, and it is up to the caller whether they want always fresh contacts or environments and if so then they should start their resumes with `update_env` and `update_contact` events.

How do we feel about that?

ericnewcomer:
Ya, that's precisely what I was alluding to above. I think @rowanseymour's point makes sense whereby, the engine is returning things that it understands as truth. The caller can inspect that, and if they want to change that, they can include an event to update it if they like. I think that satisfies everything in this thread. No splicing of sessions, nor any requirements for contacts, and the simplest caller can still be braindead.

rowanseymour:
So in principle I like that idea - practically I have some concerns about full JSON representations of contacts now being in the path events - for our orgs with 1000 contact fields that's a lot of bloat.

ericnewcomer:
I don't think anything discussed here necessitates full JSON representations in the path, so not sure what you mean by that. Can you clarify?

nicpottier:
I think he's saying if we use an update_contact event then that goes in the
path. To that point that only seems super expensive if the contact actually
has 1000 vars set, we don't need to include values that are unset do we?
Might be something we choose to optimize out as well at some point?

rowanseymour:
Yeah - if update_contact events from the caller are the payload for the complete contact, then such events applied by the engine will have the same payload.

One possible advantage of storing contact on the session json is that RapidPro could be smart about only rebuilding the JSON representation of the contact for a resume if the contact modified_on is > than the session modified_on.. tho obviously there's potential for error there if we're not careful with our timestamps.

nicpottier:
Ya I would definitely see us trying to optimize when we  actually fire `update_env` or `update_contact` events to only be when necessary.

rowanseymour:
Ok so what I think we're settling on is:

- contact is set for a start using `set_contact` ⎈
- that is stored as `session.Contact()` and returned as `{"session": {"contact": {...}}}`
- when resuming, RapidPro will determine if it has changed the contact ⎈
    - if it has, include a new `set_contact` event with updated contact

⎈ I prefer `set_contact` to `update_contact` because this isn't a partial update like the `update_contact` action - this is just overwriting the session's contact with a new representation

cmd/flowrunner/main.go
```diff
@@ -26,9 +25,8 @@ type Output struct {
 }
 
 type FlowTest struct {
-	Extra        json.RawMessage        `json:"extra,omitempty"`
-	CallerEvents []*utils.TypedEnvelope `json:"caller_events"`
-	Outputs      []json.RawMessage      `json:"outputs"`
+	CallerEvents [][]*utils.TypedEnvelope `json:"caller_events"`
```
rowanseymour: Until now we've used caller_events as a list of events where each event corresponds to a resume - but now we need initial events (e.g. set_extra), and we want to support multiple events per start/resume

cmd/flowserver/expression.go
```diff
@@ -39,11 +38,7 @@ func handleExpression(w http.ResponseWriter, r *http.Request) (interface{}, erro
 		return nil, fmt.Errorf("missing context or expression element")
 	}
 
-	// build up our context
-	context, err := runs.ReadContext(expression.Context)
```
rowanseymour: I'm not sure what the plans are for this enpoint but it's not to easy to unmarshall a run or a contact these days without other information, e.g. channels - so I'm switching this to take a generic bod as context.

cmd/flowserver/flow.go
```diff
@@ -36,12 +31,9 @@ func (r *flowResponse) MarshalJSON() ([]byte, error) {
...
 type startRequest struct {
-	Environment *json.RawMessage       `json:"environment"`
-	Flows       json.RawMessage        `json:"flows"    validate:"required"`
-	Channels    []json.RawMessage      `json:"channels,omit_empty"`
-	Contact     json.RawMessage        `json:"contact"  validate:"required"`
-	Extra       json.RawMessage        `json:"extra,omitempty"`
-	Events      []*utils.TypedEnvelope `json:"events"`
+	Assets json.RawMessage        `json:"assets"           validate:"required"`
+	Flow   flows.FlowUUID         `json:"flow_uuid"        validate:"required"`
+	Events []*utils.TypedEnvelope `json:"events"`
```
rowanseymour: Definitely wondering about starting the flow via an event...

excellent/legacy.go
```diff
@@ -175,6 +175,12 @@ func newVars() vars {
 				vars:    contactVars,
 			},
 		},
+		"channel": vars{
+			"__default__": "contact.channel.address",
+			"name":        "contact.channel.name",
+			"tel":         "contact.channel.address",
+			"tel_e164":    "contact.channel.address",
```
rowanseymour: This might require some tweaking but it would be nice if we made address always be e164 format for tel channels

flows/inputs/msg.go
```diff
@@ -87,12 +87,15 @@ func ReadMsgInput(env flows.SessionEnvironment, envelope *utils.TypedEnvelope) (
...
 
-func (r *MsgInput) MarshalJSON() ([]byte, error) {
-	envelope := msgInputEnvelope{
-		baseInputEnvelope: baseInputEnvelope{ChannelUUID: r.Channel().UUID(), CreatedOn: r.CreatedOn()},
-		URN:               r.urn,
-		Text:              r.text,
+func (i *MsgInput) MarshalJSON() ([]byte, error) {
+	var envelope msgInputEnvelope
+
+	if i.Channel() != nil {
```
rowanseymour: This is an unrelated fix to stop marshalling of run.input blowing up if input is channel-less

flows/runs/context.go
```diff
@@ -1,38 +1,31 @@
...
 
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/utils"
 )
 
-// NewContextForContact creates a new context for the passed in contact
-func NewContextForContact(contact *flows.Contact, run flows.FlowRun) flows.Context {
-	context := context{contact: contact, run: run}
-	return &context
+type runContext struct {
```
rowanseymour: It's simpler to not store off the contact at this point because the contact might change

utils/environment.go
```diff
@@ -68,13 +68,13 @@ type envEnvelope struct {
 	Languages  LanguageList `json:"languages"`
 }
 
-func ReadEnvironment(data *json.RawMessage) (*environment, error) {
+func ReadEnvironment(data json.RawMessage) (*environment, error) {
```
rowanseymour: `RawMessage` is a slice and slices are passed by reference in go

rowanseymour:
@nicpottier @ericnewcomer I'm doing a "buy-2-get-one-free" offer on converting things to events so this PR now has `set_contact`, `set_environment` and `set_extra` events

utils/environment.go
nicpottier: I think this change is ok because it is just reading, but that is not actually true about slices in go. They are essentially a pointer to an array which has length information, but that pointer can change if you append to the slice, which is why the primitive `append(slice, newVal)` has to be assigned to `slice`.

excellent/legacy.go
nicpottier: We have the same thing on contact. I think it is just a matter of e164 addresses not being terribly friendly  if you are including it in a message, thus the tel version. We could maybe add a new excellent method to do the translation instead? But that isn't super friendly either.

utils/environment.go
rowanseymour: Sure sure - `data` here contains a pointer to the data so you don't have to worry about it passing the entire buffer as a value... so it sorta acts like passing reference. I noticed the json.Marshal/Unmarshal methods don't take pointers so I think this is more goish (?).

nicpottier: Yep, change is good, just making sure you knew that slices are definitely not pass by reference as this is a common go error. (you can pass a map that way and it is ok to change it, just not for slices)

nicpottier:
So much love for this! Double plus good improvement

excellent/legacy.go
rowanseymour: I added https://github.com/nyaruka/goflow/issues/63 to record this

rowanseymour:
@nicpottier does love = approve ?

PR #65, opened 2017-07-26
Tests and fixes

rowanseymour:


excellent/legacy.go
```diff
@@ -225,12 +225,12 @@ func isDate(operand string) bool {
 	return false
 }
 
-// TranslateTemplate will take an old expression and translate it to the new format
```
rowanseymour: Translate is overloaded

flows/definition/legacy.go
```diff
@@ -147,23 +147,29 @@ type translationMap map[utils.Language]string
...
-	(*langTranslations)[uuid] = items
-	(*translations)[lang] = langTranslations
+	// ensure we have a translation set for this item
+	itemTrans, found := (*langTranslations)[itemUUID]
+	if !found {
+		itemTrans = itemTranslations{}
+		(*langTranslations)[itemUUID] = itemTrans
+	}
+
+	itemTrans[propKey] = translation
```
rowanseymour: This fixes a second property translation overwriting the first

flows/definition/legacy.go
```diff
@@ -296,34 +308,33 @@ func createAction(baseLanguage utils.Language, a legacyAction, fieldMap map[stri
...
 		}
 
-		translated, _ := excellent.TranslateTemplate(a.Value)
+		migratedValue, _ := excellent.MigrateTemplate(a.Value)
+
+		// flows now have different action for name changing
+		if a.Field == "name" || a.Field == "first_name" {
+			// we can emulate setting only the first name with an expression
+			if a.Field == "first_name" {
+				migratedValue = fmt.Sprintf("%s @(word_slice(contact.name, 2, -1))", migratedValue)
```
rowanseymour: This is bit mad but we don't currently have a concept of changing only first name in the new world

nicpottier: We have a way of changing it in the old world? I thought we just set name as a whole?

rowanseymour: https://github.com/nyaruka/rapidpro/blob/1ba705623953e28d6378957b55b41f4affc2ecf1/temba/flows/models.py#L6101-L6101

I'm planning to build a time machine and go back a few years and warn us not to add things like this

PR #72, opened 2017-07-27
New HasPattern test, more test migrations and unit testing

rowanseymour:


flows/definition/legacy.go
```diff
@@ -401,56 +412,78 @@ func createAction(baseLanguage utils.Language, a legacyAction, fieldMap map[stri
...
+	case "in_group":
+		test := groupTest{}
+		err = json.Unmarshal(r.Test.Data, &test)
+		arguments = []string{string(test.Test.UUID)}
+
+	case "subflow":
+		newType = "has_any_word"
+		test := subflowTest{}
+		err = json.Unmarshal(r.Test.Data, &test)
+		arguments = []string{test.ExitType}
```
rowanseymour: I think possible old and new world statuses are the same here - completed or expired

flows/definition/legacy.go
```diff
@@ -401,56 +412,78 @@ func createAction(baseLanguage utils.Language, a legacyAction, fieldMap map[stri
...
+
+	case "webhook_status":
+		test := webhookTest{}
+		err = json.Unmarshal(r.Test.Data, &test)
+		if test.Status == "success" {
+			arguments = []string{"success"}
+		} else {
+			arguments = []string{"response_error"}
```
rowanseymour: failure in old world covers both response_error and connection_error in new world... but this seems like the best approximation

excellent/tests.go
```diff
@@ -392,6 +392,44 @@ func HasBeginning(env utils.Environment, args ...interface{}) interface{} {
...
+//
+// Both strings are trimmed of surrounding whitespace, but otherwise matching is strict
+// without any tokenization.
+//
+//   @(has_pattern("This is my Kazoo", "^kazoo")) -> false
+//   @(has_pattern("Kazoos are great", "^kazoo")) -> true
+//   @(has_pattern("Kazoos are great", "^kazoo").match) -> "Kazoo"
+//
+// @test has_pattern(string, pattern)
+func HasPattern(env utils.Environment, args ...interface{}) interface{} {
```
rowanseymour: Will need to take care of python regex differences: https://github.com/nyaruka/goflow/issues/73

flows/definition/legacy.go
```diff
@@ -401,56 +412,78 @@ func createAction(baseLanguage utils.Language, a legacyAction, fieldMap map[stri
...
+		}
+
 	default:
-		testType = translateTest(testType)
+		return routers.Case{}, fmt.Errorf("Migration of '%s' tests no supported", r.Test.Type)
```
rowanseymour: Better to blow up than not properly migrate a test

excellent/tests.go
nicpottier: Might be worth pointing out that matches are also case insensitive.

flows/definition/testdata/test_migrations.json
```diff
@@ -0,0 +1,372 @@
...
+    {
+        "legacy_test": {
+            "type": "contains_only_phrase",
+            "test": {
+                "eng": "yes yeah",
+                "fra": "oui"
+            }
+        },
+        "expected_case": {
+            "uuid": "������������������������������������",
```
nicpottier: What's with the `�`, are these placeholders?

rowanseymour: Wildcard chars so the tests can match two strings but ignore parts like these UUIDs which are non deterministic. Using `�` instead of `?` means one doesn't have to worry about real question marks.

PR #77, opened 2017-07-31
Ruleset migration fixes and tests

rowanseymour:


flows/definition/legacy.go
```diff
@@ -563,15 +569,18 @@ func createRuleNode(lang utils.Language, r legacyRuleSet, translations *flowTran
...
 	exits, cases, defaultExit := parseRules(lang, r, translations)
 
 	switch r.Type {
 	case "subflow":
 		// subflow rulesets operate on the child flow status
 		node.router = &routers.SwitchRouter{
-			Default: defaultExit,
-			Operand: "@child.status",
-			Cases:   cases,
+			BaseRouter: baseRouter,
```
nicpottier: Passing in the base class like this is probably starting to get to the point where we should have a constructor instead.. ie, `NewSwitchRouter()`

flows/definition/legacy_test.go
```diff
@@ -127,24 +187,90 @@ func TestTestMigration(t *testing.T) {
...
 }
 
-func checkFlowLocalization(t *testing.T, flow *LegacyFlow, expectedLocalization json.RawMessage) {
+func checkFlowLocalization(t *testing.T, flow *LegacyFlow, expectedLocalizationRaw json.RawMessage, substitutionSource json.RawMessage) {
 	actualLocalization := *flow.translations.(*flowTranslations)
 	actualLocalizationRaw, _ := json.Marshal(actualLocalization)
 	actualLocalizationJSON := formatJSON(actualLocalizationRaw)
-	expectedLocalizationJSON := formatJSON(expectedLocalization)
+
+	// Because localization keys are UUIDs and some of those may be generated during migration, ordering of localized
```
nicpottier: I think Eric was talking about adding UUIDs to our current Actions in our flow defs in order to make this a bit simpler and more predictable. (I think this was also needed to simplify merging all the msgs from a broadcast together)

rowanseymour: I think that's already happened - but cases still use newly-generated UUIDs because the rule UUIDs become the exit UUIDs

PR #80, opened 2017-08-02
Run Expiration

rowanseymour:
* Add ExpiresAfterMinutes field to flows
* Ensure that's migrated on legacy flows
* Reset run expiration when input is received

### DIFF
--- a/flows/runs/run.go
+++ b/flows/runs/run.go
@@ -163,10 +163,29 @@ func (r *flowRun) SetWebhook(rr *utils.RequestResponse) {
 func (r *flowRun) Extra() utils.JSONFragment         { return r.extra }
 func (r *flowRun) SetExtra(extra utils.JSONFragment) { r.extra = extra }
 
+func (r *flowRun) ExpiresOn() *time.Time { return r.expiresOn }
+func (r *flowRun) ResetExpiration(from *time.Time) {
+	if r.Flow().ExpireAfterMinutes() != nil {
+		if from == nil {
+			now := time.Now().UTC()
+			from = &now
+		}
+
+		expiresAfterMinutes := time.Duration(*r.Flow().ExpireAfterMinutes())
+		expiresOn := from.Add(expiresAfterMinutes * time.Minute)
+
+		r.expiresOn = &expiresOn
+		r.setModifiedOn(time.Now().UTC())
+	}
+
+	if r.Parent() != nil {
+		r.Parent().ResetExpiration(r.expiresOn)
+	}
+}
+
 func (r *flowRun) CreatedOn() time.Time        { return r.createdOn }
 func (r *flowRun) ModifiedOn() time.Time       { return r.modifiedOn }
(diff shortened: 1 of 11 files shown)


flows/waits/flow.go
```diff
@@ -14,25 +13,6 @@ type FlowWait struct {
 	FlowUUID flows.FlowUUID `json:"flow_uuid"   validate:"required,uuid4"`
 }
 
-func (w *FlowWait) Resolve(key string) interface{} {
```
rowanseymour: We don't currently expose waits in the context. Not it makes sense to either since if there's an active wait, we're not in the engine by definition.

flows/runs/run.go
```diff
@@ -163,10 +163,29 @@ func (r *flowRun) SetWebhook(rr *utils.RequestResponse) {
...
 
+func (r *flowRun) ExpiresOn() *time.Time { return r.expiresOn }
+func (r *flowRun) ResetExpiration(from *time.Time) {
+	if r.Flow().ExpireAfterMinutes() != nil {
+		if from == nil {
+			now := time.Now().UTC()
+			from = &now
+		}
+
+		expiresAfterMinutes := time.Duration(*r.Flow().ExpireAfterMinutes())
```
nicpottier: Don't need the `*` here I don't think, do you?

flows/runs/run.go
```diff
@@ -163,10 +163,29 @@ func (r *flowRun) SetWebhook(rr *utils.RequestResponse) {
...
+		expiresOn := from.Add(expiresAfterMinutes * time.Minute)
+
+		r.expiresOn = &expiresOn
+		r.setModifiedOn(time.Now().UTC())
```
nicpottier: Question: should modified on just be `from` here? Shouldn't we be consistent here in what we consider modified on to represent vs expires on?

flows/waits/flow.go
nicpottier: Ya I think that's probably right.

flows/runs/run.go
rowanseymour: ExpireAfterMinutes() returns `*int` because it's an optional field.. tho should it be?

flows/runs/run.go
rowanseymour: `from` is sometimes a child run's expiredOn so wouldn't make sense to use for modifiedOn.

There is an argument for the engine not maintaining createdOn or modifiedOn and letting those be the db times on the caller side. That's probably what we need if we want run syncing over the API to still work properly.

flows/runs/run.go
nicpottier: Ya, maybe we should treat anything less than or equal to 0 as none? (or validate >= 0)

flows/runs/run.go
nicpottier: Removing those at the run level would remove the weirdness of parent and child run timestamps wouldn't it? bonus! I'm fine with either really.

Issue #44, opened 2017-06-30
Error codes for events?


ericnewcomer:
We may want to give callers more granular control on how to reliably handle error events. For example, it may be useful to take a different action if the flow blew up due to a runtime cycle.

rowanseymour:
It might be sufficient to communicate back to the caller the "class" of error. For example I see at least 3 different error "classes":

1. An error may have occurred but the show goes on (e.g. expression evaluation, maybe user didn't mean for something to be an expression)
2. An error occurred so this action is being skipped but the show goes on (e.g. your webhook URL was an expression that evaluated to an empty string)
3. An error because your flow can't be run (e.g. this test doesn't exist)

rowanseymour:
Just noting that the way we're currently going with this is that there are essentially 4 kinds of error:

* **Server error**: Internal things, e.g. server runs out of memory. 
  - Returned to the caller as an error response with message (500)
  - Caller can retry the same call later
* **Request error**: These are things that we will try to catch at the validation stage of a request, e.g. structural problems in the flow. 
  - Returned to the caller as an error response with message (400)
  - Caller shouldn't retry the same call
* **Fatal runtime error**: These are things which we can't detect at the validation stage but also can't continue from, e.g. loop detected
  - Session and run are given status=errored and event log contains an error event
  - Caller can't resume this session
* **Non-fatal runtime error**: These are more things which we can't detect at the validation stage but can safely ignore and continue from, e.g. template evaluation problem
  - Event log contains an error event
  - Caller can ignore or surface the error to the user (e.g. in simulator window)

rowanseymour:
Closing this as much has changed. We no longer have a flowserver - mailroom is calling goflow directly so there's no scope for server/request errors.

But we do have `error` and `failure` events (fatal error that prevents flow contiuation)

PR #81, opened 2017-08-04
No more recursion, and wait refactor

rowanseymour:
* Waits no longer used for subflows
* Wait is session level - describes what the engine is waiting for from the caller
* Sessions have their own status
* StartFlowAction no longer recursively starts a flow - subflows and main flows started by same mechanism
* Errors have associated action UUIDs
* Error events are either fatal and shutdown the session or non-fatal and can be ignored by caller (see #44)
* Flow entered event replaced by flow trigger event which occurs on the parent rather than child


cmd/flowrunner/testdata/flows/subflow_loop_test.json
```diff
@@ -19,10 +19,10 @@
         {
           "action_uuid": "300f02ba-e0b5-4991-bed6-4c240cdb8743",
           "event": {
-            "contact_uuid": "ba96bf7f-bc2a-4873-a7c7-254d1927c4e3",
             "created_on": "2000-01-01T00:00:00.000000000-00:00",
             "flow_uuid": "a8d27b94-d3d0-4a96-8074-0f162f342195",
-            "type": "flow_entered"
+            "parent_run_uuid": "",
```
rowanseymour: This test plays out a bit differently now since we don't use enter/exit events to detect loops but lean on the same node-level tracking in the main loop
- Parent flow runs, triggers Child Flow
- Child Flow runs, triggers Parent Flow again (run no. 2)
- Parent Flow (run no. 2) tries to start but immediately hits fatal error because of loop detection
- Child Flow is errored and exited
- Parent flow (run no. 1) is errored and exited
- Session is errored and exited

flows/actions/start_flow.go
```diff
@@ -41,57 +39,6 @@ func (a *StartFlowAction) Validate(assets flows.Assets) error {
...
-		return fmt.Errorf("child run for flow '%s' ended in error, ending execution", a.FlowUUID)
-	}
-
-	// did we complete?
-	if run.Child().Status() != flows.StatusActive {
-		// add our exit event
-		run.ApplyEvent(step, a, events.NewFlowExitedEvent(run.Child()))
-	}
-
+	run.ApplyEvent(step, a, events.NewFlowTriggeredEvent(a.FlowUUID, run.UUID()))
```
rowanseymour: Because we've removed the recursive aspect of this method, this functionality now lives in the main session event loop. Main benefit of that is that now there are only one place where we start flows, one place where we complete flows etc.

flows/actions/call_webhook.go
```diff
@@ -53,10 +53,10 @@ func (a *WebhookAction) Execute(run flows.FlowRun, step flows.Step) error {
 	// substitute any variables in our url
 	url, err := excellent.EvaluateTemplateAsString(run.Environment(), run.Context(), a.URL)
 	if err != nil {
-		run.AddError(step, err)
+		run.AddError(step, a, err)
```
rowanseymour: Now record the action with error events, same as for other event types

flows/engine/engine.go
```diff
@@ -1,209 +0,0 @@
-package engine
```
rowanseymour: This file is gone and flow execution lives entirely in session

flows/engine/session.go
```diff
@@ -7,24 +7,36 @@ import (
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/flows/events"
 	"github.com/nyaruka/goflow/flows/runs"
+	"github.com/nyaruka/goflow/flows/waits"
 	"github.com/nyaruka/goflow/utils"
 )
 
+// used to trigger a new flow run in the event loop
+type flowTrigger struct {
```
rowanseymour: We can chat naming here - but the basic idea is that we need a place on session to communicate back to the event loop that it should switch to a new flow (or just start the main flow)

flows/events/time_wait.go
```diff
@@ -0,0 +1,35 @@
...
+// ```
+//   {
+//     "type": "time_wait",
+//     "created_on": "2006-01-02T15:04:05Z",
+//     "timeout": 300
+//   }
+// ```
+//
+// @event time_wait
+type TimeWaitEvent struct {
```
rowanseymour: This isn't used anywhere yet but I think it's a good example of why having waits check run state rather than match against a specific event, makes sense. This wait doesn't check anything and could be used with timeout=0 to have a caller commit the current session and immediately call back.

flows/runs/run.go
```diff
@@ -102,41 +100,43 @@ func (r *flowRun) SetContact(contact *flows.Contact) { r.contact = contact }
...
 
-	if !e.FromCaller() {
-		r.Session().LogEvent(s, a, e)
+	if !event.FromCaller() {
+		r.Session().LogEvent(s, action, event)
 	}
+
+	// eventEnvelope, _ := utils.EnvelopeFromTyped(event)
+	// eventJSON, _ := json.Marshal(eventEnvelope)
+	// fmt.Printf("⚡︎ in run %s: %s\n", r.UUID(), string(eventJSON))
```
rowanseymour: I'm usually against leaving cruft in files but this is just sooo useful for debugging - it logs every event applied to the session.

flows/runs/run.go
```diff
@@ -146,8 +146,21 @@ func (r *flowRun) CreateStep(node flows.Node) flows.Step {
 	r.path = append(r.path, step)
 	return step
 }
-func (r *flowRun) ClearPath() {
-	r.path = nil
+
+func (r *flowRun) PathLocation() (flows.Step, flows.Node, error) {
```
rowanseymour: Not my finest naming.... get's where the engine left off in a run

cmd/flowrunner/testdata/flows/subflow_loop_test.json
rowanseymour: Am on the fence about whether we should try to prevent the second run of the parent flow - partially because we have a bunch of old rapidpro unit tests that assume that run won't be created. Could replicate the old behaviour by having the StartFlowAction check the run ancestry to check it isn't re-entering an ancestor flow.

flows/engine/session.go
```diff
@@ -51,117 +60,302 @@ func (s *session) GetRun(uuid flows.RunUUID) (flows.FlowRun, error) {
...
-// StartFlow starts the flow for the passed in contact, returning the created FlowRun
-func (s *session) StartFlow(flowUUID flows.FlowUUID, parent flows.FlowRun, callerEvents []flows.Event) error {
-	flow, err := s.assets.GetFlow(flowUUID)
+//------------------------------------------------------------------------------------------
+// Flow execution
+//------------------------------------------------------------------------------------------
+
+// StartFlow beings execution of the given flow in this session
+func (s *session) StartFlow(flowUUID flows.FlowUUID, callerEvents []flows.Event) error {
+	flow, _ := s.Assets().GetFlow(flowUUID)
```
nicpottier: Should we ignore the error here?

flows/engine/session.go
```diff
@@ -51,117 +60,302 @@ func (s *session) GetRun(uuid flows.RunUUID) (flows.FlowRun, error) {
...
+		if err != nil {
+			return err
+		}
 
-	// no first node, nothing to do (valid but weird)
-	if len(flow.Nodes()) == 0 {
-		run.Exit(flows.StatusCompleted)
-		return nil
+		// off to the races again...
+		if err = s.continueUntilWait(waitingRun, destination, step, []flows.Event{}); err != nil {
```
nicpottier: We using this style now? :(

flows/engine/session.go
```diff
@@ -51,117 +60,302 @@ func (s *session) GetRun(uuid flows.RunUUID) (flows.FlowRun, error) {
...
 }
 
-// ResumeFlow resumes our session from the last step
-func (s *session) Resume(callerEvents []flows.Event) error {
-	// find the active run
-	run := s.ActiveRun()
-	if run == nil {
-		return utils.NewValidationErrors("session: no active run to resume")
+// resumes a run that may have been waiting or a parent paused for a child subflow
+func (s *session) resumeRun(run flows.FlowRun) (flows.NodeUUID, error) {
```
nicpottier: This could probably use a better name, pretty confusing having it be `resumeRun` and not actually continuing the run. Maybe something like `findNextDestination` or something?

flows/engine/session.go
```diff
@@ -51,117 +60,302 @@ func (s *session) GetRun(uuid flows.RunUUID) (flows.FlowRun, error) {
...
+		}
 	}
 
-	// the first event resumes the wait
-	destination, step, err := resumeNode(run, node, step, callerEvents)
-	if err != nil {
-		return err
+	// a start flow action may have triggered a subflow in which case we're done on this node for now
+	// and it will be resumed when the subflow finishes
+	if s.flowTrigger != nil {
```
nicpottier: Is it weird that we execute all actions before looking at triggers, but fail immediately if any of the actions set the status to errored? Feels to me like we should be consistent perhaps, actions are "atomic" and all are run or status is checked after each.

flows/engine/session.go
```diff
@@ -51,117 +60,302 @@ func (s *session) GetRun(uuid flows.RunUUID) (flows.FlowRun, error) {
...
+		return step, noDestination, nil
 	}
 
-	err = continueRunUntilWait(run, destination, step, nil)
-	if err != nil {
-		return err
+	// if our node has a wait before its router, we hand back to the caller
+	wait := node.Wait()
```
nicpottier: Feels to me like waits should have the opportunity to exit immediately. You could imagine a wait being conditional in that in some instances it does return back to caller and others it moves forward immediately based on some state of the run for example.

flows/waits/time.go
```diff
@@ -0,0 +1,25 @@
...
+
+func (w *TimeWait) Type() string { return TypeMsg }
+
+func (w *TimeWait) Apply(run flows.FlowRun, step flows.Step) {
+	run.ApplyEvent(step, nil, &events.TimeWaitEvent{Timeout: w.Timeout})
+}
+
+// CanResume returns true for a message wait if a message has now been received on this step
+func (w *TimeWait) CanResume(run flows.FlowRun, step flows.Step) bool {
+	return true
```
nicpottier: Shouldn't this be checking the timeout?

flows/engine/session.go
nicpottier: Ya, at first wasn't so hot about it because of the overlap with Trigger in RapidPro but I guess they are really the same thing when it comes down to it. So I'm cool with it.

flows/engine/session.go
```diff
@@ -51,117 +60,302 @@ func (s *session) GetRun(uuid flows.RunUUID) (flows.FlowRun, error) {
...
-	flow, err := s.assets.GetFlow(flowUUID)
+//------------------------------------------------------------------------------------------
+// Flow execution
+//------------------------------------------------------------------------------------------
+
+// StartFlow beings execution of the given flow in this session
+func (s *session) StartFlow(flowUUID flows.FlowUUID, callerEvents []flows.Event) error {
+	flow, _ := s.Assets().GetFlow(flowUUID)
+
+	s.Trigger(flow, nil)
```
nicpottier: I think we should be careful with the naming here. Trigger can be a noun or a verb so gets confusing in this context. Would prefer `SetTrigger`

flows/engine/session.go
```diff
@@ -35,14 +47,11 @@ func (s *session) SetEnvironment(env utils.Environment) { s.env = env }
 func (s *session) Contact() *flows.Contact              { return s.contact }
 func (s *session) SetContact(contact *flows.Contact)    { s.contact = contact }
 
-func (s *session) CreateRun(flow flows.Flow, parent flows.FlowRun) flows.FlowRun {
-	run := runs.NewRun(s, flow, s.contact, parent)
-	s.addRun(run)
-	return run
+func (s *session) Trigger(flow flows.Flow, parentRun flows.FlowRun) {
```
nicpottier: Maybe `SetTrigger` here?

flows/engine/session.go
rowanseymour: No will fix. I had played with starting the main flow with a trigger_flow event, and in that case I think this sort of check would be done at the request (assets>session>events) validation stage... but starting the main flow with an event is tricky because we don't currently support events without runs - i.e. there's no step to attach them to.

flows/engine/session.go
rowanseymour: You said it was more go-ish. I don't feel strongly but I do prefer saving a line here and there

flows/waits/time.go
rowanseymour: Maybe or could argue it's better to trust the caller's timing. If we check here we should also check message timeouts aren't sent too early.

nicpottier: A CanResume method that doesn't actually check anything sure seems goofy though. Aren't other waits going to check the state appropriately? 

Seems like the engine just returns in the same state (with the same wait) if the resume isn't satisfied no?

flows/engine/session.go
nicpottier: Only argument is consistency.. I kind of dislike it because it makes that action look different in the case where you care about the scope of one of the variables and when you don't, so to me that is the main argument against it.

flows/engine/session.go
nicpottier: What does "no will fix" mean? :)

flows/engine/session.go
rowanseymour: Well one difference there is that for an error we don't have to resume again so we can just bail out in the middle of a node. We don't currently allow resuming mid-actionset so the flow trigger can only be set by the last action, i.e. our requirement that StartFlowAction come last.

nicpottier: Ok, difference in resuming is a fair point.

flows/waits/time.go
rowanseymour: I don't think it's goofy to have a special case of a wait that is essentially a null wait - and it provides an easy mechanism to let callers commit sessions before a potentially long running action.

If we return the same wait the a caller will wait for that time period again, which isn't correct.

nicpottier: The caller should either be waiting from the creation time of the wait or we should use absolute numbers. The whole name`CanResume` implies the wait is making the decision here, it makes no sense to have it do nothing.

Does `msg_wait` also do nothing and just assume there is a `msg_received`?

flows/engine/session.go
rowanseymour: Ugh what about using it when a function only returns `err`? I really like that it scopes `err`.

nicpottier: Well it isn't scoping err in this case..

flows/waits/time.go
rowanseymour: Ok there's two different questions here:

1. Is it valid to have a wait which never waits. To me this is like a `TrueTest` which always returns true or `FalseTest` that always returns false. This is just a special case of a `Wait` which always returns true. We could call it `NeverWait` or `NullWait` or whatever - it's useful for the engine to be able to return to the caller but not actually be waiting for anything. Not sure why that's so weird.
2. Should the engine check the caller waits for timeouts. Maybe. We would need a created_on time on the wait or a timestamp. It does make me a little nervous that either of those routes run into problems if the caller and engine system time are even slightly off.

nicpottier: Definitely thing 1) is valid, though this is a TimeWait not a NullWait.

For 2) the engine has to satisfy the contract of the flow and should make sure that isn't violated regardless of what the caller does. So to me it sure seems we need to be checking these things. Agree there can be races with waits but thats the world we live in.

rowanseymour: Ok well I originally had this as a null wait but then figured if I added a `Timeout` field then we could solve two problems with one wait (timeout=0 for a null wait). I'll change it back to `NullWait` and we can solve the timeout issue for time and message waits later.

cmd/flowrunner/testdata/flows/subflow_loop_test.json
rowanseymour: @nicpottier thoughts on this? I like having only one way of detecting loops... but I don't like breaking our unit tests.

flows/engine/session.go
rowanseymour: Well then I should fix this case :) I'm flicking through the Go source right now to see where they fall on this construct and best I can tell they generally use it whenever there isn't another returned variable needed in the wider scope. I'd prefer to follow their lead for now rather than create our own rules, and they have lots of other useful ideas in their [style guide](https://github.com/golang/go/wiki/CodeReviewComments).

flows/waits/time.go
rowanseymour: Called it `NothingWait` rather than `NullWait` to avoid the potentially confusing `"type"="null"` in the JSON

flows/engine/session.go
rowanseymour: Not against that tho not sure how to implement it for `NothingWait`

PR #82, opened 2017-08-07
Event to expire a run and some misc fixes

rowanseymour:
* Adds support for a `run_expired` event which the caller can send when a parent needs to be resumed due to a child expiring
* Puts the flow loop check back in `StartFlowAction` tho this time it's implemented by looking at the run "call stack" rather than the event log
* Puts `created_on` back on flow run - I took this off previously because we need to use db time on the rapidpro side, but we need this created_on to use for the ActionLog which records starting a flow. It also makes sense to have this field on run since events, inputs etc have it.
* Fix empty event log being marshaled as "null" rather than "[]"


flows/events/run_expired.go
```diff
@@ -0,0 +1,34 @@
...
+	BaseEvent
+	RunUUID flows.RunUUID `json:"run_uuid"    validate:"required,uuid4"`
+}
+
+// Type returns the type of this event
+func (e *RunExpiredEvent) Type() string { return TypeRunExpired }
+
+// Apply applies this event to the given run
+func (e *RunExpiredEvent) Apply(run flows.FlowRun) {
+	// TODO check this is correct run
```
nicpottier: Missing?

rowanseymour: Oh yeah I meant to flag this for discussion because maybe we should aim for some consistency about what events can do in their `Apply` methods, and whether a check like "does this run exist in the session" belong in a validation stage rather than here.

nicpottier: Maybe Apply should return an optional error? Seems like that would make sense as Events are potentially input from the caller so validation seems likely to be needed.

PR #85, opened 2017-08-16
Flow stack for detecting loops

rowanseymour:
Currently the logic for avoid flow loops doesn't allow something like flow A calling into flow B twice without a wait because we'll hit the same nodes in B twice. This PR introduces a "flow stack" which tracks triggered flows and visited nodes.

### DIFF
--- a/flows/engine/session.go
+++ b/flows/engine/session.go
@@ -18,17 +18,20 @@ type flowTrigger struct {
 }
 
 type session struct {
-	assets  flows.Assets
+	assets flows.Assets
+
+	// state which is maintained between engine calls
 	env     utils.Environment
 	contact *flows.Contact
+	runs    []flows.FlowRun
+	status  flows.SessionStatus
+	wait    flows.Wait
+	log     []flows.LogEntry
 
-	runs       []flows.FlowRun
-	runsByUUID map[flows.RunUUID]flows.FlowRun
-	status     flows.SessionStatus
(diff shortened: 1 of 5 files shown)


flows/interfaces.go
```diff
@@ -218,6 +218,10 @@ type LogEntry interface {
 	Event() Event
 }
 
+type FlowStack interface {
+	HasFlow(FlowUUID) bool
```
rowanseymour: This is used by StartFlowAction... otherwise flowStack could be completely internal to the engine package

flows/engine/session.go
```diff
@@ -225,7 +232,7 @@ func (s *session) continueUntilWait(currentRun flows.FlowRun, destination flows.
 
 		// if we now have a destination, go there
 		if destination != noDestination {
-			if !visited[destination] {
+			if !s.flowStack.hasVisited(destination) {
```
nicpottier: Shouldn't we have some kind of error or log when we are stopping execution due to a loop?

rowanseymour: Scroll down 20 lines

flows/interfaces.go
nicpottier: Should we just add a method on session? Then we could get rid of this interface entirely?

flows/engine/session.go
nicpottier: Doh, yep!

PR #88, opened 2017-08-17
Assume save actions in legacy flows provide field UUIDs

rowanseymour:
Legacy flows currently reference contact fields by their key but before we try to execute flows in goflow we should ensure that fields exist and have UUIDs

### DIFF
--- a/flows/definition/legacy.go
+++ b/flows/definition/legacy.go
@@ -137,9 +137,10 @@ type legacyAction struct {
 	Variables []legacyVariable         `json:"variables"`
 
 	// save actions
-	Field string `json:"field"`
-	Value string `json:"value"`
-	Label string `json:"label"`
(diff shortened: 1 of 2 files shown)


flows/definition/legacy.go
```diff
@@ -682,41 +670,52 @@ func createRuleNode(lang utils.Language, r legacyRuleSet, translations *flowTran
...
 	node := &node{}
 
 	node.uuid = a.UUID
 	node.actions = make([]flows.Action, len(a.Actions))
 	for i := range a.Actions {
-		action, err := createAction(lang, a.Actions[i], fieldMap, translations)
-		if err == nil {
-			node.actions[i] = action
-		} else {
-			fmt.Println(err)
```
rowanseymour: Errors now properly passed back up stack

PR #90, opened 2017-08-18
Message wait timeouts 

rowanseymour:
* Timeouts move from flow run to the wait on the session
* When a wait is begun, if it has a relative `timeout` value, this becomes absolute `timeout_on` value
* Engine enforces timeouts - caller can't resume prematurely

Note that currently the caller doesn't pass anything back to the engine if it's resuming from a timeout, cuz it's not clear to me what it would pass back. The caller isn't the one who decides if the session is resumed so I'm not sure it makes sense to pass back an event like "resume". If we really want an event for logging purposes then I think it's the responsibility of the engine to generate it once the wait is satisfied and the session actually is resumed.


excellent/tests.go
```diff
@@ -264,6 +265,29 @@ func HasLegacyWebhookStatus(env utils.Environment, args ...interface{}) interfac
 	return XFalseResult
 }
 
+// HasWaitTimedOut returns whether the last wait timed out.
```
nicpottier: Wouldn't `HasTimedOutWait` be a bit more natural?

flows/waits/base.go
```diff
@@ -0,0 +1,27 @@
+package waits
+
+import (
+	"time"
+
+	"github.com/nyaruka/goflow/flows"
+)
+
+type BaseWait struct {
```
nicpottier: So the idea here is that all waits can have a timeout?

Would it make more sense to have a TimeoutWait and have classes that want to have a Timeout as well extend it? That might make it a bit saner in the use of optional fields like Timeout.

PS. Is it TimeOut or Timeout? (test is TimedOut)

flows/waits/nothing.go
```diff
@@ -8,12 +8,15 @@ import (
...
 
 type NothingWait struct {
+	BaseWait
 }
 
 func (w *NothingWait) Type() string { return TypeNothing }
 
-func (w *NothingWait) Apply(run flows.FlowRun, step flows.Step) {
-	run.ApplyEvent(step, nil, &events.NothingWaitEvent{})
+func (w *NothingWait) Begin(run flows.FlowRun, step flows.Step) {
```
nicpottier: Maybe if BaseWait did nothing then this could be TimeoutWait?

flows/waits/base.go
rowanseymour: Yeah we could do that - stops users from putting a timeout on a nothing wait.

Waits in the flow definition can have a "timeout" value and session can checked on a wait to see if it "has timed out". I think this is clear, no?

PR #91, opened 2017-08-21
Cached assets attempt 1

rowanseymour:
Reworks asset representation so that we can model embedded and lazy assets

```json
{"type": "flow", "content": {}, "fetch_url": "http://rpd.io/assets/flow/1234..."}
```

```json
{"type": "flow", "uuid": "1234...", "fetch_url": "http://rpd.io/assets/flow/1234..."}
```

Haven't yet implemented fetching of lazy assets

### DIFF
--- a/cmd/flowserver/main.go
+++ b/cmd/flowserver/main.go
@@ -13,6 +13,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/nyaruka/goflow/flows/engine"
+
+	"github.com/nyaruka/goflow/flows"
+
 	"errors"
 
 	"github.com/koding/multiconfig"
@@ -33,6 +37,7 @@ import (
 )
 
 var version = "dev"
+var assetManager flows.AssetManager
 
 func main() {
 	m := multiconfig.New()
@@ -80,6 +85,8 @@ func main() {
 
 	r.NotFound(errorHandler(http.StatusNotFound, "not found"))
 	r.MethodNotAllowed(errorHandler(http.StatusMethodNotAllowed, "method not allowed"))
+
+	assetManager = engine.NewAssetManager()
 
 	httpServer := &http.Server{
 		Addr:         fmt.Sprintf(":%d", config.Port),
--- a/cmd/flowserver/main.go
+++ b/cmd/flowserver/main.go
@@ -37,7 +37,7 @@ import (
 )
 
 var version = "dev"
-var assetManager flows.AssetManager
+var assetStore flows.AssetStore
 
 func main() {
 	m := multiconfig.New()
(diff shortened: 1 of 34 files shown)


flows/engine/assets.go
```diff
@@ -3,90 +3,142 @@ package engine
...
 	"github.com/nyaruka/goflow/utils"
 )
 
-type assets struct {
-	flows    map[flows.FlowUUID]flows.Flow
-	channels map[flows.ChannelUUID]flows.Channel
+type assetContainer struct {
+	assetType  flows.AssetType
+	asset      flows.Asset
+	accessedOn *time.Time
```
rowanseymour: I'm thinking we'll eventually evict stuff from the asset cache based on when it was last accessed

flows/engine/assets.go
```diff
@@ -3,90 +3,142 @@ package engine
...
+			m.AddLazyAsset(envelope.Type, *envelope.UUID, envelope.URL)
 		}
-		channelList[c] = channel
 	}
 
-	for f := range envelope.Flows {
-		flow, err := definition.ReadFlow(envelope.Flows[f])
-		if err != nil {
-			return nil, err
+	// any non-lazy assets can be validated now
```
rowanseymour: This is a bit weird - say someone passes in a flow and a subflow which the first flow requires. If we validate each asset before adding it, the first one will trigger another request for the subflow... so I'm adding them all to the asset cache and then validating them.

cmd/flowserver/main.go
```diff
@@ -33,6 +37,7 @@ import (
 )
 
 var version = "dev"
+var assetManager flows.AssetManager
```
rowanseymour: This asset manager will exist for the lifetime of the server and be shared by all requests

flows/engine/assets.go
```diff
@@ -3,90 +3,147 @@ package engine
...
+	fetchURL   string
 }
 
-func NewAssets(flowList []flows.Flow, channelList []flows.Channel) flows.Assets {
-	flowMap := make(map[flows.FlowUUID]flows.Flow, len(channelList))
-	for _, flow := range flowList {
-		flowMap[flow.UUID()] = flow
+type assetManager struct {
+	cache      map[flows.AssetUUID]assetContainer
+	cacheMutex sync.Mutex
```
nicpottier: I think you can use a `RWMutex` here and it'll operate a bit faster for the read cases.

flows/engine/assets.go
```diff
@@ -3,90 +3,147 @@ package engine
...
+}
+
+func NewAssetManager() flows.AssetManager {
```
nicpottier: Not passionate, but I think I prefer just `Assets` over `AssetManager`.. `Manager` always smells a bit as to me it doesn't carry a lot of meaning with it.

nicpottier:
Your comment above says `fetch_url` but it seems we use `url` is that right? (I prefer `url`)

flows/engine/assets.go
rowanseymour: AssetStore?

rowanseymour:
@nicpottier yeah I changed it to just `url` at some point

Issue #50, opened 2017-07-07
Calling @webhook.json before webhook causes error


ericnewcomer:
Flow definition: [webhook-failure.json.zip](https://github.com/nyaruka/goflow/files/1132003/webhook-failure.json.zip)

ERRO[1396] request complete                              http_method=POST http_proto="HTTP/1.1" http_scheme=http panic="runtime error: invalid memory address or nil pointer dereference" remote_addr="[::1]:52361" req_id="esquandolas.local/IAYbjAE4Qo-000007" resp_bytes_length=22 resp_elasped_ms=0.592514 resp_status=500

PR #92, opened 2017-08-22
Small fixes

rowanseymour:
Fixes https://github.com/nyaruka/goflow/issues/50 which was blowing up because `variable` is a non-nil interface... containing a nil pointer.


nicpottier:
tricksy nil pointers!

PR #96, opened 2017-08-31
Assets caching and fetching

rowanseymour:
Current limitations:
 * Single lock on all cache access
 * No pruning of old items

See also https://github.com/nyaruka/rapidpro/pull/1408


cmd/docgen/run.go
```diff
@@ -71,19 +74,20 @@ var contactDef = `
 
 func createExampleSession(assetsDef string) (flows.Session, error) {
 	// read our assets
-	assets := engine.NewAssetStore()
-	if err := assets.IncludeAssets(json.RawMessage(assetsDef)); err != nil {
+	assetCache := engine.NewAssetCache()
```
rowanseymour: `AssetCache` isn't ideal name as it's also responsible for fetching assets from the caller.. but we thus far failed to come up with anything better

cmd/flowrunner/runner_test.go
```diff
@@ -64,38 +64,46 @@ func readFile(prefix string, filename string) ([]byte, error) {
 	return bytes, err
 }
 
-func runFlow(env utils.Environment, assetsFilename string, contactFilename string, flowUUID flows.FlowUUID, callerEvents [][]flows.Event) (flows.Session, []*Output, error) {
+type runResult struct {
```
rowanseymour: Returning a tuple of 4 things was starting to get confusing

flows/contact.go
```diff
@@ -125,18 +125,18 @@ func NewContactReference(uuid ContactUUID, name string) *ContactReference {
...
-	UUID        ContactUUID    `json:"uuid"              validate:"required,uuid4"`
+	UUID        ContactUUID    `json:"uuid" validate:"required,uuid4"`
 	Name        string         `json:"name"`
 	Language    utils.Language `json:"language"`
 	Timezone    string         `json:"timezone"`
 	URNs        URNList        `json:"urns"`
 	Groups      GroupList      `json:"groups"`
 	Fields      *Fields        `json:"fields,omitempty"`
-	ChannelUUID ChannelUUID    `json:"channel_uuid"      validate:"omitempty,uuid4"`
+	ChannelUUID ChannelUUID    `json:"channel_uuid,omitempty" validate:"omitempty,uuid4"`
```
ericnewcomer: Per discussion this morning, if messages require channels, seems contacts probably should too.

flows/engine/assets.go
```diff
@@ -11,95 +13,165 @@ import (
...
 
-func (m *assetStore) AddAsset(asset flows.Asset, fetchURL string) {
-	m.cache[asset.AssetUUID()] = assetContainer{
-		assetType: asset.AssetType(),
-		asset:     asset,
-		expiresOn: time.Now().Add(5 * time.Minute),
-		fetchURL:  fetchURL,
+// fetches an asset by its URL and parses it as the provided type
+func (c *AssetCache) fetchAsset(url assetURL, aType assetType, itemType assetItemType) (interface{}, error) {
+	response, err := http.Get(string(url))
```
ericnewcomer: What about authentication for these fetches? Is the expectation that none of these are gated and not publicly addressable as a requirement?

flows/contact.go
nicpottier: What about flows that don't send messages and are just messing with contact state? Wouldn't those contacts not have a channel?

ericnewcomer: Ah yes, indeed. Nevermind.

rowanseymour: You could have a new contact that's never been contacted before, no?

flows/engine/assets.go
rowanseymour: If we're using the same endpoint for the flow editor then we need some authentication. I was figuring we'd support regular session auth for the flow editor and maybe a token based system for the flow server?

rowanseymour:
@nicpottier @ericnewcomer this is updated now so that each request gets a map of asset type -> asset base URL

cmd/docgen/run.go
nicpottier: `AssetStore`?

rowanseymour: I've gone back and forth on AssetCache vs AssetStore and @ericnewcomer said he preferred the former.

PR #98, opened 2017-09-04
Group assets and actions

rowanseymour:
**Groups as assets**

Flows can request a single group (`Group`) or the set of all groups (`GroupSet`) as assets. As an optimization, we only ever fetch the set of all groups, and use that to find individual groups.

**Groups in flows**

These are now modelled as `GroupReference`, similar to `ContactReference`. They're either an object with UUID and name, or a name value which is an expression which evaluates to an existing group (see https://github.com/rapidpro/rapidpro/issues/627).

**Groups on contacts**

These are all just UUIDs now, tho they are still loaded as group objects at session read time. This will likely have to change in the future if we want to contacts to be treated as assets sometimes (e.g. the contact of a parent run if it differs from the current contact).

**Groups on events**

Likewise these are all just UUIDs which keeps `add_to_group` and `remove_from_group` consistent with other events like `send_msg`.

### DIFF
--- a/flows/definition/legacy.go
+++ b/flows/definition/legacy.go
@@ -97,8 +97,8 @@ type legacyGroupReference struct {
 	Name string          `json:"name"`
 }
 
-func (g *legacyGroupReference) Migrate() *flows.Group {
-	return flows.NewGroup(g.UUID, g.Name)
+func (g *legacyGroupReference) Migrate() *flows.GroupReference {
+	return flows.NewGroupReference(g.UUID, g.Name)
 }
 
 type legacyVariable struct {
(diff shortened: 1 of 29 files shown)


flows/definition/legacy.go
```diff
@@ -93,12 +94,40 @@ func (c *legacyContactReference) Migrate() *flows.ContactReference {
...
+
+func (g *legacyGroupReference) Migrate() *flows.GroupReference {
+	return flows.NewGroupReference(g.UUID, g.Name)
 }
 
-func (g *legacyGroupReference) Migrate() *flows.Group {
-	return flows.NewGroup(g.UUID, g.Name)
+func (g *legacyGroupReference) UnmarshalJSON(data []byte) error {
+	// group reference may be a string
+	if data[0] == '"' {
```
rowanseymour: Not sure if there's a cleaner way to do this but this handles the case where the group is just a string instead of a dictionary.

PR #99, opened 2017-09-05
Implement add_label and add labels as an asset type

rowanseymour:
* Adds `msg_uuid` to `msg_received` events
* Adds `uuid` to Input
* If run has input, then `add_label` action generates `add_label` event with `input_uuid` and `label_uuids`

### DIFF
--- a/flows/actions/base.go
+++ b/flows/actions/base.go
@@ -58,7 +58,7 @@ func resolveGroups(run flows.FlowRun, step flows.Step, action flows.Action, refe
 	return groups, nil
 }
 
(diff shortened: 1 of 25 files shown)


flows/actions/base.go
```diff
@@ -58,3 +57,43 @@ func resolveGroups(run flows.FlowRun, step flows.Step, action flows.Action, refe
 
 	return groups, nil
 }
+
+// helper function for actions that have a set of group references that must be resolved to actual groups
```
nicpottier: comment wrong

PR #100, opened 2017-09-05
Pass step and action to Event.Apply so that events can log errors

rowanseymour:
Re-arranging the `update_contact` and `set_contact_field` actions/events before implementing dynamic groups, so we're clear about who does what..

- actions generate events...
- events modify session state...
- events can generate other events which are applied inline? (e.g. applying an `update_contact` event generates `add_to_group` and `remove_from_group` events)

One small complication about events generating events is that the order in the log will be the order they were completed in, rather than created on, e.g.

 1. `add_to_group` for a dynamic group with query `state=NY`
 2. `set_contact_field` to set `state` to `NY`

So maybe they need sorted by event.created_on

### DIFF
--- a/flows/actions/add_label.go
+++ b/flows/actions/add_label.go
@@ -41,16 +41,18 @@ func (a *AddLabelAction) Validate(assets flows.SessionAssets) error {
 }
 
 // Execute runs the labeling action
-func (a *AddLabelAction) Execute(run flows.FlowRun, step flows.Step) error {
+func (a *AddLabelAction) Execute(run flows.FlowRun, step flows.Step) ([]flows.Event, error) {
 	// only generate event if run has input
 	input := run.Input()
 	if input == nil {
-		return nil
+		return nil, nil
 	}
 
-	labels, err := resolveLabels(run, step, a, a.Labels)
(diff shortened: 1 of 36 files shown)


rowanseymour:
@nicpottier @ericnewcomer updated so that action isn't passed to `Event.Apply`. This change is now pretty much just having the update_contact and set_contact_field events rather than actions do the work of changing the session's contact.

nicpottier:
Looks good... from an architectural point of view I wonder whether we should be heading towards Actions passing back Events (which are perhaps ErrorEvents in those cases) instead of actually modifying the Run itself. Having the engine be responsible for applying events feels slightly cleaner to me. Anyways, just food for thought.

rowanseymour:
I think I agree tho I tried that a while back but it wouldn't work because StartFlowAction was recursive and needed state changes to made immediately to the session. It's possible it would work now to have Action.Execute return a list of events, and that would solve the problem of keeping events in the correct order.

rowanseymour:
@nicpottier @ericnewcomer please take another peek - have reworked actions so they return events which are applied by the engine (this wasn't possible before because StartFlowAction was recursive and needed events applied immediately)

PR #101, opened 2017-09-13
Fields

rowanseymour:


cmd/docgen/run.go
```diff
@@ -36,6 +36,14 @@ var assetsDef = `
 		}
 	},
 	{
+		"type": "field",
```
rowanseymour: Can't load the contact now without the field assets being accessible

cmd/flowrunner/runner_test.go
```diff
@@ -78,18 +79,25 @@ type runResult struct {
 }
 
 func runFlow(env utils.Environment, assetsFilename string, contactFilename string, flowUUID flows.FlowUUID, callerEvents [][]flows.Event) (runResult, error) {
-	assetsJSON, err := readFile("flows/", assetsFilename)
-
+	// load both the test specific assets and default assets
+	defaultAssetsJSON, err := readFile("", "default.json")
```
rowanseymour: Moving some assets into a default file so they don't have to be duplicated in every test

cmd/flowrunner/testdata/flows/all_actions.json
```diff
@@ -39,8 +39,10 @@
                         {
                             "uuid": "7bd8b3bf-0a3c-4928-bc46-df416e77ddf4",
                             "type": "save_contact_field",
-                            "field_uuid": "f06c5b73-eb0d-4417-b7e0-4f650ed30dc8",
-                            "field_name": "Activation Token",
+                            "field": {
```
rowanseymour: This format is more like how we reference groups/labels/contacts in flows. Key won't be used, and might be wrong, but like group name etc, makes flow more readable.

excellent/tests.go
```diff
@@ -156,7 +156,7 @@ func HasValue(env utils.Environment, args ...interface{}) interface{} {
 	}
 
 	// nil is not a value
-	if args[0] == nil {
+	if utils.IsNil(args[0]) {
```
rowanseymour: oh golang nil interfaces...

nicpottier: Ya, I wonder whether this is a sign that we are doing something wrong.

flows/actions/save_contact_field.go
```diff
@@ -16,18 +16,16 @@ const TypeSaveContactField string = "save_contact_field"
...
 //   }
 // ```
 //
 // @action save_contact_field
 type SaveContactField struct {
 	BaseAction
-	FieldUUID flows.FieldUUID `json:"field_uuid"    validate:"required,uuid4"`
-	FieldName string          `json:"field_name"    validate:"required"`
-	Value     string          `json:"value"`
+	Field *flows.FieldReference `json:"field" validate:"required"`
```
nicpottier: Why do we use a nested object for the field in the action but use `field_key` in the event? Seems these should be consistent.

rowanseymour: I think pretty much all actions currently use nested objects for any sort of reference (labels, groups, field, contacts, subflows). In the label/group/field case these are things you might create on flow import and thus need extra fields. Yesterday Eric and I were discussing the possibility of including value_type with field references in flows so that import knows to create a field with a specific type.

All our events on the otherhand use single values - uuids for groups/labels/flows/contacts and in this case, key for a field. The thing being referenced in an event is assumed to exist and won't be created if it doesn't.

Action references can also sometimes include expression templates but obviously a reference to something in an event can't. So I guess I see these as different things - we should strive to be consistent across all actions and across all events, but not across both of those things.

nicpottier: I don't like it but ok.

rowanseymour: If you think events should use the same reference objects as in the flow spec then we can discuss changing them to be like that, but for now this change is at least being consistent with how things currently are.

PR #108, opened 2017-09-14
Dynamic groups

rowanseymour:
* Excellent grammar renamed to Excellent2 to clarify that it's not quite the same as what we have in rapidpro
* Added ContactQL grammar from rapidpro
* Both grammar files live in antlr/ directory in project root along with LexUnicode.g4
* ContactQL comparison operations are a superset of those supported by rapidpro


rowanseymour:
(merged the excellent2 rename separately to make this diff saner)

PR #118, opened 2017-10-05
Object references

rowanseymour:
* Move object reference structs into their own package
* Switch `start_flow` and `set_preferred_channel` action to use reference objects for consistency with other actions


flows/actions/references.go
```diff
@@ -0,0 +1,56 @@
...
+	return &FlowReference{UUID: uuid, Name: name}
+}
+
+// FieldReference is a reference to field
+type FieldReference struct {
+	Key   flows.FieldKey `json:"key" validate:"required"`
+	Label string         `json:"label"`
+}
+
+// NewFieldReference creates a new field reference with the given UUID and key
```
nicpottier: comment

nicpottier:
Do events never use these references?

rowanseymour:
@nicpottier currently events just use the identifiers, e.g. the UUIDs or the field key.

nicpottier:
Does that feel weird to you at all? I find it slightly odd that we treat these differently. Whats our argument for not being so?

rowanseymour:
I really do see them as different - the references are part of the flow spec and I'm not sure events need to match that. For example:
* Group and label references in the spec can omit their UUID, and use an excellent template for their name value - none of those make sense in an event.
* A field reference will soon include it's data type - not something an event needs to be worried about.

rowanseymour:
@nicpottier please re-review

nicpottier:
Only bit I'm not positive about is `name_match` though I can't think of anything better.

PR #119, opened 2017-10-10
Resolving with defaults 

rowanseymour:
* Replaces `has_run_status`,  `has_webhook_status` and `has_legacy_webhook_status` with `is_string_eq` tests using status as the operand
* Migrate legacy webhook rulesets to have 2 error case exits same as their new world counterparts
* Ensure that context items are "defaulted" before being converted to primitives

### DIFF
--- a/excellent/tests.go
+++ b/excellent/tests.go
@@ -107,10 +107,11 @@ var _ utils.VariableResolver = XTestResult{}
 // Tests
 //------------------------------------------------------------------------------------------
 
-// IsStringEQ returns whether two strings are equal (case sensitive).
+// IsStringEQ returns whether two strings are equal (case sensitive). In the case that they
+// are, it will return the string as the match.
 //
 //  @(is_string_eq("foo", "foo")) -> true
-//  @(is_string_eq("foo", "foo")) -> true
+//  @(is_string_eq("foo", "FOO")) -> false
 //  @(is_string_eq("foo", "bar")) -> false
 //  @(is_string_eq("foo", " foo ")) -> false
 //  @(is_string_eq(run.status, "completed")) -> true
--- a/excellent/tests.go
+++ b/excellent/tests.go
@@ -24,12 +24,11 @@ import (
 
 // XTESTS is our mapping of the excellent test names to their actual functions
 var XTESTS = map[string]XFunction{
-	"has_error":                 HasError,
-	"has_value":                 HasValue,
-	"has_group":                 HasGroup,
-	"has_status":                HasStatus,
-	"has_legacy_webhook_status": HasLegacyWebhookStatus,
-	"has_wait_timed_out":        HasWaitTimedOut,
+	"has_error":          HasError,
+	"has_value":          HasValue,
+	"has_group":          HasGroup,
+	"has_status":         HasStatus,
+	"has_wait_timed_out": HasWaitTimedOut,
 
 	"has_phrase":      HasPhrase,
 	"has_only_phrase": HasOnlyPhrase,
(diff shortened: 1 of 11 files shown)


excellent/tests.go
```diff
@@ -183,22 +183,18 @@ func HasRunStatus(env utils.Environment, args ...interface{}) interface{} {
 		return fmt.Errorf("HAS_RUN_STATUS takes exactly two arguments, got %d", len(args))
```
rowanseymour: I've left the test names as they are for now instead of merging into a single `has_status`. Seems like there is some utility in knowing what is being tested, and potentially validating the different statuses that are applicable to runs vs webhooks

utils/conversions.go
```diff
@@ -198,6 +198,9 @@ func ToJSON(env Environment, val interface{}) (JSONFragment, error) {
 		return JSONFragment(bytes), err
 	}
 
+	// if value is a variable resolver, then use it's default
+	val = ResolveDefault(val)
```
rowanseymour: By having this come first - we ensure that we always stringify the default

utils/http.go
```diff
@@ -97,7 +97,7 @@ func (r *RequestResponse) Resolve(key string) interface{} {
 }
 
 func (r *RequestResponse) Default() interface{} {
-	return r.JSON()
+	return r
```
rowanseymour: I think we've talked about having JSON serialization be explicit, via something like a `JSON()` method

utils/resolver_test.go
```diff
@@ -25,3 +28,64 @@ func TestSnakify(t *testing.T) {
...
+	switch key {
+	case "child":
+		return p.child
+	case "bar":
+		return p.bar
+	}
+	return fmt.Errorf("no such field on parent")
+}
+
+func (p *testResolvableParent) Default() interface{} {
```
rowanseymour: I've now convinced myself that Default is the right name for this afterall since it will usually return something that isn't a primitive.

excellent/tests.go
```diff
@@ -109,6 +107,37 @@ var _ utils.VariableResolver = XTestResult{}
 // Tests
 //------------------------------------------------------------------------------------------
 
+// IsStringEQ returns whether two strings are equal (case sensitive).
+//
+//  @(is_string_eq("foo", "foo")) -> true
+//  @(is_string_eq("foo", "foo")) -> true
```
nicpottier: repeat?

nicpottier: Sometimes nice to define what the TestResult will return in these examples as well. (I would guess it is the matched string)

utils/conversions.go
nicpottier: Is that right? If someone defines a `String` shouldn't that take precedence over `Default`? (do we have examples when we want those to be different?)

utils/conversions.go
```diff
@@ -285,6 +279,9 @@ func ToString(env Environment, val interface{}) (string, error) {
 		return "", nil
 	}
 
+	// if value is a variable resolver, then use it's default
+	val = ResolveDefault(val)
+
```
nicpottier: As above I find this confusing.. why `Default` take precedence over `String` when trying to Stringify something?

utils/conversions.go
rowanseymour: I think so - tho it does have the odd effect of meaning `String()` is never called on context items with a `Default()` that returns something other than themselves.

`@contact.field.age` defaults to `@contact.field.age.value`, so
`@(contact.field.age == 30)` should resolve `contact.field.age` as a decimal, i.e. `Default()` then `ToDecimal`
`@(CONCAT(@contact.field.age, "years"))` should resolve `contact.field.age` as a string, and should be consistent with `@(CONCAT(@contact.field.age.value, "years"))`

Resolving the default before converting to a primitive means we don't have two different `String` methods that need to be kept consistent.

rowanseymour:
@nicpottier have reverted the conversion/default changes so this PR is now just the changes to the xtests

PR #121, opened 2017-10-11
Triggers

rowanseymour:
* Adds triggers which describe why the session is being created and which flow to start
* Trigger is part of the serialized session 
* Two trigger types defined so far:
   - `user` - a manual flow start
   - `flow_action` - a session_start action in a run in another session
* A flow_action trigger contains a summary of the run from the triggering session which is then accessible in the context as `parent`
* Adds `start_session` action (the non-subflow equivalent of `start_flow`)
* Adds `session_triggered` event (the non-subflow equivalent of `flow_triggered`)


cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -285,7 +337,6 @@
         },
         "runs": [
           {
-            "child_uuid": "",
```
rowanseymour: Current child is now calculated as needed and not serialized

flows/engine/session.go
```diff
@@ -46,15 +48,17 @@ func NewSession(cache *AssetCache, assetURLs map[AssetItemType]string) flows.Ses
...
+func (s *session) SetEnvironment(env utils.Environment) { s.env = env }
+func (s *session) Trigger() flows.Trigger               { return s.trigger }
+func (s *session) Contact() *flows.Contact              { return s.contact }
+func (s *session) SetContact(contact *flows.Contact)    { s.contact = contact }
+
 func (s *session) FlowOnStack(flowUUID flows.FlowUUID) bool { return s.flowStack.hasFlow(flowUUID) }
 
-func (s *session) SetTrigger(flow flows.Flow, parentRun flows.FlowRun) {
-	s.flowTrigger = &flowTrigger{flow: flow, parentRun: parentRun}
+func (s *session) PushFlow(flow flows.Flow, parentRun flows.FlowRun) {
```
rowanseymour: Now that trigger means something more like triggers in rapidpro, need different name for this

flows/engine/session.go
```diff
@@ -429,12 +450,6 @@ func ReadSession(cache *AssetCache, assetURLs map[AssetItemType]string, data jso
 		s.addRun(run)
 	}
 
-	// once all runs are read, we can resolve references between runs
-	err = runs.ResolveReferences(s, s.Runs())
```
rowanseymour: Now that we're not worried about children, this isn't necessary because a run's parent will always have been loaded before it - runs in the session are chronological.

flows/actions/start_session.go
```diff
@@ -0,0 +1,57 @@
+package actions
+
+import (
+	"encoding/json"
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/flows/events"
+)
+
+// TypeStartSession is the type of our flow action
```
nicpottier: comment

flows/engine/session.go
nicpottier: NextFlow (and `SetNextFlow`?)
PendingFlow?
BatterUpFlow?

flows/triggers/user.go
```diff
@@ -0,0 +1,62 @@
...
+	"time"
+
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/utils"
+)
+
+// TypeUser is a constant for incoming messages
+const TypeUser string = "user"
+
+// UserTrigger is used when a session was triggered manually by a user
```
nicpottier: I think having this mention "user" probably doesn't make sense. Seems this is more just a "manual" or "explicit" trigger.. outside systems may not have any concept of a user, nor does this describe the user in any way, so the intrinsic piece to this object is not the user but that it is manual.

rowanseymour:
@nicpottier renamed

PR #122, opened 2017-10-12
Message attachments can contain expressions

rowanseymour:
Matches recently added functionality in RP

### DIFF
--- a/flows/actions/add_urn.go
+++ b/flows/actions/add_urn.go
@@ -40,7 +40,7 @@ func (a *AddURNAction) Validate(assets flows.SessionAssets) error {
 }
 
 // Execute runs the labeling action
-func (a *AddURNAction) Execute(run flows.FlowRun, step flows.Step, log flows.ActionLog) error {
+func (a *AddURNAction) Execute(run flows.FlowRun, step flows.Step, log flows.EventLog) error {
 	// only generate event if run has a contact
 	contact := run.Contact()
 	if contact == nil {
--- a/flows/actions/add_urn.go
+++ b/flows/actions/add_urn.go
@@ -40,26 +40,29 @@ func (a *AddURNAction) Validate(assets flows.SessionAssets) error {
 }
 
 // Execute runs the labeling action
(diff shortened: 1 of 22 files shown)


rowanseymour:
@ericnewcomer updated as discussed and also added BaseMsgAction class so that ReplyAction and SendMsgAction can share text/attachment evaluation functionality.

flows/actions/base.go
```diff
@@ -42,12 +42,12 @@ func resolveGroups(run flows.FlowRun, step flows.Step, action flows.Action, refe
...
 			evaluatedGroupName, err := excellent.EvaluateTemplateAsString(run.Environment(), run.Context(), ref.NameMatch)
 			if err != nil {
-				log = append(log, events.NewErrorEvent(err))
+				*log = append(*log, events.NewErrorEvent(err))
 			} else {
 				// look up the set of all groups to see if such a group exists
 				group = groupSet.FindByName(evaluatedGroupName)
 				if group == nil {
-					log = append(log, events.NewErrorEvent(fmt.Errorf("no such group with name '%s'", evaluatedGroupName)))
+					*log = append(*log, events.NewErrorEvent(fmt.Errorf("no such group with name '%s'", evaluatedGroupName)))
```
nicpottier: These always feel a bit dirty to me, I'd rather see us return the log than passing in pointers to lists. I think that's a bit more typical in go anyways.

flows/actions/add_urn.go
```diff
@@ -40,26 +40,29 @@ func (a *AddURNAction) Validate(assets flows.SessionAssets) error {
 }
 
 // Execute runs the labeling action
-func (a *AddURNAction) Execute(run flows.FlowRun, step flows.Step) ([]flows.Event, error) {
+func (a *AddURNAction) Execute(run flows.FlowRun, step flows.Step, log flows.ActionLog) error {
```
rowanseymour: So this is less FP... but it does kinda make it easier to ensure you're not forgetting to return some events @nicpottier ?

PR #123, opened 2017-10-16
Better errors part 1

rowanseymour:
* Do proper validation of flow actions
* Improve validation level error messages
* Move flow structural validation into `ReadFlow`

Part 2 will sort out how errors are sent back the caller


cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -36,7 +36,6 @@
           "step_uuid": ""
         },
         {
-          "action_uuid": null,
```
rowanseymour: This is now omitted when empty

flows/engine/log.go
```diff
@@ -27,18 +27,19 @@ func (s *logEntry) Event() flows.Event   { return s.event }
 //------------------------------------------------------------------------------------------
 
 type logEntryEnvelope struct {
-	StepUUID   flows.StepUUID       `json:"step_uuid"   validate:"required"`
-	ActionUUID *flows.ActionUUID    `json:"action_uuid"`
-	Event      *utils.TypedEnvelope `json:"event"       validate:"required"`
+	StepUUID   flows.StepUUID       `json:"step_uuid,omitempty"`
```
rowanseymour: Making this optional so that we can error sessions in some cases when we can't resume then properly

PR #124, opened 2017-10-16
Better errors part 2

rowanseymour:
Reworks error handling so that we return the following:

200 with session marked as errored, when:
 * Resume from node that no longer exists
 * Resume parent run from node that no longer exists
 * Runtime detection of missing assets
 * Flow loop detected

400 with errors object, when:
 * Start/resume with structurally invalid flow
 * Start/resume with flow with missing assets
 * Trying to resume an already completed/errored session

500, when:
* Something has gone horribly wrong. Panic.

Also in this PR:
* Encapsulates the flowserver in a struct so it's easier to create instances for testing purposes
* Removes contact from serialized runs as it will always be the session contact
* Adds other parts of the testify library that we also use in Courier


cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -369,7 +369,6 @@
         },
         "runs": [
           {
-            "contact_uuid": "ba96bf7f-bc2a-4873-a7c7-254d1927c4e3",
```
rowanseymour: We've been ignoring this anyway and just using the session contact

cmd/flowserver/expression.go
```diff
@@ -1,50 +0,0 @@
-package main
```
rowanseymour: Because the flow server is now a struct.. feels like it's handler functions should all be in the same file

cmd/flowserver/main.go
```diff
@@ -157,7 +95,7 @@ func jsonHandler(handler jsonHandlerFunc) http.HandlerFunc {
 		r = r.WithContext(context.WithValue(r.Context(), contextStart, time.Now()))
 		value, err := handler(w, r)
 		if err != nil {
-			writeError(w, r, http.StatusInternalServerError, err)
+			writeError(w, r, http.StatusBadRequest, err)
```
rowanseymour: Now any error is returned as a 400. `utils.ValidationErrors` is only really used to combine multiple field errors into a single error, and probably should never be returned directly since it always lack context, i.e. what object are those failures on.

cmd/flowserver/expression.go
nicpottier: Really? why? Seems breaking it out gives you smaller files and logical seperation.

rowanseymour: I'm usually all for breaking things out but in this case it seems easier to find the server struct and its functions in a single server.go file

PR #126, opened 2017-10-18
Fix waits so that they can timeout

rowanseymour:


flows/engine/session.go
```diff
@@ -157,8 +157,16 @@ func (s *session) Resume(callerEvents []flows.Event) error {
...
-		if s.wait.CanResume(waitingRun, step) {
-			waitingRun.SetStatus(flows.RunStatusActive)
+		waitCanResume := s.wait.CanResume(waitingRun, step)
+		waitHasTimedOut := s.wait.HasTimedOut()
+
+		if waitCanResume || waitHasTimedOut {
+			if waitCanResume {
+				s.wait.Resume(waitingRun)
+			} else {
+				s.wait.ResumeByTimeOut(waitingRun)
```
rowanseymour: I think only a msg wait should clear the input if it times out - so making waits handle timing out themselves.

PR #128, opened 2017-10-19
Support @contact.first_name

rowanseymour:
Thought about requiring `@(first_word(contact.name))` but seems like this is such a common case that we should make it easy?


excellent/legacy.go
```diff
@@ -129,7 +129,6 @@ func newVars() vars {
 
 	contactVars := map[string]interface{}{
 		"first_name":  "first_name",
-		"last_name":   "last_name",
```
rowanseymour: This has never been a thing

Issue #127, opened 2017-10-18
Asset server authentication


rowanseymour:
Assert server endpoint in RapidPro should probably be setup to support token auth as well as regular session auth - that way it can be easily used by the flow editor as well as goflow.

Goflow then either needs to be configured with a token, or the token(s) are included in flowserver requests from RapidPro.

Goflow currently doesn't apply any restrictions to which asset URLs it will fetch from so definitely can't have it sending auth tokens off to any URL.

rowanseymour:
Added in #129

PR #129, opened 2017-10-20
Asset server authentication

rowanseymour:
Addresses #127. Flow server requests now take an asset_server component which configures authentication as well the asset type URLs, e.g.

```json
"asset_server": {
  "auth_header": "Token 12345678901234567890",
  "type_urls": {
    "flow": "http://testserver/assets/flow",
    "group": "http://testserver/assets/group"
  }
}
```

Have also tried to simplify the cache+server asset stuff and have added a mocked asset server to enable better testing.

### DIFF
--- a/cmd/flowserver/server.go
+++ b/cmd/flowserver/server.go
@@ -123,10 +123,10 @@ func (r *sessionResponse) MarshalJSON() ([]byte, error) {
 }
 
 type startRequest struct {
-	Assets    *json.RawMessage       `json:"assets"`
-	AssetURLs engine.AssetTypeURLs   `json:"asset_urls" validate:"required"`
-	Trigger   *utils.TypedEnvelope   `json:"trigger" validate:"required"`
-	Events    []*utils.TypedEnvelope `json:"events"`
+	Assets      *json.RawMessage       `json:"assets"`
+	AssetServer *engine.AssetServer    `json:"asset_server" validate:"required"`
+	Trigger     *utils.TypedEnvelope   `json:"trigger" validate:"required"`
+	Events      []*utils.TypedEnvelope `json:"events"`
 }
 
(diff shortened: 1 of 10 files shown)


cmd/flowserver/server.go
```diff
@@ -205,13 +211,21 @@ func (s *FlowServer) handleResume(w http.ResponseWriter, r *http.Request) (inter
 		return nil, err
 	}
 
-	// read and validate our assets
-	if err = s.assetCache.Include(resume.Assets); err != nil {
+	// include any embedded assets
+	if resume.Assets != nil {
```
rowanseymour: embedded assets should always be optional

flows/engine/assets.go
```diff
@@ -88,50 +99,158 @@ func (c *AssetCache) getAsset(url assetURL, aType assetType, itemType AssetItemT
...
-func (c *AssetCache) fetchAsset(url assetURL, aType assetType, itemType AssetItemType) (interface{}, error) {
-	response, err := http.Get(string(url))
+func (s *assetServer) fetchAsset(url string, itemType assetType, isSet bool) (interface{}, error) {
+	request, err := http.NewRequest("GET", string(url), nil)
 	if err != nil {
 		return nil, err
 	}
+
+	// set request headers
+	request.Header.Set("User-Agent", "flowserver/1.0")
```
nicpottier: See courier for a way of dynamically setting this to the tagged release.

nicpottier:
Seems good, wonder if this level of abstraction is the right one vs say just having different backends the flow server can know about which define their own ways of getting flows/contacts/groups etc.. We are trying to be fancy by saying you can configure these different URLs but the reality is that the implementation of those endpoints is so specific that you don't really have much abstraction.

Nothing worth changing, just we should be realistic that this isn't really an abstraction, just a RapidPro asset store.

PR #131, opened 2017-10-24
Migrating VariableContactActions

rowanseymour:
- Rework send_msg and start_session actions so they support legacy expressions as a workaround for now
- Infer types (group name vs contact UUID) at action execution time
- Improve error messages on reference types

### DIFF
--- a/flows/actions/base.go
+++ b/flows/actions/base.go
@@ -3,10 +3,10 @@ package actions
 import (
 	"fmt"
 
-	"github.com/nyaruka/goflow/flows/events"
-
+	"github.com/nyaruka/gocommon/urns"
 	"github.com/nyaruka/goflow/excellent"
 	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/flows/events"
 )
 
 type eventLog struct {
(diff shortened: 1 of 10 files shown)


flows/actions/base.go
```diff
@@ -139,30 +139,22 @@ func (a *BaseAction) resolveLabels(run flows.FlowRun, step flows.Step, reference
 	return labels, nil
 }
 
-// BaseMsgAction is our base action for message that generate message events
-type BaseMsgAction struct {
-	BaseAction
+// MsgAction is our mixin for an action that generates message events
+type MsgAction struct {
```
rowanseymour: golang doesn't really do mixins but not sure what else you'd call this. Two of our actions (`reply`, `send_msg`) need to share some message creating functionality, two of them (`send_msg`, `start_session`) need to share some contact/group/urn resolving functionality.

flows/definition/testdata/migrations/actions.json
```diff
@@ -279,5 +283,63 @@
...
+            "groups": [
+                {"uuid": "1a8c1313-4126-41ad-80ac-f9a2d0b708ec", "name": "Testers"},
+                {"uuid": "c746a9c9-8b31-49d2-aa55-a515d2c02968", "name": "Spammers"}
+            ],
+            "legacy_vars": [
+                "@contact.fields.support_group",
+                "+12223334444"
+            ],
+            "flow": {"uuid": "879ace1b-740b-45f1-9198-c2f2f08a825f", "name": "Registration"},
+            "empty_contact": true
```
rowanseymour: discussed with @ericnewcomer and we felt this was cleaner than making it a special type of contact reference

flows/references.go
```diff
@@ -93,18 +93,26 @@ func NewVariableLabelReference(nameMatch string) *LabelReference {
...
 //------------------------------------------------------------------------------------------
 
+// GroupReferenceValidation validates that the given group reference is either a concrete
+// reference or a name matcher
 func GroupReferenceValidation(sl validator.StructLevel) {
 	ref := sl.Current().Interface().(GroupReference)
-	if len(ref.UUID) != 0 && len(ref.NameMatch) != 0 {
-		sl.ReportError(ref.UUID, "UUID", "uuid", "uuidornamematch", "")
-		sl.ReportError(ref.NameMatch, "NameMatch", "name_match", "uuidornamematch", "")
+	if !xor(string(ref.UUID), ref.NameMatch) {
```
nicpottier: `if (ref.UUID && ref.NameMatch) || (!ref.UUID && !ref.NameMatch)` ?

Not sure that's that much better, maybe we just need to be smarter? Or have a utility function which is `NeitherOrBoth`?

flows/actions/base.go
nicpottier: I think my bias would be that we duplicate the variable declarations (ie the serialization format) for the subactions and just turn these into utility methods. I don't like serialization formats being done through encapsulation / inheritance, I'd rather those be explicit, and it seems simple / clearer to just have utility methods for the resolving. I would guess that is slightly more Go-ish in that they love repeating themselves.

rowanseymour: I'm fine with that - that's how resolveLabels / resolveGroups currently work already

flows/references.go
rowanseymour: strings don't have implicit boolean values so you have to wrap everything in len(..) > 0... so I think a utility function is still helpful here

rowanseymour:
@nicpottier updated

PR #133, opened 2017-10-30
Results

rowanseymour:
Updates results to include everything we need to save results in rapidpro


rowanseymour:
@nicpottier tweaked the two_questions test so that input isn't the same as the value - also the date parsing examples where value is an ISO date

nicpottier:
Which reminds me, should we be adding `category_localized` here? Seems this is the place to do it as opposed to making the caller figure it out.

PR #137, opened 2017-11-21
Update vendors

norkans7:


rowanseymour:
Little confused - phonenumbers previously wasn't a direct dependency of goflow - it was in gocommon/vendor - now it's in the goflow/vendor folder and gocommon/vendor is gone - is that how vendoring is supposed to work?

nicpottier:
Yes I think that's right actually.

PR #142, opened 2018-01-25
Add support for quick replies on reply actions

rowanseymour:


rowanseymour:
@nicpottier gonna go get my 🍢 stand... didn't even think about making the migrated action work. Tho is added now.

Issue #63, opened 2017-07-25
Friendly formatting of URNs/channel addresses


rowanseymour:
In the old world we populate the context with `@channel.tel` and `@channel.tel_e164`. In new world users have access to contact URN paths and so far `@run.input.channel.address` / `@contact.channel.address` which is E164 formatted.

So how should we let users get a friendlier version of telephone numbers? Could be an attribute like:

 * `@contact.urns.1.path.display`
 * `@run.input.channel.address.display`

 or a function like:

 * `@(display(contact.urns.1.path))`
 * `@(display(run.input.channel.address)`

nicpottier:
Just a quick note that we resolve things like `@contact.urns` to the first URN a string context, so it would actually be `@contact.urns.path`.. we also resolve `@contact.urns.tel` to the first telephone URN. You can also use `@contact.urn` and just get the highest priority URN, so could be `@contact.urn.path`.

But ya, so for friendly things, maybe it is something more like `@(tel(contact.urn))`? No need to pass in the full path if it knows it can be passed an URN.. though it could back down to a string representation if not to let people use the function on their own numbers.

rowanseymour:
I like the simplicity of `tel(...)` - I'd assume if you give it something that's not a telephone number it would just return it as is? e.g. `@(tel("bob@example.com"))`

rowanseymour:
See https://github.com/rapidpro/rapidpro/pull/608

rowanseymour:
Revisiting this. Given a contact like:

```json
{
  "name": "Billy Bob",
  "urns": ["tel:+250781234567", "twitter:134252511151#billy_bob"]
}
```

I think the following expressions should evaluate as shown:

```
@contact.urns -> "+250781234567" // default URN
@contact.urns.scheme -> "tel"
@contact.urns.path -> "+250781234567"
@contact.urns.urn -> "tel:+250781234567"

@contact.urns.tel -> "+250781234567" // first tel URN if there are multiple
@contact.urns.tel.scheme -> "tel"
@contact.urns.tel.path -> "+250781234567"
@contact.urns.tel.display -> "+250781234567"
@contact.urns.tel.urn -> "tel:+250781234567"

@contact.urns.twitter -> "billy_bob"
@contact.urns.twitter.scheme -> "twitter"
@contact.urns.twitter.path -> "134252511151"
@contact.urns.twitter.display -> "billy_bob"
@contact.urns.twitter.urn -> "twitter:134252511151#billy_bob"

[...]

PR #145, opened 2018-01-26
URN expressions

rowanseymour:
See #63

### DIFF
--- a/excellent/evaluator.go
+++ b/excellent/evaluator.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"net/url"
 	"strconv"
 	"strings"
 	"unicode"
@@ -255,7 +256,7 @@ func EvaluateTemplate(env utils.Environment, resolver utils.VariableResolver, te
 
 	// if we had one, then just return our string evaluation strategy
 	if nextTT != EOF {
-		return EvaluateTemplateAsString(env, resolver, template)
+		return EvaluateTemplateAsString(env, resolver, template, false)
 	}
 
 	switch tokenType {
@@ -289,11 +290,11 @@ func EvaluateTemplate(env utils.Environment, resolver utils.VariableResolver, te
 	}
 
 	// different type of token, return the string representation
-	return EvaluateTemplateAsString(env, resolver, template)
+	return EvaluateTemplateAsString(env, resolver, template, false)
 }
 
 // EvaluateTemplateAsString evaluates the passed in template returning the string value of its execution
-func EvaluateTemplateAsString(env utils.Environment, resolver utils.VariableResolver, template string) (string, error) {
+func EvaluateTemplateAsString(env utils.Environment, resolver utils.VariableResolver, template string, urlEncode bool) (string, error) {
 	var buf bytes.Buffer
 	var errors TemplateErrors
 	scanner := newXScanner(strings.NewReader(template))
@@ -317,6 +318,10 @@ func EvaluateTemplateAsString(env utils.Environment, resolver utils.VariableReso
 				buf.WriteString(token)
 			} else {
 				strValue, _ := utils.ToString(env, value)
+				if urlEncode {
+					strValue = url.QueryEscape(strValue)
+				}
+
 				buf.WriteString(strValue)
 			}
 		case EXPRESSION:
@@ -325,6 +330,10 @@ func EvaluateTemplateAsString(env utils.Environment, resolver utils.VariableReso
 				errors = append(errors, err)
 			} else {
 				strValue, _ := utils.ToString(env, value)
+				if urlEncode {
+					strValue = url.QueryEscape(strValue)
+				}
+
 				buf.WriteString(strValue)
 			}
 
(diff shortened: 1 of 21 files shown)


excellent/evaluator.go
```diff
@@ -317,6 +318,10 @@ func EvaluateTemplateAsString(env utils.Environment, resolver utils.VariableReso
 				buf.WriteString(token)
 			} else {
 				strValue, _ := utils.ToString(env, value)
+				if urlEncode {
```
rowanseymour: same as in old engine - need this for when the expressions should be URL encoded, but the template itself shouldn't

excellent/legacy.go
```diff
@@ -39,41 +41,56 @@ func (v vars) String() string {
 	return fmt.Sprintf("%s", v["__default__"])
 }
 
-type arbitraryVars struct {
-	vars       vars
-	base       string
-	nesting    string
-	nestedVars vars
+type varMapper struct {
+	substitutions    map[string]string
```
rowanseymour: these are things which get totally rewritten - e.g. becomes a function call

excellent/legacy.go
```diff
@@ -191,13 +204,15 @@ func newVars() vars {
...
-			nesting: "json",
+		"extra": varMapper{
+			base:             "run.webhook",
+			arbitraryNesting: "json",
 		},
-	})
+	}
 }
 
+var rootVarMapper = newRootVarMapper()
```
rowanseymour: avoids re-initializing this for every expression

excellent/legacy.go
```diff
@@ -247,15 +271,15 @@ func migrateLegacyTemplateAsString(resolver utils.VariableResolver, template str
 			} else {
 				strValue, _ := toString(value)
 
-				// date context references get rewritten as functions
```
rowanseymour: so do other things now - hence `wrapRawExpression`

rowanseymour:
@nicpottier please re-review - format_urn now accepts a list

excellent/functions.go
```diff
@@ -1610,6 +1613,44 @@ func Now(env utils.Environment, args ...interface{}) interface{} {
...
+		return fmt.Errorf("FORMAT_URN takes one argument, got %d", len(args))
+	}
+
+	// if we've been passed a slice like a URNList, use first item
+	urnArg := args[0]
+	if utils.IsSlice(urnArg) {
+		sliceLen, _ := utils.SliceLength(urnArg)
+		if sliceLen >= 1 {
+			urnArg, _ = utils.LookupIndex(urnArg, 0)
+		}
```
nicpottier: Maybe 0 case should be explicitly handled here? Either as an error or an empty string? (I lean towards empty string though could be easily swayed to it being an error)

excellent/functions.go
```diff
@@ -1610,6 +1613,44 @@ func Now(env utils.Environment, args ...interface{}) interface{} {
...
 
+//----------------------------------------------------------------------------------------
+// URN Functions
+//----------------------------------------------------------------------------------------
+
+// FormatURN turns `urn` into a human friendly string
+//
+//   @(format_urn("tel:+250781234567")) -> 0781 234 567
+//   @(format_urn("twitter:134252511151#billy_bob")) -> billy_bob
+//   @(format_urn("NOT URN")) -> ERROR
```
nicpottier: should we add a `@(format_urn(contact.urns))` example?

excellent/legacy_test.go
```diff
@@ -17,12 +17,12 @@ func TestTranslate(t *testing.T) {
...
-		{old: "@contact.telegram", new: "@contact.urns.telegram"},
-		{old: "@contact.twitter", new: "@contact.urns.twitter"},
-		{old: "@contact.facebook", new: "@contact.urns.facebook"},
-		{old: "@contact.mailto", new: "@contact.urns.mailto"},
+		{old: "@contact.tel", new: "@(format_urn(contact.urns.tel.0))"},
+		{old: "@contact.tel_e164", new: "@contact.urns.tel.0.path"},
+		{old: "@contact.telegram", new: "@(format_urn(contact.urns.telegram.0))"},
+		{old: "@contact.twitter", new: "@(format_urn(contact.urns.twitter.0))"},
+		{old: "@contact.facebook", new: "@(format_urn(contact.urns.facebook.0))"},
+		{old: "@contact.mailto", new: "@(format_urn(contact.urns.mailto.0))"},
```
nicpottier: Do we need the `.0`s here?

rowanseymour: No but I think better the users see that these are lists and we treat passing a list to `format_urn` as a convenience.

nicpottier: Hah, funny, feels the opposite to me. Since 99% of users will never have more than one URN, present them with the simplest option that works.

rowanseymour: Well I guess I'd be ok with making this look simpler. I actually wonder tho how common `format_urn` usage is going to be since I imagine most of the time people want to reference contact URNs, it's not to display them in a message, but to pass them to a webhook or something like that

excellent/functions.go
rowanseymour: I prefer empty string so that `@contact.urns` and `@(format_urn(contact.urns))` both evaluate to empty (when stringified)

PR #146, opened 2018-02-02
Tweaks to match RapidPro

rowanseymour:
* Allow blank values in update_contact actions (e.g. clear language)
* Migrate old save action with field=tel_e164 to a new add_urn action
* Remove new lines from email subjects
* Some extra whitespace trimming

### DIFF
--- a/flows/events/update_contact.go
+++ b/flows/events/update_contact.go
@@ -44,12 +44,15 @@ func (e *UpdateContactEvent) Apply(run flows.FlowRun) error {
 	if e.FieldName == "name" {
 		run.Contact().SetName(e.Value)
 	} else {
-		lang, err := utils.ParseLanguage(e.Value)
-		if err != nil {
-			return err
+		if e.Value != "" {
+			lang, err := utils.ParseLanguage(e.Value)
+			if err != nil {
+				return err
+			}
+			run.Contact().SetLanguage(lang)
+		} else {
+			run.Contact().SetLanguage(utils.NilLanguage)
 		}
-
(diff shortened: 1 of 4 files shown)


flows/events/update_contact.go
```diff
@@ -44,12 +44,15 @@ func (e *UpdateContactEvent) Apply(run flows.FlowRun) error {
...
-		if err != nil {
-			return err
+		if e.Value != "" {
+			lang, err := utils.ParseLanguage(e.Value)
+			if err != nil {
+				return err
+			}
+			run.Contact().SetLanguage(lang)
+		} else {
+			run.Contact().SetLanguage(utils.NilLanguage)
```
ericnewcomer: This one feels pretty weird, maybe we should be looking at changing RP behavior instead?

rowanseymour: This is an update_contact action with field=language value="" to clear the language. Doesn't so weird to me. What would you prefer for doing that?

rowanseymour: To be clear this isn't the invalid language case - we won't even run a flow with a nonsense language in the new world

ericnewcomer: Ya, it's the bogus language case I was thinking about. I thought this was sort of capturing that behavior with ParseLanguage, but I see that case is actually bubbling up an error and this is only the empty evaluated value.

rowanseymour: well we wouldn't even get to that because the Validate method also calls ParseLanguage -
 we'd blow up with a validation error before we even start a session

PR #151, opened 2018-02-27
Events refactor part 1

rowanseymour:
Renames events to be _noun_verb_ where verb is past tense. Actions describe commands, and events should describe state changes that have happened, and need to be communicated between caller and engine. 

Contact related events:

 * `add_to_groups` &rarr; `groups_added`
 * `add_urn` &rarr; `urn_added`
 * `remove_from_groups` &rarr; `groups_removed`
 * `save_contact_field` &rarr; `contactfield_changed` (single contact field value)
 * `update_contact` &rarr; `contactproperty_changed` (single name/language/timezone property)
 * `set_contact` &rarr; `contact_changed` (entire contact - typical use would be caller knowing that something about the contact has changed but not knowing what specifically)

Non-contact related events:

 * `add_label` &rarr; `labels_added`
 * `save_flow_result` &rarr; `result_changed`
 * `send_email` &rarr; `email_sent`
 * `send_msg` &rarr; `msg_sent`
 * `set_environment` &rarr; `environment_changed`

Changes for a subsequent PR:
* `set_extra` will be replaced by immutable parameters on the session trigger, accessible in a flow as `trigger.params` ?
* `set_environment` and `set_contact` will no longer be sent to initialize a session, but like flow, they will be properties of the session trigger. Their renamed events (`contact_changed` and `environment_changed`), would only be sent by the caller if a change is detected when resuming a session.


rowanseymour:
@ericnewcomer not surprisingly on a PR all about choosing names for things we have a few  disagreements! But I'll let @nicpottier explain his disagreements

ericnewcomer:
But has the email really been sent? :)

rowanseymour:
@ericnewcomer from the POV of the engine.. sorta yeah. Admittedly it's a bit weird, but having `msg_sent` and `email_sent` be bit weird seems preferable to the current lack of consistency

nicpottier:
Ya, Rowan and I were debating on the `contact_changed` side of things.

It is subtle, but in the engine, we only record events when things actually change. IE, we won't create a `group_added` event if you are already part of a group, and we won't create a `contactfield_changed` (sure seems like we should have an underscore there btw) if the field already has that value. To me, that means `_changed` means this is a diff, so I don't like `contact_changed` because it implies this isn't the new state, but rather a set of things to update. (Rowan views this as the contact as a whole being changed, not the parts inside)

I would prefer `contact_set` for that reason, in that it is unambiguous that we are setting everything, not representing just diffs. (I think it would also allow us to use the more natural `contact_changed` for property changes like language, name, timezone)

ericnewcomer:
The engine doesn't operate in a vacuum though. These are events to be processed by a caller who is getting an event that says email_sent. That is more than a bit weird. I'm not necessarily arguing against this, just pointing out something that smells. It feels like anything the engine expects the outside world to do is more like a command than something that occurred in the past to me.

nicpottier:
I actually would support moving sending of emails to inside the engine, emails are now sent using org specific attributes anyways, so there's no reason for Django to do that. That would also fit the webhook case better.

ericnewcomer:
Still leaves message sending sadly, but ya, that's not something I feel strongly on, I was just trying to guess what the primary objection might be :)

I definitely concur on `changed` being a diff and `set` being a state declaration.

nicpottier:
Ya, msgs are a bit different, though geez, maybe engine should do that too somehow? :)

ericnewcomer:
What if we just put everything in there 🤔

nicpottier:
I feel you that it is a bit weird to have them be in the past despite nothing having actually happened. Maybe it can be something like `msg_created` which then doesn't imply that the message was actually sent. (I'm also bothered by us having both `_sent` and `_received`) as opposed to that being a single event. Could we get away with a `msg_created` event with a direction?

ericnewcomer:
Hmm... `msg_created` isn't terrible. Definitely closer to what the engine has actually done. It definitely never sent a message (no matter how hard you perturb your perspective), which is I think what feels the dirtiest to me. But ya, you're right this would imply direction to be embedded which actually feels closer to right to me anyways.

rowanseymour:
One reason to keep `msg_received` and `msg_sent` distinct is that `msg_received` is the expected event in response to a `msg_wait`... there's a nice simplicity to events being satisfied by an event of particular type. Also currently `msg_sent` unlike `msg_received` doesn't map to what we would call a message in rapidpro - it's potentially many messages to different groups, contacts and URNs - but that's maybe a different issue?

Am still really not getting the objection to `contact_changed`. All the events ending in `*_changed` record that something has changed, and then their payload is the thing that changed. The events that are actually are diffs are the ones that end with `*_added` and `*_removed`.

nicpottier:
Maybe should be `broadcast_created` for outgoing then? It would be really nice to move away from `msg_sent` because ya, that's just totally inaccurate. `_received` is still maybe ok in that case? I find those being different events weird though, and wouldn't have any issue with having `msg_wait` be waiting for a `msg_created` event with a direction of `I`. But the argument that these have different attributes is a strong one, so maybe `msg_received` and `broadcast_created`? I think that's yet another reason against `msg_sent` in that it isn't mapping to a single message.

Rowan and I are also now observing our inconsistencies around calling out `contact` in `contactfield_changed` but not in `urn_added`. Seems we should either always prepend contact (which although longer seems more correct to me because it makes the engine more generic) or never prepend contact.

rowanseymour:
I'm down with `broadcast_created` and `msg_created` (or `reply_created`). Useful to distinguish those two, especially if `msg_created` is gonna do the work of resolving channels and urns - which obviously `broadcast_created` can't do.

Concur we should include `contact` in the event names in the name of keeping the engine generic. It was intentional to give all the event names have clear left (the thing) and right (the verb) sides so (I don't feel that strongly) but I do prefer `contactfield_changed` over `contact_field_changed`. If others don't like that, can we introduce some more punctuation here? Like `contact_field.changed` or `contact_field:changed`?

ericnewcomer:
I'm cool with `contactfield_changed` and `contact_field_changed`. Really don't like `contact_field:changed` or `contact_field.changed`. Not sure what that buys you other than a bit of confusion over which punctuation to use.

rowanseymour:
Ok this the current state of events. I added the extra underscores but I honestly don't yet get which `*_changed` events should become something else.

In theory these events can go either way:

|Name                                    | Engine/caller should |
|---------------------------|----------------------------|
|contact_changed                | Replace the contact             |
|contact_channel_changed | Replace the contact channel value  |
|contact_field_changed       | Replace the contact field value       |
|contact_groups_added      | Apply the contact groups diff |
|contact_groups_removed  | Apply the contact groups diff |
|contact_property_changed | Replace the contact property value |
|contact_urn_added | Add the URN to the contact |
|error                         | Nothing, just records an error |
|environment_changed | Replace the session environment |
|flow_triggered  | Nothing, just records that a subflow was started |
|input_labels_added | Add the labels to the input |
|run_result_changed | Replace the run result value |
|webhook_called | Update the webhook on the run |

There are some events that only the caller can receive:

|Name                                    | Caller should |
|---------------------------|----------------------------|
|broadcast_created              | Send the broadcast              |
|email_created               | Send the email... might be replaced with email_sent in future |
|msg_wait | Resume with a msg_received or run_expired |
|nothing_wait | Resume with nothing... |
|session_triggered | Start new engine sessions for the given contacts |

There are two events that only the engine can receive:

|Name                                    | Engine should |
|---------------------------|----------------------------|
|msg_received | Replace the run input |
|run_expired | Resume parent run |

There are two events we might add in future:

|Name                                    | Engine/caller should |
|---------------------------|----------------------------|
|email_sent               | Nothing, just records that email was sent |
|msg_created              | Send the message to the run contact on the specific URN  |

Clearly we also need to think about how/where these are versioned. 

@nicpottier @ericnewcomer

nicpottier:
Weren't we talking about having a `msg_created` (and use that for the `reply` action which we were talking about changing to `msg`?)

PR #154, opened 2018-02-28
Event refactor part 2

rowanseymour:
* Allows environment and contact to be set on the session trigger alongside flow etc
* Replaces `set_extra` event with `params` on the trigger. This is then exposed in expressions as `@trigger.params`
* Legacy references to `@extra` are mapped to an expression which checks both `@trigger.params` and `@run.webhook.json`


cmd/flowrunner/runner_test.go
```diff
@@ -21,23 +20,22 @@ import (
 )
 
 var flowTests = []struct {
-	assets  string
-	contact string
-	output  string
+	assets string
+	output string
```
rowanseymour: 'output' contains a trigger and the events, so since contact is now on trigger, these tests become a bit more self-contained. Previously contact and environment were being set programatically so not appearing anywhere in the output

excellent/legacy.go
```diff
@@ -107,6 +121,39 @@ func (v *varMapper) String() string {
...
+	return m
+}
+
+func (m *extraMapper) String() string {
+	switch m.extraAs {
+	case ExtraAsWebhookJSON:
+		return fmt.Sprintf("run.webhook.json.%s", m.path)
+	case ExtraAsTriggerParams:
+		return fmt.Sprintf("trigger.params.%s", m.path)
+	case ExtraAsFunction:
```
rowanseymour: I added three different ways to map `@extra` because although the `@(if(...))` solution should work in all cases, it seemed nice to map to a simple expression if the flow didn't have any webhook actions.

But making that work with legacy flow parsing is gonna require a significant work of legacy flow parsing (e.g. UnmarshalJSON methods with no access to state about the wider flow), so I wanna check how people feel about this first.

excellent/legacy.go
```diff
@@ -155,8 +202,7 @@ var functionTemplates = map[string]functionTemplate{
 	"time": {name: "time", params: "(%s %s %s)"},
 }
 
-func newRootVarMapper() *varMapper {
-
+func newMigrationBaseVars() map[string]interface{} {
```
rowanseymour: all so that the root var mapper can be parameterized (i.e. the different ways to map `@extra`), without having to re-initialize this giant map everytime

PR #163, opened 2018-03-02
better toml reading, better env variables

nicpottier:


flowserver.toml
```diff
@@ -10,12 +10,12 @@
 #-------------------------------------------------------------------------------------
 
 # Which port to listen on
-Port = 8080
+port = 8800
```
rowanseymour: will update rapidpro

Issue #143, opened 2018-01-25
Add validation for HTTP verbs


ycleptkellan:
Seems like an MVP implementation of the webhook router should include this verb as updating an existing resource appears to be a fairly common use case for the 'Call Webhook' step.

nicpottier:
Current implementation just uses the method passed in:
  https://github.com/nyaruka/goflow/blob/master/flows/actions/call_webhook.go

So this will actually work as is. May make sense to add some validation for real HTTP verbs into this action though.

ycleptkellan:
Ah OK thanks. Makes sense to me, updated issue

rowanseymour:
Added in #164

PR #164, opened 2018-03-03
Validate HTTP methods on webhook actions 

rowanseymour:
Addresses https://github.com/nyaruka/goflow/issues/143

### DIFF
--- a/cmd/flowrunner/main.go
+++ b/cmd/flowrunner/main.go
@@ -21,6 +21,10 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
+const (
+	outputIndent string = "    "
+)
+
 type Output struct {
 	Session json.RawMessage   `json:"session"`
 	Log     []json.RawMessage `json:"log"`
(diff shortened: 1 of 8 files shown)


cmd/flowrunner/main.go
```diff
@@ -21,6 +21,10 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
+const (
+	outputIndent string = "    "
```
rowanseymour: Unrelated but I formatted all the JSON files in a prev PR to use 4 space indentation, so need to update this to not rewrite them with 2 spaces

utils/validator.go
```diff
@@ -86,6 +86,8 @@ func validate(obj interface{}, objName string) error {
 			problem = fmt.Sprintf("must have a maximum of %s items", fieldErr.Param())
 		case "mutually_exclusive":
 			problem = fmt.Sprintf("is mutually exclusive with '%s'", fieldErr.Param())
+		case "http_method":
+			problem = "is not a valid HTTP method"
```
rowanseymour: one advantage of a custom tag is that it's easier to customize the error message

cmd/flowrunner/testdata/flows/all_actions.json
```diff
@@ -163,7 +163,7 @@
                         {
                             "uuid": "06153fbd-3e2c-413a-b0df-ed15d631835a",
                             "type": "call_webhook",
-                            "method": "get",
+                            "method": "GET",
```
rowanseymour: editor saves as uppercase so I think we can be strict here and require correct case

PR #167, opened 2018-03-05
Test improvements

rowanseymour:
* Don't clear UUIDs in tests but instead use a fixed sequence of UUIDs for testing
 * Add more validation now that we don't need to allow empty UUIDs all over the place


cmd/flowrunner/runner_test.go
```diff
@@ -168,27 +170,53 @@ func newTestHTTPServer() *httptest.Server {
...
 	// save away our server URL so we can rewrite our URLs
 	serverURL = server.URL
 
 	for _, test := range flowTests {
 		testJSON, err := readFile("flows/", test.output)
-		if err != nil {
-			t.Errorf("Error reading output file for flow '%s' and output '%s': %s", test.assets, test.output, err)
-			continue
-		}
+		require.NoError(t, err, "Error reading output file for flow '%s' and output '%s': %s", test.assets, test.output, err)
```
rowanseymour: `require.NoError` is a clean way to fail fast on things that aren't really test failures so much as fundamental problems

cmd/flowrunner/runner_test.go
```diff
@@ -168,27 +170,53 @@ func newTestHTTPServer() *httptest.Server {
 	return server
 }
 
+type fixedUUIDGenerator struct {
+	uuids []string
+	index int
+}
+
+func fixedUUIDGeneratorFromFile(path string) (*fixedUUIDGenerator, error) {
```
nicpottier: Does this actually need to be from a file? Can't we seed the random number generator and get deterministic UUIDs?

rowanseymour:
@nicpottier the satori library is using `crypto/rand` and best I can tell only `math/rand` allows seeding, but I guess it we can just write one based on `math/rand`

PR #168, opened 2018-03-06
msg_created event

rowanseymour:
* New `msg_created` event which takes a `MsgOut` as its payload
* New `msg_received` updated to take a `MsgIn` as its payload instead of a `MsgInput` - we generate the input from the message when the event is applied
* `Contact.Channel` replaced with URN specific preferred channels
* `set_preferred_channel` action implemented as it is RP
* `reply` action updated to pick a channel, and not generate `msg_created` events for any URNs it can't find a channel for.


cmd/flowrunner/testdata/default.json
```diff
@@ -53,13 +53,31 @@
...
     {
         "type": "channel",
-        "url": "http://testserver/assets/channel/57f1078f-88aa-46f4-a59a-948a5739c03d",
-        "content": {
-            "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d",
-            "name": "Twilio Channel",
-            "address": "+12345671111",
-            "type": "T"
-        }
+        "url": "http://testserver/assets/channel",
```
rowanseymour: need some extra channels here because we only create messages when have a channel to send it

cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -8,10 +8,9 @@
                         "name": "Nexmo",
                         "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d"
                     },
-                    "created_on": "2000-01-01T00:00:00.000000000-00:00",
                     "text": "Ryan Lewis",
                     "urn": "tel:+12065551212",
-                    "uuid": "84f8a3cf-0f2c-4881-9502-2d7b114bf01f"
```
rowanseymour: if you change the number of calls to `utils.NewUUID`... all the UUIDs are going to change in the test data

cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -204,59 +203,99 @@
...
                 {
                     "action_uuid": "d98c1e02-69df-4f95-8b89-8587a57ae0c3",
                     "event": {
                         "created_on": "2000-01-01T00:00:00.000000000-00:00",
-                        "text": "This is a message to each of Ben Haggerty's urns.",
-                        "type": "broadcast_created",
-                        "urns": [
-                            "tel:+12065551212",
-                            "facebook:1122334455667788",
-                            "mailto:ben@macklemore",
```
rowanseymour: note that there is no msg_created event for this URN because we can't send to it

cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -24,25 +23,27 @@
...
+                    "step_uuid": "8707af30-d50f-440b-9803-f4a851d20f2b"
                 },
                 {
                     "event": {
                         "created_on": "2000-01-01T00:00:00.000000000-00:00",
                         "timeout_on": "2000-01-01T00:00:00.000000000-00:00",
                         "type": "msg_wait"
                     },
-                    "step_uuid": "6bce5330-ed5e-418c-a13f-9238de2a0b76"
+                    "step_uuid": "5fa51f39-76ea-421c-a71b-fe4af29b871a"
```
nicpottier: Would our thrash of UUIDs be diminished on changes if we reset the seed on every test run?

flows/actions/reply.go
```diff
@@ -43,12 +49,55 @@ func (a *ReplyAction) Validate(assets flows.SessionAssets) error {
...
 	} else {
-		log.Add(events.NewMsgCreatedEvent(evaluatedText, evaluatedAttachments, evaluatedQuickReplies, run.Contact().Reference()))
+		// send to first URN which has a corresponding channel (i.e. is sendable)
+		for _, u := range run.Contact().URNs() {
+			channel := getChannelForURN(channels, u)
+			if channel != nil {
+				destinations = append(destinations, msgDestination{urn: u.URN, channel: channel})
+				break
+			}
+		}
```
nicpottier: These loops look the same except you could do the break conditionally based on `allURNs`

flows/actions/reply.go
```diff
@@ -43,12 +49,55 @@ func (a *ReplyAction) Validate(assets flows.SessionAssets) error {
...
+	for _, dest := range destinations {
+		msg := flows.NewMsgOut(dest.urn, dest.channel, evaluatedText, evaluatedAttachments, evaluatedQuickReplies)
+		log.Add(events.NewMsgCreatedEvent(msg))
+	}
+
+	return nil
+}
+
+// returns the best channel for the given URN
+func getChannelForURN(channels []flows.Channel, urn flows.ContactURN) flows.Channel {
```
nicpottier: This feels like maybe this should be on ChannelSet instead of in Reply Action

flows/channel.go
```diff
@@ -54,66 +97,75 @@ func (c *channel) String() string {
...
 
-// ReadChannels decodes channels from the passed in JSON
-func ReadChannels(data []json.RawMessage) ([]Channel, error) {
-	channels := make([]Channel, len(data))
-	var err error
-	for c := range data {
-		channels[c], err = ReadChannel(data[c])
-		if err != nil {
-			return nil, err
+func (s *ChannelSet) WithRole(role ChannelRole) []Channel {
```
nicpottier: to me `with_` methods generally should return the same type as the base object, so this is a bit weird to me. Might be able to just simplify all this by sticking the above method used for finding the right sending channel on ChannelSet entirely.

flows/events/contact_channel_changed.go
```diff
@@ -39,6 +42,40 @@ func (e *ContactChannelChangedEvent) Apply(run flows.FlowRun) error {
...
+		if newPreferredURN != nil {
+			newURNs := make(flows.URNList, 1)
+			newURNs[0] = *newPreferredURN
+
+			for _, urn := range contactURNs {
+				if urn.URN != newPreferredURN.URN {
+					newURNs = append(newURNs, urn)
+				}
+			}
+		}
```
nicpottier: Is the above working? How are `contactURNs` being updated? Seems you are creating a new list and reordering it but how does that end up updating the contact urns?

Also, can't this just be a two element swap instead of creation of a whole new list?

cmd/flowrunner/main.go
```diff
@@ -232,7 +231,7 @@ func main() {
 		scanner.Scan()
 
 		// create our event to resume with
-		msg := inputs.NewMsgInput(flows.InputUUID(utils.NewUUID()), nil, time.Now(), contact.URNs()[0].URN, scanner.Text(), []flows.Attachment{})
+		msg := flows.NewMsgIn(flows.MsgUUID(utils.NewUUID()), contact.URNs()[0].URN, nil, scanner.Text(), []flows.Attachment{})
```
nicpottier: Seems nicer to just add a `NewMsgUUID` function here..

rowanseymour: Agree and I also think it would be nice to move those UUID subtype definitions closer to the structs that use them rather than having a list in interfaces.go But we do this everywhere in goflow now so maybe can change in different PR.

flows/channel.go
rowanseymour: Moved to `ChannelSet.GetForURN`

PR #169, opened 2018-03-06
Contact groups and channel should be references instead of single UUIDs

rowanseymour:
Like how they are elsewhere

### DIFF
--- a/flows/contact.go
+++ b/flows/contact.go
@@ -196,14 +196,14 @@ type fieldValueEnvelope struct {
 }
 
 type contactEnvelope struct {
-	UUID        ContactUUID                     `json:"uuid" validate:"required,uuid4"`
-	Name        string                          `json:"name"`
-	Language    utils.Language                  `json:"language"`
-	Timezone    string                          `json:"timezone"`
-	URNs        URNList                         `json:"urns"`
-	GroupUUIDs  []GroupUUID                     `json:"group_uuids,omitempty" validate:"dive,uuid4"`
-	Fields      map[FieldKey]fieldValueEnvelope `json:"fields,omitempty"`
-	ChannelUUID ChannelUUID                     `json:"channel_uuid,omitempty" validate:"omitempty,uuid4"`
+	UUID     ContactUUID                     `json:"uuid" validate:"required,uuid4"`
+	Name     string                          `json:"name"`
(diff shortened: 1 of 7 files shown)


flows/contact.go
```diff
@@ -196,14 +196,14 @@ type fieldValueEnvelope struct {
...
-	Fields      map[FieldKey]fieldValueEnvelope `json:"fields,omitempty"`
-	ChannelUUID ChannelUUID                     `json:"channel_uuid,omitempty" validate:"omitempty,uuid4"`
+	UUID     ContactUUID                     `json:"uuid" validate:"required,uuid4"`
+	Name     string                          `json:"name"`
+	Language utils.Language                  `json:"language"`
+	Timezone string                          `json:"timezone"`
+	URNs     URNList                         `json:"urns"`
+	Groups   []*GroupReference               `json:"groups,omitempty" validate:"dive"`
+	Fields   map[FieldKey]fieldValueEnvelope `json:"fields,omitempty"`
+	Channel  *ChannelReference               `json:"channel,omitempty" validate:"omitempty,dive"`
```
rowanseymour: possibly these shouldn't be pointers, but if not that's for another day as they are pointers everywhere else

nicpottier:
What's the argument again for having references over UUIDs?

rowanseymour:
@nicpottier we use references elsewhere like on input/events/actions. Events and actions did use to be different in that regard but you pushed for making them consistent.

nicpottier:
But what was our reasoning again? Why do we want refs as opposed to UUIDs? Maybe we should document these decisions and the thought behind them.

rowanseymour:
@nicpottier consistency and readability was what I recall

rowanseymour:
@nicpottier if we're formulating a rule - I don't think every like between two things in the session session needs to be a reference object, e.g. `run.parent_uuid`. But it is kinda nice for references to assets which aren't defined in the session JSON.

nicpottier:
So what's the hard rule as to when a naked UUID is ok vs not ok?

rowanseymour:
@nicpottier use a reference object when referencing an asset (e.g. group, label). Use a naked UUID when referencing something defined in the session itself (e.g. run, step)

nicpottier:
k, that works.

PR #170, opened 2018-03-08
Add support for URNs with channels as query params

rowanseymour:
Introduces a `ContactURN` as a wrapper for URNs on a contact so we can extract their channel like https://github.com/nyaruka/goflow/compare/new_urns?expand=1#diff-488bb6c851b01295dc4960583fb229d8R38

### DIFF
--- a/cmd/flowserver/static/start.json
+++ b/cmd/flowserver/static/start.json
@@ -37,7 +37,7 @@
             "timezone": "America/Guayaquil",
             "urns": [
                 "tel:+12065551212",
-                "facebook:macklemore"
+                "facebook:5637322626267272"
             ]
         },
         "flow": {
--- a/flows/contact.go
+++ b/flows/contact.go
@@ -200,7 +200,7 @@ type contactEnvelope struct {
 	Name     string                          `json:"name"`
 	Language utils.Language                  `json:"language"`
 	Timezone string                          `json:"timezone"`
-	URNs     []urns.URN                      `json:"urns"`
+	URNs     []urns.URN                      `json:"urns" validate:"dive,urn"`
 	Groups   []*GroupReference               `json:"groups,omitempty" validate:"dive"`
 	Fields   map[FieldKey]fieldValueEnvelope `json:"fields,omitempty"`
 	Channel  *ChannelReference               `json:"channel,omitempty" validate:"omitempty,dive"`
--- a/flows/contact.go
+++ b/flows/contact.go
@@ -280,7 +280,7 @@ func (c *Contact) MarshalJSON() ([]byte, error) {
 	ce.Name = c.name
 	ce.UUID = c.uuid
 	ce.Language = c.language
-	ce.URNs = c.urns.RawURNs()
+	ce.URNs = c.urns.RawURNs(true)
 	if c.timezone != nil {
 		ce.Timezone = c.timezone.String()
 	}
--- a/flows/contact.go
+++ b/flows/contact.go
@@ -77,7 +77,7 @@ func (c *Contact) URNs() URNList { return c.urns }
 func (c *Contact) AddURN(urn urns.URN) {
 	// TODO normalize and check we're not adding duplicates
 
-	c.urns = append(c.urns, urn)
+	c.urns = append(c.urns, &ContactURN{URN: urn})
 }
 
 // Groups returns the groups that this contact belongs to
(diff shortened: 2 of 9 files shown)


flows/urns.go
```diff
@@ -27,8 +28,78 @@ func ValidateURNScheme(fl validator.FieldLevel) bool {
...
+type URNList []*ContactURN
+
+func ReadURNList(session Session, rawURNs []urns.URN) (URNList, error) {
+	l := make(URNList, len(rawURNs))
+
+	for u := range rawURNs {
+		scheme, path, query, display := rawURNs[u].ToParts()
+
+		// re-create the URN without the query component
+		queryLess, err := urns.NewURNFromParts(scheme, path, "", display)
```
rowanseymour: might be worth having an exported way to rebuild a URN without doing validation again

cmd/flowserver/static/start.json
```diff
@@ -37,7 +37,7 @@
             "timezone": "America/Guayaquil",
             "urns": [
                 "tel:+12065551212",
-                "facebook:macklemore"
```
rowanseymour: nothing against macklemore - this just isn't a valid facebook URN

nicpottier: :(

flows/contact.go
```diff
@@ -169,7 +169,7 @@ func (c *Contact) ResolveQueryKey(key string) []interface{} {
 		urnsWithScheme := c.urns.WithScheme(key)
 		vals := make([]interface{}, len(urnsWithScheme))
 		for u := range urnsWithScheme {
-			vals[u] = string(urnsWithScheme[u])
+			vals[u] = string(urnsWithScheme[u].URN)
```
nicpottier: Shouldn't this be returning a list of `ContactURN` objects for the scheme? IE, shouldn't this kind of scheme filtering be returning a filtered list of the same type as the main URN list? Or am I understanding this wrong?

flows/urns.go
```diff
@@ -27,8 +28,78 @@ func ValidateURNScheme(fl validator.FieldLevel) bool {
...
+	return fmt.Errorf("no field '%s' on URN", key)
+}
+
+func (u *ContactURN) Default() interface{} { return u }
+
+func (u *ContactURN) String() string { return string(u.URN) }
+
 // URNList is the list of a contact's URNs
-type URNList []urns.URN
+type URNList []*ContactURN
```
nicpottier: Why pointer?

flows/urns.go
```diff
@@ -27,8 +28,78 @@ func ValidateURNScheme(fl validator.FieldLevel) bool {
...
+	for u := range rawURNs {
+		scheme, path, query, display := rawURNs[u].ToParts()
+
+		// re-create the URN without the query component
+		queryLess, err := urns.NewURNFromParts(scheme, path, "", display)
+		if err != nil {
+			return nil, err
+		}
+
+		parsedQuery, err := url.ParseQuery(query)
```
nicpottier: Occurs to me we may want to think about what the URN value that is exposed is.. gets kinda tricky, but ya, a URN like `facebook:12345#rowanseymour` is treated the same in our system as `facebook:12345#rowan`, the display portion is not part of the "identity" of the URN.

I'm not worried about us not doing the right thing here, but wondering about users using these URNs in outside systems and the like, do they have an easy way to get a URN `identity` as we would call it? (stripped of channel AND display)

flows/urns.go
rowanseymour: Because I have an irrational fear of passing around structs that I need to get over?

flows/contact.go
rowanseymour: So the practical reason is - we have three places URNs occur in goflow. 1) on contacts 2) events 3) actions. Contacts are unmarshaled in the context of a session. They have a separate envelope struct and real struct and they can resolve things like channels as they are unmarshalled. Actions and events are somewhat simpler beasts - they are unmarshaled straight into their struct without access to the session. 

Viewed one way that's a bit inconsistent. Viewed another way that's a nice optimization. A session might have a lot of old events and it's nice to keep their unmarshaling relatively lightweight. It also means we can parse a session with events that reference assets which are no longer available. For example an old event might reference a channel which has been since deleted - we don't want to blow up in that case because we can't fetch the channel.

PR #174, opened 2018-03-10
Event validation

rowanseymour:
* All caller events are validated before starting or resuming a session
 * Events declare their allowed origins (engine vs caller vs either)


flows/events/contact_property_changed.go
```diff
@@ -38,6 +38,16 @@ func NewContactPropertyChangedEvent(property string, value string) *ContactPrope
 // Type returns the type of this event
 func (e *ContactPropertyChangedEvent) Type() string { return TypeContactPropertyChanged }
 
+// AllowedOrigin determines where this event type can originate
```
nicpottier: Instead of repeating these can't we embed proper base types that implement these?

rowanseymour: updated

Issue #152, opened 2018-02-27
Panic when executing reply action without contact


ericnewcomer:
Should instead return a friendly error.
```
runtime/debug.Stack(0xc420500e68, 0x1720760, 0x217f370)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x79
github.com/nyaruka/goflow/vendor/github.com/pressly/lg.RequestLogger.func1.1.1(0xc42045ca60, 0x21434c0, 0xc420142f80, 0xed2279d75, 0x22713320, 0x218f1e0)
	/Users/eric/code/go/src/github.com/nyaruka/goflow/vendor/github.com/pressly/lg/middleware.go:30 +0x14e
panic(0x1720760, 0x217f370)
	/usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/nyaruka/goflow/flows/actions.(*ReplyAction).Execute(0xc4204e4300, 0x214dda0, 0xc4205fc000, 0x2143fa0, 0xc420112850, 0x213bde0, 0xc420353440, 0x0, 0x0)
	/Users/eric/code/go/src/github.com/nyaruka/goflow/flows/actions/reply.go:46 +0x176
github.com/nyaruka/goflow/flows/engine.(*session).visitNode(0xc42008ca00, 0x214dda0, 0xc4205fc000, 0x21404c0, 0xc4204e42a0, 0xc420379640, 0x1, 0x1, 0xc4205fc000, 0xc4205012a8, ...)
```

rowanseymour:
We've got a few actions which don't make so much sense without a contact - and currently some of them noop. But error does seem more appropriate.

ericnewcomer:
For sure, just making sure we log cases where the caller can blow things up with unexpected input so we can return error messages instead.

rowanseymour:
Fixed in #175

PR #175, opened 2018-03-10
Don't blow up in sessions without contacts

rowanseymour:
* An action that requires a contact should generate an error event if the session doesn't have a contact
 * An event that requires a contact should return an err if the session doesn't have a contact

Addresses https://github.com/nyaruka/goflow/issues/152


README.md
```diff
@@ -28,7 +28,7 @@ This utility provides a quick way to test evaluation of expressions which can be
 % $GOPATH/bin/exptester '@(TITLE("foo"))'
 ```
 
-## Running Tests
+## Development
 
 You can run all the tests (excluding tests in vendor packages) with:
```
nicpottier: While you are at it you can just change the below to `go test github.com/nyaruka/goflow/...`

PR #180, opened 2018-03-13
Broadcast languages

rowanseymour:
Should use the flow translation languages rather than the env languages - which doesn't actually mean the same thing as org.languages in RapidPro - it's the list of backdowns instead.

### DIFF
--- a/flows/definition/localization.go
+++ b/flows/definition/localization.go
@@ -25,8 +25,16 @@ func (t *languageTranslations) GetTextArray(uuid utils.UUID, key string) []strin
 // flowTranslations are our top level container for all the translations for a language
 type flowTranslations map[utils.Language]*languageTranslations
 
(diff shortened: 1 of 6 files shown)


flows/definition/localization.go
```diff
@@ -25,8 +25,16 @@ func (t *languageTranslations) GetTextArray(uuid utils.UUID, key string) []strin
 // flowTranslations are our top level container for all the translations for a language
 type flowTranslations map[utils.Language]*languageTranslations
 
-func (t *flowTranslations) GetLanguageTranslations(lang utils.Language) flows.Translations {
-	translations, found := (*t)[lang]
+func (t flowTranslations) Languages() utils.LanguageList {
```
rowanseymour: now with less pointers

nicpottier: fewer :)

rowanseymour: ARRRGHHHHHHHHHHH

PR #181, opened 2018-03-13
Don't add duplicate URNs to contacts

rowanseymour:


flows/contact.go
```diff
@@ -76,10 +76,24 @@ func (c *Contact) Name() string { return c.name }
 func (c *Contact) URNs() URNList { return c.urns }
 
 // AddURN adds a new URN to this contact
-func (c *Contact) AddURN(urn urns.URN) {
-	// TODO normalize and check we're not adding duplicates
+func (c *Contact) AddURN(urn urns.URN) bool {
+	if c.HasURN(urn) {
+		return false
```
rowanseymour: originally had this as error case but then noticed we don't blow up if contact_group_added event tries to add a duplicate group

flows/contact.go
```diff
@@ -76,10 +76,24 @@ func (c *Contact) Name() string { return c.name }
...
+	c.urns = append(c.urns, &ContactURN{URN: urn.Normalize("")})
+	return true
+}
+
+// HasURN checks whether the contact has the given URN
+func (c *Contact) HasURN(urn urns.URN) bool {
+	urn = urn.Normalize("")
 
-	c.urns = append(c.urns, &ContactURN{URN: urn})
+	for u := range c.urns {
```
nicpottier: `for _, u := range c.urns {`?

Issue #155, opened 2018-03-01
Move legacy migration code into new top-level legacy package


rowanseymour:
excellent/legacy.go and flow/definition/legacy.go should move elsewhere so we're not filling those namespaces with legacy specific stuff

rowanseymour:
Addressed in #187

PR #187, opened 2018-03-15
Move legacy stuff into its own package part 2/2

rowanseymour:
I already moved all the excellent v1 stuff to legacy/expressions.go - this PR now moves all the legacy flow stuff to legacy/definition.go. Nice to have all the guff in one place and untangle some dependencies.

Addresses https://github.com/nyaruka/goflow/issues/155


cmd/flowserver/server.go
```diff
@@ -269,11 +269,19 @@ func (s *FlowServer) handleMigrate(w http.ResponseWriter, r *http.Request) (inte
 		return nil, fmt.Errorf("missing flows element")
 	}
 
-	flows, err := definition.ReadLegacyFlows(migrate.Flows)
+	legacyFlows, err := legacy.ReadLegacyFlows(migrate.Flows)
```
rowanseymour: previously flow migration only fully happened when re-marshalling back to JSON. Now you get a `legacy.Flow` with a `Migrate` method. I think this is simpler and makes it easier to apply the same validation rules to migrated flows.

flows/actions/send_broadcast.go
```diff
@@ -53,7 +53,7 @@ func (a *SendBroadcastAction) Execute(run flows.FlowRun, step flows.Step, log fl
 	}
 
 	translations := make(map[utils.Language]*events.BroadcastTranslation)
-	languages := append(utils.LanguageList{run.Flow().Language()}, run.Flow().Translations().Languages()...)
+	languages := append(utils.LanguageList{run.Flow().Language()}, run.Flow().Localization().Languages()...)
```
rowanseymour: in the flow spec we call the top level thing `localization` so I've renamed it in the code and that helps distinguish it from languageTranslations and itemTranslations which it's composed of.

legacy/definition.go
```diff
@@ -41,68 +41,64 @@ var legacyWebhookBody = `{
 	"channel": {}
 }`
 
-// LegacyFlow imports an old-world flow so it can be exported anew
-type LegacyFlow struct {
-	flow
-	envelope legacyFlowEnvelope
+// Flow is a flow in the legacy format
+type Flow struct {
```
rowanseymour: golint doesn't like names like `foo.FooBar` and complains about stuttering

legacy/testdata/migrations/rulesets.json
```diff
@@ -218,7 +218,7 @@
                 {
                     "category": {"eng": "< 10"}, 
                     "uuid": "1c75fd71-027b-40e8-a819-151a0f8140e6", 
-                    "destination": "7d40faea-723b-473d-8999-59fb7d3c3ca2", 
+                    "destination": "5b977652-91e3-48be-8e86-7c8094b4aa8f",
```
rowanseymour: now that loading legacy flows uses same validation rules - these actually have to be existing nodes

PR #197, opened 2018-03-20
Restructure event output so that events aren't wrapped in log entries

rowanseymour:


cmd/flowserver/config.go
```diff
@@ -17,7 +17,7 @@ type Config struct {
 func NewDefaultConfig() *Config {
 	return &Config{
 		Port:             8800,
-		LogLevel:         "error",
+		LogLevel:         "info",
```
rowanseymour: making 'info' the default but can update rapidpro to override to 'warn'

PR #203, opened 2018-03-28
remove int in expressions, map to round_down for legacy

### LINKED ISSUE #194: Remove INT() function
We have `round_down`, no need for duplicates.

nicpottier:
fixes #194


nicpottier:
Sure seems that way to me no? I mean to a simpleton `round_up` is pretty obvious while `ceil` is super weird no?

PR #206, opened 2018-03-29
Resolving refactor part 1

rowanseymour:
Playing around with variable resolving. The idea would be:

 * All variables in excellent should be one of:
    - A primitive of an `XAtom` type
    - A complex object that implements `VariableAtomizer`, i.e. can reduce itself to an `XAtom`
    - Lists of either of the above (these would cease to be `XAtom` types themselves)
* Make sure the excellent visitor can only produce such types, and we can greatly simply our conversion functions to not worry about other golang types.
* Replace `Default()` and `String()` with `Atomize()` which does the reduction to a primitive (usually a `string`). For example datetime field value returns itself as a `time.Time` - it doesn't need to know how to stringify itself - the conversion routines in utils have that logic in one place.

In this PR I've left `String()` around on most objects and it takes priority over `Atomize()` if it is defined.

### DIFF
--- a/utils/resolver.go
+++ b/utils/resolver.go
@@ -19,7 +19,6 @@ func Snakify(text string) string {
 // VariableResolver defines the interface used by Excellent objects that can be indexed into
 type VariableResolver interface {
 	Resolve(key string) interface{}
-	Default() interface{}
 	String() string
 }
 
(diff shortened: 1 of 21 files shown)


utils/resolver.go
```diff
@@ -19,8 +19,11 @@ func Snakify(text string) string {
 // VariableResolver defines the interface used by Excellent objects that can be indexed into
 type VariableResolver interface {
 	Resolve(key string) interface{}
-	Default() interface{}
-	String() string
+}
+
+// VariableAtomizer defines the interface used by Excellent objects that can be indexed into
```
nicpottier: comment

PR #211, opened 2018-03-30
Variables refactor part 2

rowanseymour:
* Introduces `utils.Indexable` interface for context items which can be indexed into and have a length
 * Adds `Array` type an indexable holder of other context items
 * Simplify conversions by removing golang slice types - these types should never appear in the context


flows/engine/session_test.go
```diff
@@ -28,9 +28,10 @@ func TestEvaluateTemplateAsString(t *testing.T) {
 		{"@contact.first_name", "Ben", false},
 		{"@contact.language", "eng", false},
 		{"@contact.timezone", "America/Guayaquil", false},
-		{"@contact.urns", "tel:+12065551212", false}, // TODO should be list
+		{"@contact.urns", "tel:+12065551212, facebook:1122334455667788, mailto:ben@macklemore", false},
```
rowanseymour: this is an `Array` which `ToString` converts to CSV by default

excellent/evaluator.go
```diff
@@ -238,7 +238,7 @@ func (s *xscanner) Scan() (xToken, string) {
 }
 
 // EvaluateExpression evalutes the passed in template, returning the raw value it evaluates to
-func EvaluateExpression(env utils.Environment, resolver utils.VariableResolver, template string) (interface{}, error) {
+func EvaluateExpression(env utils.Environment, resolver utils.Resolvable, template string) (interface{}, error) {
```
rowanseymour: Non-atom items in the context can now be `Resolvable` (i.e. can be keyed into), `Indexable` (i.e. can be indexed into) and/or `Atomizable` (i.e. can reduce themselves to an atom type)

utils/json.go
```diff
@@ -39,45 +40,81 @@ type JSONFragment []byte
...
+	length := 0
+	jsonparser.ArrayEach(j, func(value []byte, dataType jsonparser.ValueType, offset int, err error) {
+		length++
+	})
+	return length
+}
+
+var _ Indexable = JSONArray{}
+
+func jsonTypeToXAtom(data []byte, valType jsonparser.ValueType) interface{} {
```
rowanseymour: @nicpottier now any time we resolve something out of a `JSONFragment`, we see if we can convert it to one of our atomic types. `JSONArray` is a special case of a fragment.. which is an array and thus can be indexed into and has length.

PR #214, opened 2018-03-30
Remove no longer used types from conversions

rowanseymour:


excellent/functions.go
```diff
@@ -1638,7 +1638,7 @@ func ToEpoch(env utils.Environment, args ...interface{}) interface{} {
 		return err
 	}
 
-	return date.UnixNano()
+	return decimal.New(date.UnixNano(), 0)
```
rowanseymour: @nicpottier I'm leaning toward getting rid of ints altogether (like Javascript) and always using decimal - which anyway is a bigint plus an exponent which can be zero. It's one less thing that can exist in the context, one less thing to worry about converting correctly. Thoughts?

nicpottier: I don't think there's a huge plus to that honestly. Would require us to change all our functions to return those as well. I think keeping the normal golang primitives makes sense.

rowanseymour: It only requires changing (I think..) 3 functions which currently return an `int` to return a `Decimal` instead. Those are now the only ways ints can get into the context. Seems changing 3 return statements is worth it to not have to worry about another data type occuring in the context. `WordCount` already returns a decimal for some reason.

nicpottier: I disagree, I don't think having support for int is a big deal.

nicpottier: Let me try to explain the rational a bit better and what I think we should do instead.

Right now, everything is basically dynamic, we are returning `interface{}` all over the place in Excellent. We can make a convention of making that a certain set of types as you are, but five years from now some other person may add something to the context and only later find out that it isn't one of the supported types. Because we are being loosy-goody with types, I think it is worth having the slightest more centralized complexity (and I think it is vastly overstating it to even say "complexity" when we are talking about another case in a switch block) rather than pushing that complexity to everybody who generates or interacts with the context in some way.

Everything already HAS to do type-cracking and casting, so there's very little benefit in reducing our number of supported types to include the primitives, it makes the larger body of code (functions / tests / resolvers) use the most natural type possible, while hiding the support for that in a few extra switch statements. To me that is a fundamentally better design (slightly higher centralized complexity rather than slightly higher decentralized complexity) given how everything is dynamic.

That said, I think both options are wrong. I think we should instead be enforcing that functions/tests/resolvers return an interface, perhaps something called `XAtom`. That interface should satisfy the two thing we know every single thing in the context must do, stringify itself and jsonify itself. We create wrappers around bool, decimal, string, time which satisfy this as well as the existing resolvers and list interfaces you added. Tests and functions and resolvers must return something that satisfies that interface, they also take that interface as arguments.

We'll still have type casting to go from an `XAtom` to say an `XDecimal`, and that will still have a switch statement iterating across `XDecimal` and `XString`, but the attack surface of a future coder messing things up is a lot smaller, they would have to create a new XAtom implementor entirely to do so instead of using those provided and then not update the casting code. Even in that case casts wouldn't fail, since they always have a `String()` to fall back to and just say sorry, that's not a Decimal I can deal with. We no longer have the possibility of our switches getting to things they can't deal with.

To me that leverages the advantages we have with golang being statically typed to make sure we do the right thing. We know for sure there are currently lurking bugs in the context and return values which won't/don't properly JSONify or stringify, and though we will build tests to ferret those out and try to guarantee that isn't the case going forward, we know that accomplishing total coverage and maintaining it across that is much more difficult than having compile time checks. There's no reason we shouldn't be leaning on the compiler to make sure Excellent is proven to be correct and that all values in the Context play well with Excellent.

rowanseymour: So I think this is the crux of our disagreement:

> but five years from now some other person may add something to the context and 
> only later find out that it isn't one of the supported types

Frankly it sounds like you're saying we should build in more loosey-gooseyness for a hypothetical where someone adds some bad code - when from where I'm standing it seems trivial to add sanity checks to make sure things only ever evaluate to a set of known types. For example we can easily check at what `ResolveVariable` is returning. We could probably put something into the visitor to make sure every node in the parse tree evaluates to a known type if we really wanted to.

The direction I've been going over the past week is that we have a set of known Excellent types, and they each map to a Golang type, i.e.

Excellent Type | Golang 
------------ | -------------
XTypeNil | nil
XTypeError | error
XTypeString | string
XTypeDecimal | decimal.Decimal
XTypeTime | time.Time
XTypeBool | bool
XTypeArray | utils.Array

I think that's bought us good gains in simplicity - especially with the use of `Array` - now we don't have to know about slices of every golang type, and we can easily have complex objects (e.g. `@contact.groups`) that behave like any other array. And we're very close to being able to verify that everything that *can* be the context, is a valid type.

So I'm struggling to wrap my head around the idea that it's beneficial to allow a `float64` or a `[]bool` to suddenly popup during evaluation, and I really think you're overstating the effort required to ensure functions and context objects only ever return a valid type. As I've pointed out, Excellent's visitor can currently only generate one of those types with the exception of a handful of functions that return an `int` ✢, so that work is pretty much already done.

Another reason I don't like playing loose with types is that I think Excellent should be as well defined as any other part of the flow spec. It shouldn't behave differently depending on the particularities of certain golang types. Hypothetically it should be possible to write an Excellent evaluator in any language.

I do agree that the ideal solution would have these all wrapped in something that represents an Excellent variable which could be an `XAtom` primitive or a complex object that implements `XAtomizable`, but this feels like a first step to getting there, with almost all the benefits.

I would challenge you to come up with a concrete example of where stricter types would be a problem in the future. I think if there's something that comes up in the future and someone just absolutely has to return a `float32` instead of a `decimal`, we can cross that bridge when we get there. But right now I can't imagine what that would be.

✢ My specific objection to `int` being an atom type is that it's not functionally different to decimal in anyway. The `1` in `@(word(run.input, 1))` is already a decimal. There's nowhere in the excellent syntax that we treat `int` and `decimal` as different, so it seems better to not have two different ways of modeling the same thing.

nicpottier:
Not really going to get into this any more, I think the right solution is
to type these and just move on, we both agree that's better.

rowanseymour:
@nicpottier I'd prefer it if you addressed the points I've made like I tried to address your points, as I'm not sure what "move on" means.

nicpottier:
Move on means implement a statically typed solution we both agree is
better, that's all. We can disagree on which suboptimal solution is best. :)

rowanseymour:
Ok but you're going to have to clarify what "move on" means for this PR. 

If the end goal that we both agree on is a static set of types, then it seems to me like locking down which golang types are used to store each `XType` is a useful first step.

nicpottier:
I'm ok with any path that gets us there, if you want to merge this first go
for it, just continue to static types.

rowanseymour:
I'm gonna add a temporary sanity check in `ResolveVariable` to check the returned thing is a valid type and have created an issue for switching to static typing https://github.com/nyaruka/goflow/issues/215

PR #226, opened 2018-04-03
Reorganize part 2

rowanseymour:
* Router tests move from `excellent/tests.go` to `flows/routers/tests/*`
 * They register themselves dynamically with excellent
 * Fix `is_string_eq` not being recognized as a test 
 * Rework docgen to use a common session and handler setup for all events/actions/tests/functions
 * Move `runEnvironment` to flows/runs/environment.go and add interface in `flows/interfaces.go`
 * Move location searching functionality into run environment


excellent/visitor.go
```diff
@@ -71,15 +71,9 @@ func (v *Visitor) VisitFunctionCall(ctx *gen.FunctionCallContext) interface{} {
 	var function XFunction
 	var found bool
 
-	// this is a test, look it up from those
-	if strings.HasPrefix(functionName, "has_") {
```
rowanseymour: `is_string_eq` doh!

nicpottier: Are we doing `is` prefixes for tests now? If so maybe we should rename some of the other tests to be more natural? `has_string_eq`? Ya, thats weird.

rowanseymour: Here's the list...

```
has_error
has_value
has_group
has_wait_timed_out
is_string_eq
has_phrase
has_only_phrase
has_any_word
has_all_words
has_beginning
has_text
has_pattern
has_number
has_number_between
has_number_lt
has_number_lte
has_number_eq
has_number_gte
has_number_gt
has_date
has_date_lt
has_date_eq
has_date_gt
has_phone
has_email
has_state
has_district
has_ward
```

The only one that might be clearer with `is_` I think is `has_error` given that it checks whether the value IS an error. The others are "contains a" tests in some sense or another. `is_error`?

nicpottier: sure.

rowanseymour: I guess `has_value` could be `is_null`.. works either way

Issue #219, opened 2018-04-01
Move conversions/resolvers functionality into excellent


rowanseymour:
There's a growing amount of X functionality in `utils` that feels like it should be in `excellent` (or a sub-package within excellent). If something (e.g. `utils.RequestResponse`) depends on that stuff, it should move into `flows`.

PR #227, opened 2018-04-03
Move typey things from utils to excellent/types

rowanseymour:
Please review every line of this tiny PR in detail...

Addresses: #219


contactql/parser.go
```diff
@@ -78,23 +79,19 @@ func (c *Condition) evaluateValue(env utils.Environment, val interface{}) (bool,
 		return stringComparison(val.(string), c.comparator, c.value)
 
 	case decimal.Decimal:
-		asDecimal, err := utils.ToDecimal(env, c.value)
+		asDecimal, err := decimal.NewFromString(c.value)
```
rowanseymour: this is always a regular string-> decimal conversion so shouldn't go through excellent's `ToDecimal` which doesn't things like 0/O substitutions

contactql/parser.go
```diff
@@ -78,23 +79,19 @@ func (c *Condition) evaluateValue(env utils.Environment, val interface{}) (bool,
...
 
 	case time.Time:
-		asDate, err := utils.ToDate(env, c.value)
+		asDate, err := utils.DateFromString(env, c.value)
 		if err != nil {
 			return false, err
 		}
 		return dateComparison(val.(time.Time), c.comparator, asDate)
 
-	case *utils.Location:
```
rowanseymour: locations are passed back just as names now

excellent/errors.go
```diff
@@ -0,0 +1,46 @@
+package excellent
```
rowanseymour: figured while I was reorganizing everything I'd split evaluator.go into evaluator.go, scanner.go and errors.go

contactql/parser.go
rowanseymour: Means `contactql` package only depends on `utils` and not `excellent` or `flows`

Issue #221, opened 2018-04-02
Stringify lists and maps like JSON arrays and objects of their ToString value


rowanseymour:
Currently an array-like object like `@contact.urns` stringifies as CSV.. which for a single item would be something like `tel:+2424521464` - potentially confusing for users as it looks like it is just a string. 

Same applies to map-like objects. It would be better for users to see that these things a multi-value objects and realise that if they want a single item then they should index/resolve to get that.

So...

```
@contact.urns -> ["tel:+2424521464"]
@(to_json(contact.urns)) -> [{"scheme": "tel", "path": "+2424521464"}]

@contact.groups -> ["Testers", "Customers"]
@(to_json(contact.groups)) -> [{"uuid": "1213-2141", "name": "Testers"}, ...]

@contact.fields -> {"age": 1234, "gender": "M"}
@(to_json(contact.fields)) -> {"age": {"text": "34years", "decimal": "34"}, "gender": {...}}

@run.results -> {"color": "red", "joined": "2018-04-02T12:34:67Z"}
@(to_json(run.results)) -> {"color": {"value": "red", "category": "Red", "input": "I like red", ...}, ...}
```

rowanseymour:
Addressed in #228

Issue #218, opened 2018-04-01
Consistent truthyness of complex objects


rowanseymour:
Part of #217 is adding length to map-like things. If we do that then we also have a consistent way of determining truthyness for any complex object that has `Length()`, e.g.

```
@(if(split("a, b, c"), true, false))  // true because has length > 0
@(if(contact.urns, true, false))      // true when length(urns) > 0
@(if(run.results, true, false))       // true when length(results) > 0
@(if(run.webhook.json["things"], true, false))   // JSON fragments also have length
```

`ToBool` wouldn't have to look inside a `JSONFragment` to determine of if it's true - it just calls `Length()` as it does on any `Lengthable`

Also any non-nil `Atomizable` instance should be implicitly true, e.g.

```
@(if(run.input.channel, true, false))  // true if channel is set
```

So `ToBool` would work like..

1. Handle primitives (e.g. any decimal != 0 is true)
2. If val is `Lengthable` then `Length() > 0`
3. If val is non-nil `Atomizable` then `true`
4. `false`

rowanseymour:
Addressed in #228

Issue #215, opened 2018-03-31
Use a static typing system for expressions


rowanseymour:
Maybe a hierarchy of interfaces like:

```
XValue
 * XAtom
    - XNil
    - XError
    - XBool
    - XString
    - XDecimal
    - XTime
    - XArray
 * XAtomizable 
```

Then we can require things like:
  * functions take `XValue...` as arguments and can only return an X type
  * anything coming out of the context is an X type

rowanseymour:
Addressed in #228

PR #228, opened 2018-04-04
Static types

rowanseymour:
Addresses: #215, #218 and #221


excellent/types/base.go
```diff
@@ -0,0 +1,262 @@
...
+	Type() XType
+	ToJSON() XString
+}
+
+// XPrimitive is the base interface of all Excellent primitive types
+type XPrimitive interface {
+	XValue
+
+	ToString() XString
+	ToBool() XBool
```
rowanseymour: `ToString` and `ToBool` are somewhat different to other conversions since every type should be able to render itself as these without blowing up

excellent/types/base.go
```diff
@@ -0,0 +1,262 @@
...
+
+// ToXString converts the given value to a string
+func ToXString(value XValue) XString {
+	switch x := value.(type) {
+	case XPrimitive:
+		return x.ToString()
+	case XObject:
+		return x.Reduce().ToString()
+	}
+	panic(fmt.Sprintf("can't convert type %v to a string", value))
```
rowanseymour: using panics here because if those goes wrong we've coded something wrong - i.e. its more serious than giving the user back an `XError`

excellent/types/base.go
```diff
@@ -0,0 +1,262 @@
...
+// the supported types
+const (
+	XTypeString XType = iota
+	XTypeNumber
+	XTypeBool
+	XTypeTime
+	XTypeArray
+	XTypeObject
+	XTypeError
+	XTypeNil
```
nicpottier: Do we need XType anymore? This was mostly added for equality and we are really only supporting equality in excellent for bool and numeric types, so that's vastly simplified. Seems we could probably remove these and just use normal type assertions instead.

excellent/functions/functions.go
```diff
@@ -41,51 +42,51 @@ var XFUNCTIONS = map[string]XFunction{
 	"legacy_add": LegacyAdd,
 
 	"round":      Round,
-	"round_up":   RoundUp,
-	"round_down": RoundDown,
+	"round_up":   OneNumberFunction("round_up", RoundUp),
```
rowanseymour: @nicpottier here's what functions look like using wrappers - can brainstorm if there's a better way to do this! Does suck having to repeat the function name so that's something to improve.

excellent/visitor.go
```diff
@@ -62,55 +65,63 @@ func (v *Visitor) VisitStringLiteral(ctx *gen.StringLiteralContext) interface{}
...
+	return types.XBoolFalse
 }
 
 // VisitArrayLookup deals with lookups such as foo[5]
 func (v *Visitor) VisitArrayLookup(ctx *gen.ArrayLookupContext) interface{} {
-	context := v.Visit(ctx.Atom())
-	lookup, err := types.ToString(v.env, v.Visit(ctx.Expression()))
-	if err != nil {
-		return err
+	context := toXValue(v.Visit(ctx.Atom()))
```
rowanseymour: @nicpottier would be nice to replace these runtime checks with compile time checks

cmd/exptester/main.go
```diff
@@ -11,22 +11,22 @@ import (
 )
 
 func main() {
-	vars := make(map[string]interface{})
-	vars["int1"] = 1
-	vars["string1"] = "string1"
-	vars["int2"] = 2
+	vars := types.NewXMap(map[string]types.XValue{
```
rowanseymour: unlike `XArray`, users can't create these but we use them internally

excellent/types/error.go
```diff
@@ -0,0 +1,58 @@
...
+	return xerror{err: err}
+}
+
+// NewXErrorf creates a new XError
+func NewXErrorf(format string, a ...interface{}) XError {
+	return NewXError(fmt.Errorf(format, a...))
+}
+
+// NewXResolveError creates a new XError when a key can't be resolved on an XResolvable
+func NewXResolveError(resolvable XResolvable, key string) XError {
```
rowanseymour: we could create several of these factory methods, so that errors of same type are created the same way. Then at some point we could add error codes or the like to make it easier for consumers like the simulator to localize these

excellent/types/conversions.go
```diff
@@ -1,303 +1,138 @@
...
-	if arg1Type != arg2Type {
-		return 0, fmt.Errorf("Cannot compare different types of %#v and %#v", arg1, arg2)
-	}
+	return int(intPart), nil
+}
 
-	arg1Str := fmt.Sprintf("%v", arg1)
-	arg2Str := fmt.Sprintf("%v", arg2)
+// RequireMarshalToXString calls json.Marshal in the given value and panics in the case of an error
+func RequireMarshalToXString(x interface{}) XString {
```
nicpottier: I like this, though I think the golang convention is `must` instead of `require`

rowanseymour: can change - copied Require from `decimal.RequireFromString`

nicpottier: Oh, well maybe I'm wrong, maybe worth checking what is used more, no biggie either way.

rowanseymour: I think you're right about Must being more standard - and `MustMarshalToXString` sounds marginally better

PR #232, opened 2018-04-11
XValues should be JSONable via json.Marshal

rowanseymour:
An `XValue` can have two different JSON representations:

 1. our internal representation, created via the usual json.Marshal methods
 2. the public representation returned from `@(to_json(...))`

### DIFF
--- a/cmd/flowrunner/testdata/flows/dynamic_groups_test.json
+++ b/cmd/flowrunner/testdata/flows/dynamic_groups_test.json
@@ -81,7 +81,7 @@
                 "contact": {
                     "fields": {
                         "age": {
-                            "decimal": "17",
+                            "decimal": 17,
                             "text": "17"
                         },
                         "first_name": {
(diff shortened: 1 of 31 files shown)


cmd/flowrunner/testdata/flows/dynamic_groups_test.json
```diff
@@ -156,7 +156,7 @@
                                         "name": "Android Channel",
                                         "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d"
                                     },
-                                    "text": "Current groups: Males, Old Men",
+                                    "text": "Current groups: [\"Males\",\"Old Men\"]",
```
rowanseymour: unrelated - just forgot to regenerate test files after #228

excellent/types/array.go
```diff
@@ -18,6 +18,9 @@ type xarray struct {
 
 // NewXArray returns a new array with the given items
 func NewXArray(values ...XValue) XArray {
+	if values == nil {
+		values = []XValue{}
```
rowanseymour: so that empty array always marshals as [] rather than null

nicpottier:
Sorry, I'm confused, when are these different?

rowanseymour:
@nicpottier remember we talked about contact JSON being different? one having exploded URNs?

nicpottier:
oh right right, this is excellent json vs goflow json.

rowanseymour:
@nicpottier fine with `ToXJSON` - don't think we should introduce another type since it is just an XString as far as Excellent is concerned

PR #236, opened 2018-04-12
Implement ToXJSON on all XValue types

rowanseymour:


cmd/flowrunner/testdata/flows/two_questions.json
```diff
@@ -200,7 +200,7 @@
                             "type": "call_webhook",
                             "url": "http://localhost/?cmd=success",
                             "method": "POST",
-                            "body": "{ \"contact\": @(to_json(contact.uuid)), \"soda\": @(to_json(run.results.soda)) }"
+                            "body": "{ \"contact\": @(to_json(contact.uuid)), \"soda\": @(to_json(run.results.soda.value)) }"
```
rowanseymour: this is so that we still just get the value - `@(to_json(run.results.soda))` includes category etc

excellent/evaluator_test.go
```diff
@@ -53,6 +41,11 @@ func (v *testXObject) Resolve(key string) types.XValue {
 	}
 }
 
+// ToXJSON is called when this type is passed to @(to_json(...))
+func (v *testXObject) ToXJSON() types.XString {
+	return types.ResolveKeys(v, "foo", "bar").ToXJSON()
```
rowanseymour: Using this to generate the XJSON ensures that each included field has the exact same value as what you'd get if you JSONified that individual field.

flows/inputs/msg.go
```diff
@@ -17,7 +17,7 @@ const TypeMsg string = "msg"
 // MsgInput is a message which can be used as input
 type MsgInput struct {
 	baseInput
-	urn         urns.URN
+	urn         *flows.ContactURN
```
rowanseymour: so that the URN at @run.input.urn behaves the same way as other context URNs in expressions

flows/engine/session_test.go
```diff
@@ -77,6 +80,53 @@ func TestEvaluateTemplateAsString(t *testing.T) {
...
+		{"run.input", `{"attachments":[{"content_type":"image/jpeg","url":"http://s3.amazon.com/bucket/test_en.jpg?a=Azuay"}],"channel":{"address":"+12345671111","name":"Nexmo","uuid":"57f1078f-88aa-46f4-a59a-948a5739c03d"},"created_on":"2000-01-01T00:00:00.000000Z","text":"Hi there","type":"msg","urn":{"display":"","path":"+12065551212","scheme":"tel"},"uuid":"84f8a3cf-0f2c-4881-9502-2d7b114bf01f"}`},
+
+		// TODO add way to mock call calls to Now() so we can have deterministic tests without doing text substitution of dates?
```
rowanseymour: this would be nice too for the tests for flowrunner and not so different to how we mock new UUID generation

PR #248, opened 2018-04-16
String, bool and date equality

### LINKED ISSUE #245: String equality/inequality
In #212 we made `=`, `!=`, `>=`, `>`, `<=`, `<` require numbers as arguments rather than the legacy behavior of doing either numerical or lexical comparison based on the argument types. We added `string_cmp` which is fine for the unlikely case that someone needs to do lexical comparison of values, but `string_cmp(foo, bar) == 0` is hardly ideal for the common case of `foo = bar` where both args are strings.

We also have no easy migration path from an expression like `@(contact.language = "eng")`

rowanseymour:
Fixes #245

### DIFF
--- a/excellent/evaluator_test.go
+++ b/excellent/evaluator_test.go
@@ -221,13 +221,27 @@ func TestEvaluateTemplate(t *testing.T) {
 		{"@(4*3/4)", xi(3), false},
 		{"@(4/2*4)", xi(8), false},
 		{"@(2^2^2)", xi(16), false},
-		{"@(11=11=11)", xs(""), true},
-		{"@(1<2<3)", xs(""), true},
 		{"@(\"a\" & \"b\" & \"c\")", xs("abc"), false},
 		{"@(1+3 <= 1+4)", types.XBoolTrue, false},
 
+		// numerical equality
 		{"@((1 = 1))", types.XBoolTrue, false},
 		{"@((1 != 2))", types.XBoolTrue, false},
+		{"@(1 = 1)", types.XBoolTrue, false},
+		{"@(1 = 2)", types.XBoolFalse, false},
+		{"@(1 != 2)", types.XBoolTrue, false},
+		{"@(1 != 1)", types.XBoolFalse, false},
+		{"@(-1 = 1)", types.XBoolFalse, false},
+		{"@(11=11=11)", types.XBoolFalse, false}, // 11=11 -> TRUE, then TRUE != 11
+
+		// string equality
+		{`@("asdf" = "asdf")`, types.XBoolTrue, false},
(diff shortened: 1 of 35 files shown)


excellent/functions/functions.go
```diff
@@ -202,6 +203,21 @@ func Array(env utils.Environment, args ...types.XValue) types.XValue {
...
 }
 
+// Number tries to convert `value` to a number. An error is returned if the value can't be converted.
+//
+//   @(number(10)) -> 10
+//   @(number("123.45000")) -> 123.45
+//   @(number("what?")) -> ERROR
+//
+// @function number(value)
+func Number(env utils.Environment, value types.XValue) types.XValue {
```
rowanseymour: was going to call this `to_number` but `number` seems more natural and is consistent with `date` which essentially does the same thing

excellent/evaluator_test.go
```diff
@@ -221,13 +221,43 @@ func TestEvaluateTemplate(t *testing.T) {
...
+		{"@(1 = 1)", types.XBoolTrue, false},
+		{"@(1 = 2)", types.XBoolFalse, false},
+		{"@(1 != 2)", types.XBoolTrue, false},
+		{"@(1 != 1)", types.XBoolFalse, false},
+		{"@(-1 = 1)", types.XBoolFalse, false},
+		{"@(1.0 = 1)", types.XBoolTrue, false},
+		{"@(1.1 = 1.10)", types.XBoolTrue, false},
+		{"@(1.1234 = 1.10)", types.XBoolFalse, false},
+		{`@(1 = number("1.0"))`, types.XBoolTrue, false},
+		{"@(11=11=11)", types.XBoolFalse, false}, // 11=11 -> TRUE, then TRUE != 11
```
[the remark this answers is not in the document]
rowanseymour: Go works similarly (first `11==11` -> `true`), but Go doesn't let you compare a bool and an int so it errors:

```
cannot convert 11 (type untyped number) to type bool
invalid operation: 11 == 11 == 11 (mismatched types bool and int)
```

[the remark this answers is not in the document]
nicpottier: Ya that seems totally fine.

excellent/functions/functions.go
nicpottier: I'm ok with that but then we lose the symmetry of `from_` and `to_`. What are we going to call the thing that parses JSON?

rowanseymour: I don't forsee that being a commonly function since JSON typically enters excellent via something like `run.webhook.json` and I just had to look at the code because I couldn't remember what it returns (JSON from a string). I think I prefer something like `parse_json`.

nicpottier: works for me.

PR #256, opened 2018-04-17
Fix contact timezone defaulting to UTC

### LINKED ISSUE #250: Dates should be parsed in the environment timezone
See `parse_date` examples which shouldn't be UTC

rowanseymour:
It should default to the session environment timezone

Fixes #250

### DIFF
--- a/utils/dates.go
+++ b/utils/dates.go
@@ -121,11 +121,8 @@ func DateToString(env Environment, date time.Time) string {
 func DateFromString(env Environment, str string) (time.Time, error) {
 	// first see if we can parse in any known iso formats, if so return that
 	for _, format := range isoFormats {
-		parsed, err := time.Parse(format, strings.Trim(str, " \n\r\t"))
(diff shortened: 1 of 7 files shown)


utils/dates.go
```diff
@@ -121,11 +121,8 @@ func DateToString(env Environment, date time.Time) string {
 func DateFromString(env Environment, str string) (time.Time, error) {
 	// first see if we can parse in any known iso formats, if so return that
 	for _, format := range isoFormats {
-		parsed, err := time.Parse(format, strings.Trim(str, " \n\r\t"))
+		parsed, err := time.ParseInLocation(format, strings.Trim(str, " \n\r\t"), env.Timezone())
```
rowanseymour: if the string contains an explicit tz, we should keep that

nicpottier: How does a user get something in the timezone of their choice from a UTC date again?

PR #262, opened 2018-04-20
Empty values

rowanseymour:
Introduces the concept of _empty_ to determine what `default()` and `has_value()` should return

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -600,12 +600,15 @@ Converts the passed in `year`, `month` and `day`
 
 ## default(test, default)
 
-Takes two arguments, returning `test` if not an error or nil, otherwise returning `default`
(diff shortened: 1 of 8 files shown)


docs/docs.md
```diff
@@ -600,12 +600,15 @@ Converts the passed in `year`, `month` and `day`
...
-Takes two arguments, returning `test` if not an error or nil, otherwise returning `default`
+Takes two arguments, returning `test` if not an error or nil or empty text, otherwise returning `default`
 
 
 ```objectivec
 @(default(undeclared.var, "default_value")) → default_value
 @(default("10", "20")) → 10
+@(default("", "value")) → value
+@(default(array(1, 2), "value")) → ["1","2"]
+@(default(array(), "value")) → value
```
rowanseymour: not 100% about this but feels consistent treatment for empty strings... @nicpottier ?

PR #271, opened 2018-04-24
Webhook test

### LINKED ISSUE #270: Legacy webhook ruleset create different run results
Legacy webhook rulesets use a `WebhookStatusTest` which saves the webhook response body as the match, and leaves input empty, e.g.

```json
{
  "category": "Success",
  "node_uuid": "9b5147c2-fa6c-4ef2-8684-4bf20c2c6fcb",
  "name": "Response 1",
  "value": "{ \"text\": \"Get\", \"blank\": \"\" }",
  "created_on": "2018-04-23T22:37:39.071144546Z",
  "input": ""
}
```

New webhook rulesets us a `is_text_eq` test on the webhook status, so their match and input is the status string, e.g.

```json
{
  "category": "Success",
  "node_uuid": "9b5147c2-fa6c-4ef2-8684-4bf20c2c6fcb",
  "name": "Response 1",
  "value": "success",
  "created_on": "2018-04-23T22:37:39.070738018Z",
  "input": "success"
}
```

Two possible solutions:

 1. Add a webhook status test that replicates behavior of `WebhookStatusTest`
 2. Add an additional `set_run_result` action to explicitly save a result like what `WebhookStatusTest` would produce.

rowanseymour:
Introduces `has_webhook_status` test which produces results like:

```json
"response_1": {
    "category": "Success",
    "node_uuid": "87000b5e-7cdb-4e6f-8d85-4c34c2a6f6c5",
    "name": "Response 1",
    "value": "{ \"text\": \"Get\", \"blank\": \"\" }",
    "created_on": "2018-04-24T14:52:16.33261135Z",
    "input": "GET http://localhost:49999/check_order.php?phone=%2B250788383383"
}
```

Note that in legacy engine input is blank - but having something here seems preferable and means we don't have to treat webhooks as a special case - result input is always the test operand stringified. Could update legacy webhooks to match this behaviour.

Fixes #270

### DIFF
--- a/cmd/flowrunner/testdata/flows/webhook_persists.json
+++ b/cmd/flowrunner/testdata/flows/webhook_persists.json
@@ -48,7 +48,7 @@
                         }
                     ],
                     "router": {
-                        "operand": "@run.webhook.status",
+                        "operand": "@run.webhook",
                         "cases": [
                             {
                                 "arguments": [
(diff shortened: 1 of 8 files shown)


nicpottier:
Nice, I really like that input is the method / url

PR #272, opened 2018-04-24
Configurable limit on webhook response bodies

rowanseymour:


cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -678,7 +678,7 @@
                         "webhook": {
                             "body": "{ \"ok\": \"true\" }",
                             "request": "GET /?cmd=success HTTP/1.1\r\nHost: 127.0.0.1:49999\r\nUser-Agent: goflow-testing\r\nAccept-Encoding: gzip\r\n\r\n",
-                            "response": "HTTP/1.1 200 OK\r\nContent-Length: 16\r\nContent-Type: text/plain; charset=utf-8\r\nDate: Wed, 11 Apr 2018 18:24:30 GMT\r\n\r\n{ \"ok\": \"true\" }",
+                            "response": "HTTP/1.1 200 OK\r\nContent-Length: 16\r\nContent-Type: text/plain; charset=utf-8\r\nDate: Wed, 11 Apr 2018 18:24:30 GMT\r\n\r\n",
```
rowanseymour: since body is included separately above - seems preferable to not duplicate it here

flows/webhook.go
```diff
@@ -150,34 +161,36 @@ func newWebhookCallFromResponse(requestTrace string, r *http.Response) (*Webhook
...
+		// if we have no remaining bytes, error because the body was too big
+		if bodyReader.(*io.LimitedReader).N <= 0 {
+			return nil, fmt.Errorf("webhook response body exceeds %d bytes limit", maxBodyBytes)
+		}
+
+		w.body = string(bodyBytes)
 	} else {
 		// no body for non-text responses but add it to our Response log so users know why
-		w.response = w.response + "\nNon-text body, ignoring"
+		w.response = w.response + "Non-text body, ignoring"
```
rowanseymour: wondering if there's a better place for this?

flows/webhook.go
```diff
@@ -150,34 +161,36 @@ func newWebhookCallFromResponse(requestTrace string, r *http.Response) (*Webhook
...
+	w.response = string(responseDump)
+
+	// only save response body's if we have a supported content-type
+	contentType := response.Header.Get("Content-Type")
+	mediaType, _, _ := mime.ParseMediaType(contentType)
+	saveBody := saveResponseContentTypes[mediaType]
+
+	if saveBody {
+		// only read up to our max body bytes limit
+		bodyReader := io.LimitReader(response.Body, int64(maxBodyBytes)+1)
```
rowanseymour: so even if a webhook returns a 10GB video - we only read up to a certain number of bytes

cmd/flowrunner/testdata/flows/all_actions_test.json
nicpottier: Agree we don't want it twice.. but this seems awkward as request and response don't really pair well now. Maybe we should only have response and parse it for body when it is asked for?

rowanseymour:
@nicpottier now parsing method and body from the response trace as needed, and added some unit tests... and somehow coverage went +3.9%

Issue #274, opened 2018-04-27
Test evaluation of migrated legacy templates


rowanseymour:
See `TestLegacyTests`. This will require remapping the provided variables dict, e.g.

```json
{"flow": {"foo": "bar"}}
```

to

```json
{"run": {"results": {"foo": "bar"}}}
```

Issue #273, opened 2018-04-27
Panic when migrating template with unterminated text literal


rowanseymour:
Have removed this test from `legacy_tests.json` for now:

```
{
    "template": "@(\")",
    "context": {
        "variables": {},
        "timezone": "UTC",
        "date_style": "day_first"
    },
    "url_encode": false,
    "output": "@(\")",
    "errors": []
},
```

```
expressions_test.go:270: panic migrating template '@(")': &runtime.TypeAssertionError{interfaceString:"", concreteString:"", assertedString:"antlr.ParserRuleContext", missingMethod:""}
```

PR #275, opened 2018-04-27
Legacy expression migration fixes

rowanseymour:
* Use a separate grammar/visitor for legacy and new expressions - this allows us to better handle the differences, e.g. `<>` vs `!=`, `TRUE()` vs `true`, and keep new features out of the old grammar, e.g. array indexing `[0]`, `NULL` keyword
 * Ensure all legacy functions map to a new function
 * Test trying to migrate the [test templates](https://github.com/rapidpro/expressions/blob/master/common/template_tests.json) from legacy expression engine. Note for now I've removed some of these tests because they cause panics, see #273 and it will require a bit more work to test evaluating these templates (see #274)


docs/docs.md
```diff
@@ -991,20 +1003,20 @@ A single random integer in the given inclusive range.
 @(rand_between(1, 10)) → 10
 ```
 
-<a name="function:read_code"></a>
+<a name="function:read_digits"></a>
 
-## read_code(code)
+## read_digits(code)
```
rowanseymour: Old function name here `read_digits` seems clearer

nicpottier: I think the change was because it isn't number specific.. IE, could be 6 alphanums, but you want them spaced out.

rowanseymour: Ah ok... I can revert or can we think of something more specific than "code"... "chars" ?

rowanseymour:
@nicpottier ok going with `read_chars`. we already have `code()` to get codepoint of a character

PR #279, opened 2018-05-02
Locations

### LINKED ISSUE #252: Location support
We had some location support but haven't updated it since rapidpro switched to using location paths (country > state > district > ward)

rowanseymour:
* Use > paths
 * Implement same parsing logic as RP

Fixes #252


flows/assets/cache.go
```diff
@@ -134,7 +134,7 @@ func readAsset(data json.RawMessage, itemType assetType) (interface{}, error) {
 	var assetReader func(data json.RawMessage) (interface{}, error)
 
 	if itemType == assetTypeLocationHierarchy {
-		assetReader = func(data json.RawMessage) (interface{}, error) { return flows.ReadLocationHierarchy(data) }
+		assetReader = func(data json.RawMessage) (interface{}, error) { return utils.ReadLocationHierarchy(data) }
```
rowanseymour: Location stuff moves back into utils as Location objects are no longer exposed in expressions etc - only their paths

flows/contact.go
```diff
@@ -262,7 +264,7 @@ func (c *Contact) ResolveQueryKey(key string) []interface{} {
 			switch typed := fieldValue.(type) {
 			case nil:
 				return nil
-			case *Location:
```
rowanseymour: this where its useful to have `LocationPath`  be a type rather than just a string

flows/fields.go
```diff
@@ -113,26 +114,94 @@ func (f FieldValues) clone() FieldValues {
...
+		text:     asText,
+		datetime: asDateTime,
 		number:   asNumber,
+		state:    asState,
+		district: asDistrict,
+		ward:     asWard,
+	}
+}
+
+func (f FieldValues) getFirstLocationValue(env RunEnvironment, fieldSet *FieldSet, valueType FieldValueType) *utils.Location {
```
rowanseymour: bit of awkward legacy logic

PR #286, opened 2018-05-03
Include contact name on migrate

ericnewcomer:


legacy/definition.go
```diff
@@ -127,10 +127,11 @@ func (l *LabelReference) UnmarshalJSON(data []byte) error {
 
 type ContactReference struct {
 	UUID flows.ContactUUID `json:"uuid"`
+	Name string
```
rowanseymour: `json:"name"`

PR #289, opened 2018-05-04
Timeouts, waits and inputs

rowanseymour:


cmd/flowrunner/testdata/flows/all_actions_test.json
```diff
@@ -670,7 +670,6 @@
                             "gender": {
                                 "category": "Male",
                                 "created_on": "2000-01-01T00:00:00.000000000-00:00",
-                                "input": "",
```
rowanseymour: result.input is now nullable if there is no input - but will be omitted when null - included when ""

cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -155,6 +155,7 @@
                     "type": "manual"
                 },
                 "wait": {
+                    "timeout": 600,
```
rowanseymour: waits now keep the timeout duration around - callers should use this to control how long they wait for - and that way it doesn't matter if there's time differences between caller and engine

cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -165,9 +166,9 @@
                 {
                     "category": "",
                     "created_on": "2000-01-01T00:00:00.000000000-00:00",
+                    "input": "Ryan Lewis",
```
rowanseymour: we renamed operand to input to match RP on results, but we should also do here on the event

cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -406,10 +407,6 @@
                     },
                     "triggered_on": "2000-01-01T00:00:00Z",
                     "type": "manual"
-                },
```
rowanseymour: there was a bug previously where waits wouldn't be cleared so runs would end with waits

excellent/evaluator.go
```diff
@@ -68,12 +68,6 @@ func EvaluateTemplate(env utils.Environment, context types.XValue, template stri
 	switch tokenType {
 	case IDENTIFIER:
 		value := ResolveValue(env, context, token)
-
```
rowanseymour: removing this so that expressions can evaluate to nil - i.e. `@run.input` as an operand when there is no input

flows/contact.go
```diff
@@ -53,9 +53,9 @@ type Contact struct {
 }
 
 // NewContact returns a new contact
-func NewContact(uuid ContactUUID, name string, language utils.Language, timezone *time.Location) *Contact {
+func NewContact(name string, language utils.Language, timezone *time.Location) *Contact {
```
rowanseymour: this function is just for tests...

flows/routers/tests/tests.go
```diff
@@ -146,8 +147,20 @@ func HasWaitTimedOut(env utils.Environment, value types.XValue) types.XValue {
 		return types.NewXErrorf("must be called with a run as first argument")
 	}
 
-	if run.Session().Wait() != nil && run.Session().Wait().HasTimedOut() {
-		return XTestResult{true, nil}
+	// look to see if the last input event was a message or a timeout
+	runEvents := run.Events()
```
rowanseymour: previously this test worked by examining the wait on the session which is why we could never clear waits - now it checks events, and waits are cleared on flow resumption

PR #292, opened 2018-05-09
Better error messages from expression evaluation

rowanseymour:


utils/misc.go
```diff
@@ -20,3 +20,11 @@ func IsNil(v interface{}) bool {
 
 	return false
 }
+
+// MinInt returns the minimum of two integers
+func MinInt(x, y int) int {
```
rowanseymour: only in go you write your own `min` function...

cmd/flowserver/server_test.go
```diff
@@ -303,7 +303,7 @@ func (ts *ServerTestSuite) TestExpression() {
 	// try the expression endpoint with a missing variable
 	status, body = ts.testHTTPRequest("POST", "http://localhost:8800/expression", `{"expression": "@(foo + 2)", "context": {}}`)
 	ts.Equal(200, status)
-	ts.assertExpressionResponse(body, "", []string{"unable to resolve 'foo'"})
+	ts.assertExpressionResponse(body, "", []string{"error evaluating @(foo + 2): json object has no property 'foo'"})
```
rowanseymour: templates can have multiple expressions so error messages now include the expression they came from

nicpottier: ooh nice

utils/misc.go
nicpottier: heh

PR #294, opened 2018-05-11
Expression migrating fixes

rowanseymour:
* Fix `@` at end of expression
 * Fix migrating `@@` escape sequences
 * Fix `clean()` to match old behaviour
 * Fix panics when migrating expressions with errors https://github.com/nyaruka/goflow/issues/273

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -514,12 +514,11 @@ Returns the rune for the passed in codepoint, `num`, which may be unicode, this 
 
 ## clean(text)
 
-Strips any leading or trailing whitespace from `text`
+Strips any non-printable characters from `text`
 
 
 ```objectivec
-@(clean("\nfoo\t")) → foo
-@(clean(" bar")) → bar
+@(clean("😃 Hello \nwo\tr\rld")) → 😃 Hello world
 @(clean(123)) → 123
 ```
 
--- a/excellent/scanner.go
+++ b/excellent/scanner.go
@@ -37,6 +37,7 @@ type xscanner struct {
 	unreadRunes         []rune
 	unreadCount         int
 	identifierTopLevels []string
+	unescapeBody        bool // unescape @@ sequences in the body
 }
 
 // NewXScanner returns a new instance of our excellent scanner
@@ -45,7 +46,12 @@ func NewXScanner(r io.Reader, identifierTopLevels []string) *xscanner {
 		reader:              bufio.NewReader(r),
 		unreadRunes:         make([]rune, 4),
 		identifierTopLevels: identifierTopLevels,
+		unescapeBody:        true,
 	}
+}
+
+func (s *xscanner) SetUnescapeBody(unescape bool) {
+	s.unescapeBody = unescape
 }
 
 // gets the next rune or EOF if we are at the end of the string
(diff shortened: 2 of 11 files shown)


docs/docs.md
```diff
@@ -514,12 +514,11 @@ Returns the rune for the passed in codepoint, `num`, which may be unicode, this
 
 ## clean(text)
 
-Strips any leading or trailing whitespace from `text`
+Strips any non-printable characters from `text`
```
rowanseymour: maybe we should have a separate `strip()` function that does what this was doing?

excellent/scanner.go
```diff
@@ -37,6 +37,7 @@ type xscanner struct {
 	unreadRunes         []rune
 	unreadCount         int
 	identifierTopLevels []string
+	unescapeBody        bool // unescape @@ sequences in the body
```
rowanseymour: so that when we're using this for expression migrating, we can leave the `@@` sequences in the body

legacy/expressions/migrate.go
```diff
@@ -30,19 +29,20 @@ func MigrateTemplate(template string, extraAs ExtraVarsMapping) (string, error)
 	return migrateLegacyTemplateAsString(migrationVarMapper, template)
 }
 
-func migrateLegacyTemplateAsString(resolver types.XValue, template string) (string, error) {
+func migrateLegacyTemplateAsString(resolver Resolvable, template string) (string, error) {
```
rowanseymour: simpler to have the migrator visitor not try to use xvalues for everything

legacy/expressions/migrate.go
```diff
@@ -115,7 +115,6 @@ func migrateExpression(env utils.Environment, resolver types.XValue, expression
 	p.AddErrorListener(errListener)
 
 	// speed up parsing
-	p.SetErrorHandler(antlr.NewBailErrorStrategy())
```
rowanseymour: not sure if its a problem in ANTLR, but this causes panics when there are syntax errors

legacy/expressions/migrate_test.go
```diff
@@ -300,35 +344,40 @@ func TestLegacyTests(t *testing.T) {
 	require.NoError(t, err)
 
 	var tests []legacyTest
-	err = json.Unmarshal(legacyTestData, &tests)
+	d := json.NewDecoder(bytes.NewReader(legacyTestData))
+	d.UseNumber()
```
rowanseymour: means numbers are parsed as strings -by default it was using `float64` but that was causing failing tests like 6.23 != 6.299999....

legacy/expressions/migrate_test.go
```diff
@@ -300,35 +344,40 @@ func TestLegacyTests(t *testing.T) {
...
+
+			migratedVars := tc.Context.Variables.Migrate()
+			migratedVarsJSON, _ := json.Marshal(migratedVars)
+
+			_, err = excellent.EvaluateTemplateAsString(env, migratedVars, migratedTemplate, tc.URLEncode, runs.RunContextTopLevels)
+
+			if len(tc.Errors) > 0 {
+				assert.Error(t, err, "expecting error evaluating template '%s' (migrated from '%s') with context %s", migratedTemplate, tc.Template, migratedVarsJSON)
+			} else {
+				// TODO enable checking of output
```
rowanseymour: Enabling this produces a fair number of failures - some of which are clearly not important (`TRUE` vs `true`) but will continue looking through them to find ones that do matter (see https://github.com/nyaruka/goflow/issues/274)

legacy/expressions/migrate.go
nicpottier: Did you benchmark this at all? Wasn't this the thing causing crazy performance degradation. We can catch the panic it if comes to that.

PR #296, opened 2018-05-14
Allow webhook calls to be disabled

rowanseymour:


flows/engine/config.go
```diff
@@ -1,17 +1,51 @@
...
+	return &config{disableWebhooks: false, maxWebhookResponseBytes: 10000}
 }
 
+func (c *config) DisableWebhooks() bool        { return c.disableWebhooks }
 func (c *config) MaxWebhookResponseBytes() int { return c.maxWebhookResponseBytes }
+
+type configEnvelope struct {
+	DisableWebhooks         *bool `json:"disable_webhooks"`
+	MaxWebhookResponseBytes *int  `json:"max_webhook_response_bytes"`
+}
```
rowanseymour: pointers so that all of these things can be optional

flows/webhook.go
```diff
@@ -68,7 +68,16 @@ type WebhookCall struct {
 // MakeWebhookCall fires the passed in http request, returning any errors encountered. RequestResponse is always set
 // regardless of any errors being set
 func MakeWebhookCall(session Session, request *http.Request) (*WebhookCall, error) {
-	response, requestDump, err := session.HTTPClient().DoWithDump(request)
+	var response *http.Response
+	var requestDump string
+	var err error
+
+	if session.EngineConfig().DisableWebhooks() {
+		response, requestDump, err = session.HTTPClient().MockWithDump(request, "MOCKED")
```
rowanseymour: this is where we'd do configurable mock requests for each URL/method in future

PR #306, opened 2018-05-26
Remove all groups

### LINKED ISSUE #304: Allow flow actions like remove_from_group with no groups?
We sometimes get these in legacy flows but they fail goflow validation

rowanseymour:
Fixes #304

### DIFF
--- a/legacy/definition.go
+++ b/legacy/definition.go
@@ -537,6 +537,7 @@ func migrateAction(baseLanguage utils.Language, a Action, localization flows.Loc
 
 		return &actions.RemoveContactGroupsAction{
 			Groups:     groups,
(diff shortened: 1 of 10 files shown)


legacy/definition.go
```diff
@@ -537,6 +537,7 @@ func migrateAction(baseLanguage utils.Language, a Action, localization flows.Loc
 
 		return &actions.RemoveContactGroupsAction{
 			Groups:     groups,
+			AllGroups:  len(groups) == 0,
```
ericnewcomer: so old flows represented this as an empty list then?

rowanseymour: Yup, checked it

PR #311, opened 2018-06-11
Migrate rulesets where there is an explicit Other category

rowanseymour:
You can create an explicit Other category (e.g. to match "other" as a keyword) and it doesn't merge with the implicit Other category. So when we migrate, these need to kept as distinct exits.

captura de pantalla 2018-06-11 a la s 17 15 43

### DIFF
--- a/legacy/testdata/rulesets.json
+++ b/legacy/testdata/rulesets.json
@@ -234,6 +234,105 @@
     },
     {
         "legacy_ruleset": {
+            "uuid": "10e483a8-5ffb-4c4f-917b-d43ce86c1d65", 
+            "rules": [
+                {
+                    "uuid": "ea304225-332e-49d4-9768-1e804cd0b6c2", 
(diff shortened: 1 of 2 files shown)


legacy/testdata/rulesets.json
```diff
@@ -232,6 +232,105 @@
...
+                },
+                {
+                    "uuid": "633dfe67-988e-4190-a95f-70f477494032",
+                    "destination_node_uuid": "5b977652-91e3-48be-8e86-7c8094b4aa8f",
+                    "name": "Other"
+                },
+                {
+                    "uuid": "1fc4c133-d038-4f75-a69e-6e7e3190e5d8",
+                    "destination_node_uuid": "833fc698-d590-42dc-93e1-39e701b7e8e4",
+                    "name": "Other"
```
ericnewcomer: 🤷‍♀️

Issue #312, opened 2018-06-20
Support dynamic groups based on created_on and language


rowanseymour:

rowanseymour:
Fixed in #315

PR #315, opened 2018-06-20
Dynamic groups on created_on and language

rowanseymour:
Addresses #312

### DIFF
--- a/test/environment.go
+++ b/test/environment.go
@@ -3,6 +3,7 @@ package test
 import (
 	"time"
 
+	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/utils"
 )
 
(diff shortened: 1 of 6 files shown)


test/environment.go
```diff
@@ -33,3 +34,17 @@ func (e *TestEnvironment) Now() time.Time {
 func (e *TestEnvironment) SetNow(now time.Time) {
 	e.now = now
 }
+
+func (e *TestEnvironment) FindLocations(string, utils.LocationLevel, *utils.Location) ([]*utils.Location, error) {
```
rowanseymour: so that this env type can be used as a RunEnvionment which is required by FieldValues.setValue because it potentially looks up locations.

Issue #238, opened 2018-04-12
Simplify function call migrations that result in X - 1 parameters


rowanseymour:
`@(WORD(foo, 2))` in legacy expression becomes `@(word(foo, 2 - 1))` in new expression - would be nice to simplify `2 - 1` to just `1`

rowanseymour:
Added in #317

PR #317, opened 2018-06-29
🚖 Improvements to legacy function migration

### LINKED ISSUE #237: word, word_count and word_slice can all take a boolean final arg in legacy expressions
It defaults to false and indicates that tokenizing should only happen by whitespace. Need to be able to migrate this or verify it's not being used.

rowanseymour:
Fixes #237 and #238


legacy/expressions/migrate_test.go
```diff
@@ -130,15 +130,21 @@ func TestMigrateTemplate(t *testing.T) {
 
 		// functions
 		{old: `@(REMOVE_FIRST_WORD(flow.favorite_color))`, new: `@(remove_first_word(run.results.favorite_color))`},
-		{old: `@(WORD_SLICE(flow.favorite_color, 2))`, new: `@(word_slice(run.results.favorite_color, 2 - 1))`},
-		{old: `@(WORD_SLICE(flow.favorite_color, 2, 4))`, new: `@(word_slice(run.results.favorite_color, 2 - 1, 4 - 1))`},
-		{old: `@(FIELD(flow.favorite_color, 2, ","))`, new: `@(field(run.results.favorite_color, 2 - 1, ","))`},
+		{old: `@(WORD_SLICE(flow.favorite_color, 2))`, new: `@(word_slice(run.results.favorite_color, 1))`},
```
rowanseymour: we now always simplify things like `2 - 1` to `1`

excellent/functions/functions.go
```diff
@@ -471,14 +506,9 @@ func WordCount(env utils.Environment, text types.XText) types.XValue {
...
-// @function field(text, offset, delimeter)
-func Field(env utils.Environment, args ...types.XValue) types.XValue {
-	source, xerr := types.ToXText(env, args[0])
-	if xerr != nil {
-		return xerr
-	}
-
-	field, xerr := types.ToInteger(env, args[1])
+// @function field(text, offset, delimiter)
+func Field(env utils.Environment, text types.XText, args ...types.XValue) types.XValue {
```
rowanseymour: not sure what to do with `Field`.. if we change it to use the same logic as the other tokenizing functions then it won't parse things like `1,,3` as 3 items with the 2nd item being empty

nicpottier: Ya I think field has a very specific meaning where that behavior needs to be maintained.

PR #321, opened 2018-07-03
💣 Panic recovery middleware

rowanseymour:
and protect against nil pointer panics in type conversion functions

### DIFF
--- a/cmd/flowserver/middleware.go
+++ b/cmd/flowserver/middleware.go
@@ -1,13 +1,9 @@
 package main
 
 import (
-	"bytes"
 	"fmt"
-	"io"
-	"io/ioutil"
 	"net/http"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/go-chi/chi/middleware"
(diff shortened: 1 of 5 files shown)


cmd/flowserver/middleware.go
```diff
@@ -44,23 +40,3 @@ func requestLogger() func(next http.Handler) http.Handler {
...
-func traceErrors() func(next http.Handler) http.Handler {
-	return func(next http.Handler) http.Handler {
-		fn := func(w http.ResponseWriter, r *http.Request) {
-			body := bytes.Buffer{}
-			r.Body = ioutil.NopCloser(io.TeeReader(r.Body, &body))
-			ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
-			next.ServeHTTP(ww, r)
-
-			// we are returning a system error of some kind, log the incoming request body
-			if ww.Status() >= 500 && strings.ToLower(r.Method) == "post" {
```
rowanseymour: so this never happens - panics end up being converted to 502s at the nginx level

excellent/types/text.go
```diff
@@ -82,5 +82,10 @@ func ToXText(env utils.Environment, x XValue) (XText, XError) {
 		return XTextEmpty, x.(XError)
 	}
 
-	return x.Reduce(env).ToXText(env), nil
+	primitive, isPrimitive := x.(XPrimitive)
+	if isPrimitive {
+		return primitive.ToXText(env), nil
+	}
+
+	return ToXText(env, x.Reduce(env))
```
rowanseymour: `Reduce` could return a nil in the case of unset field value etc

cmd/flowserver/middleware.go
nicpottier: are we still logging the request somewhere? Seems `Recoverer` actually traps the panic then changes the response to a 500 so maybe we were just missing that piece?

rowanseymour: Maybe - but I've now changed it to use a combination of raven's own recovery middleware (not made for chi) and chi's own implementation. The recovery middleware comes before the request logger middleware - so the former recovers from the panic, logs it to sentry along with the request, converts it to 500 response - the latter will then log that. In theory. Not the easiest thing to test.

PR #329, opened 2018-07-05
🤙 Resthooks

rowanseymour:
Some difference with old world functionality here - in old world if a resthook doesn't have any subscribers, then we fake a HTTP 200 response and call it a success... in this PR `@run.webhook` won't be set because there won't be any `resthook_subscriber_called` events to set it.. so then `has_webhook_status` will error. I really don't like faking an HTTP response but erroring the run also seems a bit harsh. @ericnewcomer @nicpottier


flows/actions/call_resthook.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		webhook, err := flows.MakeWebhookCall(run.Session(), req)
+
+		if err != nil {
+			log.Add(events.NewErrorEvent(err))
+		} else {
+			log.Add(events.NewResthookSubscriberCalledEvent(a.Resthook, webhook.URL(), webhook.Status(), webhook.StatusCode(), webhook.Request()))
+
+			if run.Webhook() == nil {
```
nicpottier: Why do we only set it on the first one? What if you call multiple resthooks in a flow step? (not just different subscribers but different webhooks)

nicpottier: Oh nm, see the setting to nil above. Why the test then?

flows/actions/call_webhook.go
```diff
@@ -92,6 +93,5 @@ func (a *CallWebhookAction) Execute(run flows.FlowRun, step flows.Step, log flow
 		log.Add(events.NewWebhookCalledEvent(webhook.URL(), webhook.Status(), webhook.StatusCode(), webhook.Request(), webhook.Response()))
 	}
 
-	run.SetWebhook(webhook)
```
rowanseymour: its supposed to be the responsibility of events to update the session state so this moves to the event

flows/actions/call_resthook.go
rowanseymour: This has changed - too quick on your review!

cmd/flowrunner/testdata/flows/resthook_test.json
```diff
@@ -0,0 +1,202 @@
...
+                        }
+                    ],
+                    "created_on": "2018-07-06T12:30:03.123456789Z",
+                    "payload": "{\n\t\"contact\": {\"uuid\": \"ba96bf7f-bc2a-4873-a7c7-254d1927c4e3\", \"name\": \"Ben Haggerty\", \"urn\": \"tel:+12065551212\"},\n\t\"flow\": {\"name\":\"Resthook\",\"revision\":0,\"uuid\":\"76f0a02f-3b75-4b86-9064-e9195e1b3a02\"},\n\t\"path\": [{\"arrived_on\":\"2018-07-06T12:30:02.123456Z\",\"exit_uuid\":\"\",\"node_uuid\":\"10e483a8-5ffb-4c4f-917b-d43ce86c1d65\",\"uuid\":\"692926ea-09d6-4942-bd38-d266ec8d3716\"}],\n\t\"results\": {},\n\t\"run\": {\"uuid\": \"d2f852ec-7b4e-457f-ae7f-f8b243c49ff5\", \"created_on\": \"2018-07-06T12:30:00.123456Z\"},\n\t\"input\": null,\n\t\"channel\": null\n}",
+                    "resthook": "new-registration",
+                    "step_uuid": "692926ea-09d6-4942-bd38-d266ec8d3716",
+                    "type": "resthook_called"
+                },
+                {
+                    "category": "Failure",
```
rowanseymour: because one of the subscriber URLs has failed

rowanseymour:
@nicpottier please take another look at this - updated according to conversation yesterday

flows/events/resthook_called.go
```diff
@@ -0,0 +1,94 @@
...
+//     "payload": "{...}",
+//     "calls": [
+//       {
+//         "url": "http://localhost:49998/?cmd=success",
+//         "status": "success",
+//         "status_code": 200
+//       },{
+//         "url": "https://api.ipify.org?format=json",
+//         "status": "success",
+//         "status_code": 410
```
nicpottier: I feel like we should put the response in here as well, going to be really annoying to debug error cases with only a status code don't you think?

PR #330, opened 2018-07-05
🕰️ Make time mockable everywhere

rowanseymour:
Need this to procede with #329 as it generates test data with the default webhook payload which has all sorts of time values in it


cmd/docgen/handlers.go
```diff
@@ -219,7 +219,7 @@ func eventsForAction(action flows.Action) (json.RawMessage, error) {
 		}
 
 		// give all our example events a fixed created on time
-		event.SetCreatedOn(session.Environment().Now())
+		event.SetCreatedOn(utils.Now())
```
rowanseymour: Previously some times were mockable via the environment. Now utils.Now() provides system wide mocking and environment.Now() is just that time in the env timezone.

cmd/flowrunner/main.go
```diff
@@ -31,6 +31,15 @@ type FlowTest struct {
 	Outputs      []json.RawMessage        `json:"outputs"`
 }
 
+func normalizeJSON(data json.RawMessage) ([]byte, error) {
```
rowanseymour: this works by loading the JSON into a map and then re-marshalling so that keys are serialized in alphabetical order

flows/fields.go
```diff
@@ -102,7 +102,10 @@ func (v *FieldValue) Reduce(env utils.Environment) types.XPrimitive {
 }
 
 // ToXJSON is called when this type is passed to @(json(...))
-func (v *FieldValue) ToXJSON(env utils.Environment) types.XText { return v.Reduce(env).ToXJSON(env) }
+func (v *FieldValue) ToXJSON(env utils.Environment) types.XText {
```
rowanseymour: unrelated bug that I found because we can now test @JSON() calls on things with time values - the field value might return `nil` from `Reduce`

utils/uuid.go
```diff
@@ -22,18 +22,18 @@ func (g defaultUUID4Generator) Next() UUID {
 	return UUID(uuid.NewV4().String())
 }
 
-// SeededUUID4Generator generates a seedable random v4 UUID using math/rand
-type SeededUUID4Generator struct {
+// generates a seedable random v4 UUID using math/rand
+type seededUUID4Generator struct {
```
rowanseymour: doesn't need to be exported

Issue #328, opened 2018-07-05
When migrating rulesets, record switch type in _ui


rowanseymour:

rowanseymour:
Added in #331

PR #331, opened 2018-07-06
Migrate ruleset_types to editor types in _ui

rowanseymour:
Adds #328

### DIFF
--- a/legacy/definition.go
+++ b/legacy/definition.go
@@ -831,14 +831,15 @@ type fieldConfig struct {
 }
 
 // migrates the given legacy rulset to a node with a router
-func migrateRuleSet(lang utils.Language, r RuleSet, localization flows.Localization) (flows.Node, error) {
+func migrateRuleSet(lang utils.Language, r RuleSet, localization flows.Localization) (flows.Node, string, error) {
 	var newActions []flows.Action
 	var router flows.Router
 	var wait flows.Wait
+	var uiType string
 
 	exits, cases, defaultExit, err := parseRules(lang, r, localization)
 	if err != nil {
-		return nil, err
+		return nil, "", err
 	}
 
 	resultName := r.Label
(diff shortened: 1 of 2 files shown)


legacy/definition.go
```diff
@@ -863,12 +864,13 @@ func migrateRuleSet(lang utils.Language, r RuleSet, localization flows.Localizat
 
 		// subflow rulesets operate on the child flow status
 		router = routers.NewSwitchRouter(defaultExit, "@child.status", cases, resultName)
+		uiType = "subflow"
```
rowanseymour: @ericnewcomer not seeing a complete list of these so thus far just have:

* "subflow"
* "split_by_groups"
* "wait_for_response"
* "split_by_expression"

ericnewcomer: I think that's the only ones we have implemented so far.

What else is there, I think this is close if not complete:
split_by_sms_form, split_by_contact_field, split_by_flow_result

rowanseymour: after discussion, we're using these names:

* wait_for_response
* subflow
* split_by_groups
* split_by_expression
* split_by_contact_field
* split_by_run_result
* split_by_run_result_delimited

legacy/ui.go
```diff
@@ -0,0 +1,76 @@
...
+	UINodeTypeSubflow                   UINodeType = "subflow"
+	UINodeTypeWebhook                   UINodeType = "webhook"
+	UINodeTypeResthook                  UINodeType = "resthook"
+	UINodeTypeSplitByGroups             UINodeType = "split_by_groups"
+	UINodeTypeSplitByExpression         UINodeType = "split_by_expression"
+	UINodeTypeSplitByContactField       UINodeType = "split_by_contact_field"
+	UINodeTypeSplitByRunResult          UINodeType = "split_by_run_result"
+	UINodeTypeSplitByRunResultDelimited UINodeType = "split_by_run_result_delimited"
+	UINodeTypeSplitByRandom             UINodeType = "split_by_random"
+)
```
rowanseymour: @ericnewcomer as discussed but also added resthook and random

PR #334, opened 2018-07-13
TransferTo (for discussion)

rowanseymour:
Started playing around with this to see what is possible, but seems like a good place to pause for discussion:
 * This PR would implement a legacy airtime ruleset as a new action type (`transfer_airtime`), new event type (`airtime_transferred`) and new router test (`has_airtime_status`) which checks the status of the last event. Other way to implement this would be an entirely new router type?
 * Because it's tied to 3rd party service, the code is kept in `/contrib` and wouldn't be considered part of the "official flow spec". (Django uses "contrib" to mean non-core parts like the auth app but maybe that's the wrong name)
 * TransferTo config stuff would be sent in the config section of the request or could be set globally in the server config (up to the caller how it wants to do that)


contrib/transferto/event.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/flows/events"
+)
+
+func init() {
+	events.RegisterType(TypeAirtimeTransfered, func() flows.Event { return &AirtimeTransferedEvent{} })
+}
+
+// TypeAirtimeTransfered is the type of our airtime transferred event
+const TypeAirtimeTransfered string = "airtime_transferred"
```
nicpottier: Variable name wrong, constant right!

PR #335, opened 2018-07-18
💸 TransferTo

rowanseymour:
Addresses #324. Doesn't yet include support for open-range transfers but I'd prefer to add that later when we've found a better way to test this.

By default this won't make any actual requests to transfer-to.com during flowserver trials because we aren't sending and transfer-to configs in the env.


cmd/flowrunner/testdata/flows/airtime.json
```diff
@@ -0,0 +1,51 @@
...
+                        "default_exit_uuid": "",
+                        "operand": "@run",
+                        "cases": [{
+                            "uuid": "d2f852ec-7b4e-457f-ae7f-f8b243c49ff5",
+                            "type": "has_airtime_status",
+                            "arguments": ["success"],
+                            "exit_uuid": "6103fa71-6ca9-4300-aec6-929f50fa1ae0"
+                        }, {
+                            "uuid": "692926ea-09d6-4942-bd38-d266ec8d3716",
+                            "type": "has_airtime_status",
```
rowanseymour: I'm leaning toward keeping the names of the actions/events/tests generic - i.e. `airtime` vs `transferto`

extensions/transferto/action.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// TypeTransferAirtime is the type constant for our airtime action
+var TypeTransferAirtime = "transfer_airtime"
+
+// TransferAirtimeAction attempts to make a TransferTo airtime transfer to the contact
+type TransferAirtimeAction struct {
+	actions.BaseAction
+
+	Amounts map[string]decimal.Decimal `json:"amounts"`
```
rowanseymour: In legacy flows, we have a whole dict of information here about each country/currency - but most of that is for the UI, not the engine. So need to discuss where it belongs instead.

extensions/transferto/client/client.go
```diff
@@ -0,0 +1,218 @@
...
+	for _, line := range strings.Split(string(asBytes), "\r\n") {
+		parts := strings.SplitN(line, "=", 2)
+
+		if len(parts) == 2 {
+			data[parts[0]] = parts[1]
+		}
+	}
+
+	// marshal to JSON and then into the destination struct
+	respJSON, _ := json.Marshal(data)
```
rowanseymour: this is a little weird - parse the custom response format, into a map, marshal to JSON, and then leverage golang's JSON unmarshaling to get that into our structs. I tried using the mapstructure package instead but it doesn't support custom type specific unmarshaling. Could probably rewrite all of this to XML and golang's builtin XML marshalling.

extensions/transferto/action.go
rowanseymour: We discussed and decided this should be currency->amount

PR #344, opened 2018-08-14
to_epoch() becomes epoch() and returns fractional seconds

rowanseymour:


rowanseymour:
@nicpottier yeah... maybe no harm in this being a convenience way of doing that... I dunno... before we stop doing runtime migrations of flows, we can audit our function list

PR #347, opened 2018-08-28
📠 Scanner fixes

rowanseymour:
* Fixes not parsing `@contact...` correctly by only including trailing periods when the following char is a valid name char
 * Fixes `@(")")` so that the `)` in the string literal doesn't close the expression

### DIFF
--- a/excellent/evaluator_test.go
+++ b/excellent/evaluator_test.go
@@ -2,7 +2,6 @@ package excellent
 
 import (
 	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/nyaruka/goflow/excellent/types"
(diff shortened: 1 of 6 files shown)


excellent/evaluator_test.go
```diff
@@ -341,10 +340,9 @@ var errorTests = []struct {
 	errorMsg string
 }{
 	// parser errors
-	{`@(")`, `error evaluating @("): syntax error at "`},
```
rowanseymour: This is no longer an expression error because it's no longer an expression. The `"` starts a string literal that never closes so the expression never closes.

excellent/scanner.go
```diff
@@ -9,11 +9,12 @@ import (
 	"unicode"
 )
 
-type xToken int
+// XTokenType is a set of types than can be scanned
+type XTokenType int
```
rowanseymour: go wasn't happy about some of these things being unexported but returned from exported functions

PR #363, opened 2018-09-13
💼 Asset references

rowanseymour:
* `Parent()` of a channel asset is a reference instead of just a UUID - this is the way it was before the Great Assets Refactor (#362), but required that..
 * Asset reference structs all move into the `goflow/assets` package.. which is where they belong anyway.


nicpottier:
From using it a bit, I find that UUID types and reference types are in `assets.` really bizarre. I feel like we need some other package for that kind of stuff though I don't have a great naming suggestion.

PR #365, opened 2018-09-13
Adds optional ui config option for nodes

ericnewcomer:
* Moves non-legacy ui stuff out of legacy and into definition
* Promotes some uses of interface to concrete types
* Updates ruleset migration tests to also test UI migration


flows/definition/ui.go
```diff
@@ -0,0 +1,94 @@
...
+		NodeType_:     nodeType,
+		UiNodeConfig_: uiNodeConfig,
+		Position_: &position{
+			left: x,
+			top:  y,
+		},
+	}
+}
+
+func (uie uiNodeDetails) GetNodeConfig() flows.UINodeConfig {
```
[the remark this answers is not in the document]
rowanseymour: also bind variables are always one char so `uie` would be `u`

flows/definition/ui.go
```diff
@@ -0,0 +1,86 @@
...
+//------------------------------------------------------------------------------------------
+// Details for a specific node's configuration
+//------------------------------------------------------------------------------------------
+
+type position struct {
+	left int
+	top  int
+}
+
+func (p position) GetLeft() int { return p.left }
```
rowanseymour: `Left()`

flows/definition/ui.go
```diff
@@ -0,0 +1,86 @@
...
+func (p position) GetTop() int  { return p.top }
```
rowanseymour: `Top()`

Issue #367, opened 2018-09-14
JSON data is normalized in legacy engine 


rowanseymour:
When a webhook returns JSON or a flow is started via the API with extra, the saved extra is normalized, e.g. `{"Foo Bar": "Hello"}` -> `{"Foo_Bar": "Hello"}`. When it's referenced in an expression, the key lookups are case-insensitive, e.g. `@extra.foo_bar` returns `"Hello"`. 

In contrast, goflow just stores the raw JSON and uses the `jsonparser` library to resolve keys as needed. To look up the above JSON you would write something like `@(run.webhook.json["Foo Bar"])`.

So we have two options:
  - Switch to normalizing all JSON like the old engine. It could still be lazy - when you try to resolve `run.webhook.json` we parse the webhook body into something that resolves it like the old world. 
  - Add a function to maintain normalized lookups for `@extra` in migrated flows, e.g. `@(legacy_extra(run.webhook.json).foo_bar)`

Is it better that `run.webhook.json` behaves like the rest of the context, or behaves like actual JSON?

One complication with second approach is that `@extra` is already migrated to a fairly complicated expression, since what was `@extra` is now in one of two different places, e.g. 

`@(if(is_error(run.webhook.json.foo_bar), trigger.params.foo_bar, run.webhook.json.foo_bar))`. 

But maybe this allows us to simplify that for migrated flows, and lean on the custom function more, e.g. 

`@(legacy_extra(trigger.params, run.webhook.json).foo_bar)`

or even...

`@(legacy_extra().foo_bar)`

nicpottier:
I don't think we can do `@(legacy_extra())` can we? excellent doesn't have globals apart from env does it? I guess we could stuff in a `@(legacy_extra.foo_bar)` into the context though and maybe that's the simplest? Who has `legacy_extra` populated though? Only migrated flows?

rowanseymour:
Well it feels cheaky but we can because we have the environment and that's an instance of `RunEnvironment` which has the run. We do this for tests like `has_state` to get at the location data which is associated with the session.

nicpottier:
Ya, that feels like a much worse bastardization than location data which feels more 'environmenty' for sure. I would rather stuff the context with stuff.

rowanseymour:
Addressed in #368

PR #368, opened 2018-09-17
👴🏻 Legacy @extra

rowanseymour:
Adds `@legacy_extra` to the context which mimics `@extra` in legacy flows. Addresses #367. By looking at events on the run, it builds up extra from:

1. Trigger params
2. Webhook/resthook responses that can be parsed as JSON
3. Results from `has_pattern` tests

It's constructed lazily when `@legacy_extra` occurs in an expression so no performance hit for flows not using this, and it's not persisted on the session.

### DIFF
--- a/flows/route.go
+++ b/flows/route.go
@@ -2,9 +2,9 @@ package flows
 
 // Route describes leaving a node
 type Route struct {
-	exit       ExitUUID
-	match      string
-	matchExtra map[string]string
+	exit  ExitUUID
+	match string
+	extra map[string]string
 }
 
 // Exit returns the UUID of the chosen exit
@@ -13,13 +13,13 @@ func (r Route) Exit() ExitUUID { return r.exit }
 // Match returns the match which led to this route being chosen
 func (r Route) Match() string { return r.match }
 
-// MatchExtra returns additional data from the match
-func (r Route) MatchExtra() map[string]string { return r.matchExtra }
+// Extra returns additional data from the match
+func (r Route) Extra() map[string]string { return r.extra }
 
 // NoRoute is used when a router can't find a route
 var NoRoute = Route{}
 
 // NewRoute creates a new route
-func NewRoute(exit ExitUUID, match string, matchExtra map[string]string) Route {
-	return Route{exit, match, matchExtra}
+func NewRoute(exit ExitUUID, match string, extra map[string]string) Route {
+	return Route{exit, match, extra}
 }
--- a/flows/engine/session.go
+++ b/flows/engine/session.go
@@ -453,7 +453,7 @@ func (s *session) pickNodeExit(run flows.FlowRun, node flows.Node, step flows.St
 
 	// save our results if appropriate
 	if router != nil && router.ResultName() != "" {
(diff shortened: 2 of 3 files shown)


docs/md/expressions.md
```diff
@@ -1518,8 +1518,6 @@ Both text values are trimmed of surrounding whitespace and matching is case-inse
 @(has_pattern("Sell cheese please", "buy (\w+)")) → false
 @(has_pattern("Buy cheese please", "buy (\w+)")) → true
 @(has_pattern("Buy cheese please", "buy (\w+)").match) → Buy cheese
-@(has_pattern("Buy cheese please", "buy (\w+)").match.groups[0]) → Buy cheese
-@(has_pattern("Buy cheese please", "buy (\w+)").match.groups[1]) → cheese
```
rowanseymour: This is really not user-friendly as a way to get stuff because you have to repeat whatever expression you're using as the operand. I've removed it and think we need to come up with a better replacement for `@extra.0...1` which for now would be `@legacy_extra.0...1`

flows/events/run_result_changed.go
```diff
@@ -32,16 +32,17 @@ type RunResultChangedEvent struct {
...
-	CategoryLocalized string         `json:"category_localized,omitempty"`
-	NodeUUID          flows.NodeUUID `json:"node_uuid" validate:"required,uuid4"`
-	Input             *string        `json:"input,omitempty"`
+	Name              string            `json:"name" validate:"required"`
+	Value             string            `json:"value"`
+	Category          string            `json:"category"`
+	CategoryLocalized string            `json:"category_localized,omitempty"`
+	NodeUUID          flows.NodeUUID    `json:"node_uuid" validate:"required,uuid4"`
+	Input             *string           `json:"input,omitempty"`
+	Extra             map[string]string `json:"extra,omitempty"`
```
rowanseymour: For now this just occurs on the event and doesn't go into the actual result dict

docs/md/expressions.md
nicpottier: Why remove the example though? Could still see people wanting to get at matches within an expression without having a totally separate split etc..

flows/route.go
```diff
@@ -2,8 +2,9 @@ package flows
 
 // Route describes leaving a node
 type Route struct {
-	exit  ExitUUID
-	match string
+	exit       ExitUUID
+	match      string
+	matchExtra map[string]string
```
nicpottier: don't feel strongly but would support this just being `extra` :)

docs/md/expressions.md
rowanseymour: We added a separate function for getting regex group matches, `regex_group(pattern, text, num)` - that's simpler than this `....match.groups[1]` syntax.

flows/route.go
rowanseymour: I'm fine with that

docs/md/expressions.md
rowanseymour: Tho curiously I've just noticed we added it to the old engine but it seems I forgot to add it in the new engine

PR #374, opened 2018-09-20
⛓ More migrations

ericnewcomer:
* Add migration type for Airtime
 * Make non-router nodes declare a type
 * Nest operand for various ui configs for consistency

### DIFF
--- a/legacy/ui.go
+++ b/legacy/ui.go
@@ -7,7 +7,9 @@ import (
 
 // the different node types supported by the editor
 const (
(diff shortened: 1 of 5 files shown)


legacy/ui.go
```diff
@@ -7,7 +7,9 @@ import (
 
 // the different node types supported by the editor
 const (
+	UINodeTyepActionSet                 flows.UINodeType = "execute_actions"
```
rowanseymour: typo

PR #377, opened 2018-09-24
🚧 Events refactor

rowanseymour:
* event functionality moves back into actions
 * events only generated when they make actual state changes
 * events can't be bi-directional - a subset of events can be sent by caller and applied in engine
 * `value` of `contact_field_changed` is now full dict of types
 * `contact_groups_added` and `contact_groups_removed` merged into `contact_groups_changed`
 * dynamic group re-evaluation generates `contact_groups_changed`
 * webhook and resthook actions both generate the same `webhook_called` events
 * both of those also generate a `run_result_changed` event if they save an event
 * simplifications to how contact field values are created


cmd/flowrunner/testdata/flows/all_actions.json
```diff
@@ -178,7 +178,7 @@
                         {
                             "uuid": "e0ae5679-5145-4bdd-9737-3481932bf095",
                             "type": "set_contact_language",
-                            "language": "eng"
+                            "language": ""
```
rowanseymour: needed to change language but have the flow still work.. so just clearing it instead which still generates a `contact_language_changed` event

cmd/flowrunner/main.go
```diff
@@ -158,7 +157,7 @@ func main() {
 			log.Fatalf("Error unmarshalling output: %s", err)
 		}
 
-		err = session.Resume([]flows.Event{event})
+		err = session.Resume([]flows.CallerEvent{event})
```
rowanseymour: only `CallerEvent` events come from the caller and can be validated and applied

cmd/flowrunner/testdata/default.json
```diff
@@ -53,7 +53,7 @@
                 "name": "Kigali City Members"
             },
             {
-                "uuid": "d7ff4872-9238-452f-9d38-2f558fea89e0",
+                "uuid": "bf282a79-aa74-4557-9932-22a9b3bce537",
```
rowanseymour: the additional group change logging showed we were getting this mixed up because it had the same UUID as another group

cmd/flowrunner/testdata/default.json
```diff
@@ -67,22 +67,37 @@
...
                 "address": "+12345671111",
-                "schemes": ["tel"],
-                "roles": ["send", "receive"]
+                "schemes": [
+                    "tel"
+                ],
+                "roles": [
+                    "send",
+                    "receive"
+                ]
```
rowanseymour: vscode format-on-save doing its thing...

cmd/flowrunner/testdata/flows/dynamic_groups_test.json
```diff
@@ -40,43 +67,74 @@
...
+                },
+                {
+                    "created_on": "2018-07-06T12:30:12.123456789Z",
                     "msg": {
                         "channel": {
                             "name": "Android Channel",
                             "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d"
                         },
-                        "text": "Current groups: [\"Males\",\"Old Men\"]",
+                        "text": "Current groups: [\"Males\",\"Old Men\",\"MTN Callers\"]",
```
rowanseymour: contact URN additions weren't previously triggering a dynamic group evaluation.. despite me apparently including one in this test

cmd/flowrunner/testdata/flows/no_contact_test.json
```diff
@@ -33,7 +33,14 @@
                 },
                 {
                     "created_on": "2018-07-06T12:30:05.123456789Z",
-                    "fatal": true,
+                    "fatal": false,
```
rowanseymour: chatted with @ericnewcomer about this briefly and we felt an action that requires a contact being run in a session without a contact should just a log an error and move on

nicpottier:
Do dynamic groups work with `created_on` and `language` per chance? (know that's unrelated but this reminded me of that change)

flows/field.go
```diff
@@ -23,23 +23,23 @@ func (f *Field) Asset() assets.Field { return f.Field }
...
-	number   *types.XNumber
-	state    LocationPath
-	district LocationPath
-	ward     LocationPath
+	Text     types.XText      `json:"text"`
+	Datetime *types.XDateTime `json:"datetime,omitempty"`
+	Number   *types.XNumber   `json:"number,omitempty"`
+	State    LocationPath     `json:"state,omitempty"`
+	District LocationPath     `json:"district,omitempty"`
+	Ward     LocationPath     `json:"ward,omitempty"`
```
nicpottier: Yay, had just created my own mailroom type for this since I couldn't make them.

rowanseymour: yeah we had a `fieldValueEnvelope` previously but then noticed it was the same as `Value`

rowanseymour:
@nicpottier they do - side note about dynamic groups - don't be surprised if you see unexpected `contact_groups_changed` events because it's possible that contacts start off in the wrong groups and at some point we trigger a re-eval and things are fixed/changed. I ran into this in a test but it's not hard to imagine there are a few contacts in RP in the wrong groups.

nicpottier:
hrmm, ya, that's an interesting thing.. should we maybe always reevaluate them? seems slightly less weird that running through a flow will ALWAYS fix your groups rather than only doing it sometimes?

PR #380, opened 2018-09-26
🌐 Add resthook slug to webhook_called event if it exists

rowanseymour:
Caller potentially needs to remove a subscriber if the call has a 410 response code


legacy/definition.go
```diff
@@ -677,7 +677,7 @@ func migrateRuleSet(lang utils.Language, r RuleSet, localization flows.Localizat
 		}
 
 		// resthook rulesets operate on the webhook status, saved as category
-		router = routers.NewSwitchRouter(defaultExit, fmt.Sprintf("@results.%s.category", utils.Snakify(resultName)), cases, "")
+		router = routers.NewSwitchRouter(defaultExit, fmt.Sprintf("@(default(results.%s.category, \"Success\"))", utils.Snakify(resultName)), cases, "")
```
rowanseymour: this is needed when we follow a resthook action in case it didn't save any results because it didn't have any subscribers

PR #382, opened 2018-09-27
🐃 Cleanup field value code so that we no longer need empty values

rowanseymour:


docs/md/sessions.md
```diff
@@ -66,7 +66,12 @@ Is used when another session triggered this run using a trigger_flow action.
             "language": "",
             "timezone": "",
             "created_on": "0001-01-01T00:00:00Z",
-            "urns": []
+            "urns": [],
+            "fields": {
+                "gender": {
+                    "text": "Male"
+                }
+            }
```
rowanseymour: see below - this trigger previously had a non-existent field so it wasn't appearing when the trigger was later serialized

flows/contact.go
```diff
@@ -365,9 +371,8 @@ func (c *Contact) ResolveQueryKey(env utils.Environment, key string) []interface
 	if key == "language" {
 		if c.language != utils.NilLanguage {
 			return []interface{}{string(c.language)}
-		} else {
-			return nil
 		}
+		return nil
```
rowanseymour: gofmt

flows/field.go
```diff
@@ -76,11 +76,17 @@ func (v *FieldValue) TypedValue() types.XValue {
...
 	case assets.FieldTypeDistrict:
-		return v.District
+		if v.District != "" {
+			return v.District
+		}
 	case assets.FieldTypeWard:
-		return v.Ward
+		if v.Ward != "" {
+			return v.Ward
+		}
```
rowanseymour: this method now always returns `nil` if a value has no value in its field's type

flows/field.go
```diff
@@ -23,7 +23,7 @@ func (f *Field) Asset() assets.Field { return f.Field }
 
 // Value represents a value in each of the field types
 type Value struct {
-	Text     types.XText      `json:"text"`
+	Text     types.XText      `json:"text" validate:"required"`
```
nicpottier: So this handles the validation case coming in from JSON but if our contract is not this can't be an empty string we should probably add some assertions or errors on new value creation. (and perhaps make these fields private in the process?)

If my understanding is correct I believe these changes would not work correctly as mailroom is, but no error of validation would show that. (IE, we pass in empty `Value` structs per previous implementation)

rowanseymour: Yeah we can validate these

rowanseymour: I've added a check in `NewFieldValues` which already can return an error. I can encapsulate `Value` a bit more if you think it's worth it (interface?) but I just tried using the inner anon struct approach and couldn't figure out how to initialize it in `NewValue`.

PR #389, opened 2018-09-28
⏲️ Record time taken in webhook_called events

rowanseymour:


rowanseymour:
@nicpottier it's millis (see https://github.com/nyaruka/goflow/pull/389/files#diff-985ffcfee3989f9b6f25cac4b05c2058R47)

rowanseymour:
@nicpottier all the tests give as `1000` because the mocked time returns time values one second apart

nicpottier:
ah ok.. ya for these types of thing I always like having a unit in there because they are so random, but whatever. `elapsed_ms` would be my first choice. :)

rowanseymour:
I'm fine with that (https://github.com/nyaruka/goflow/issues/391)

PR #392, opened 2018-10-02
🤾🏼 Resumes

rowanseymour:
- [x] replace resuming caller events with resumes
 - [x] replace initial caller events with msg trigger type
 - [x] do diffing of contact/environment on resume so we only generate events if they're different
 - [x] improve msg triggers so they describe why a message started a session


cmd/flowrunner/testdata/flows/brochure_test.json
```diff
@@ -170,23 +152,37 @@
...
+                        "channel": {
+                            "name": "Nexmo",
+                            "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d"
+                        },
+                        "text": "Ryan Lewis",
+                        "urn": "tel:+12065551212",
+                        "uuid": "9bf91c2b-ce58-4cef-aacc-281e03f69ab5"
+                    },
+                    "step_uuid": "c34b6c7d-fa06-4563-92a3-d648ab64bccb",
+                    "type": "msg_received"
```
rowanseymour: previously such events were considered caller events and not included in the session event log - so callers do have to be aware `msg_received`, `run_expired`, `wait_timed_out`, `contact_changed` and `environment_changed` are not events that they have to act on

extensions/transferto/action.go
```diff
@@ -49,18 +49,18 @@ func (a *TransferAirtimeAction) AllowedFlowTypes() []flows.FlowType {
 }
 
 // Execute runs this action
-func (a *TransferAirtimeAction) Execute(run flows.FlowRun, step flows.Step, log flows.EventLog) error {
+func (a *TransferAirtimeAction) Execute(run flows.FlowRun, step flows.Step) error {
```
rowanseymour: now that events don't do anything there's no need for this model of applying them after actions, so we don't need the whole event log thing - actions just log events directly on the run

flows/interfaces.go
```diff
@@ -244,23 +244,26 @@ type Trigger interface {
 	types.XValue
 	types.XResolvable
 
+	Initialize(Session) error
+	InitializeRun(FlowRun, Step) error
```
rowanseymour: not the greatest naming but triggers need to do 2 things - setup the session and then additionally do stuff to the first run, i.e. if we have a message, we need to set it as input on the first run

flows/resumes/base.go
```diff
@@ -0,0 +1,92 @@
...
+
+		// TODO diffing
+
+		run.LogEvent(step, events.NewEnvironmentChangedEvent(r.Environment()))
+	}
+	if r.contact != nil {
+		run.SetContact(r.contact)
+		run.Session().SetContact(r.contact)
+
+		// TODO diffing
```
rowanseymour: still pondering on giving these things complex .Equal() methods or relying on string comparison of their JSON forms

nicpottier: Ya, its a bit of work though diffing would lead to clearer (and smaller!) logs so may be worth it.

flows/runs/run.go
```diff
@@ -63,9 +61,8 @@ func (r *flowRun) UUID() flows.RunUUID               { return r.uuid }
 func (r *flowRun) Session() flows.Session            { return r.session }
 func (r *flowRun) Environment() flows.RunEnvironment { return r.environment }
 
-func (r *flowRun) Flow() flows.Flow                  { return r.flow }
-func (r *flowRun) Contact() *flows.Contact           { return r.contact }
-func (r *flowRun) SetContact(contact *flows.Contact) { r.contact = contact }
+func (r *flowRun) Flow() flows.Flow        { return r.flow }
+func (r *flowRun) Contact() *flows.Contact { return r.session.Contact() }
```
rowanseymour: we don't need contact on run and removing it makes it simpler to keep run.contact always in sync with session.contact

PR #412, opened 2018-10-12
📯 Simpler typed things

rowanseymour:
aka, the last great refactor...

All typed things are now directly marshal-able to JSON with `json.Marshal` - no need for converting to typed envelopes first.


nicpottier:
way better, also like all the new constructors for actions, that'll come in handy for testing.

PR #414, opened 2018-10-15
💯 Improved number parsing

rowanseymour:


flows/routers/tests/tests.go
```diff
@@ -347,7 +349,7 @@ func HasPattern(env utils.Environment, text types.XText, pattern types.XText) ty
 //
 // @test has_number(text)
 func HasNumber(env utils.Environment, text types.XText) types.XValue {
-	return testNumber(env, text, types.XNumberZero, isNumberTest)
+	return testNumber(env, text, types.XNumberZero, types.XNumberZero, isNumberTest)
```
rowanseymour: reworked these number test functions to take 2 test args so that `HasNumberBetween` can use the same code path

flows/routers/tests/tests.go
```diff
@@ -810,44 +802,79 @@ func hasOnlyPhraseTest(origHays []string, hays []string, pins []string) XTestRes
 // Numerical Test Functions
 //------------------------------------------------------------------------------------------
 
-var parseableNumberRegex = regexp.MustCompile(`^[$£€]?([\d,][\d,\.]*([\.,]\d+)?)\D*$`)
-
+var altNumerals = map[rune]rune{
+	'٠': '0', // Arabic
```
rowanseymour: the neat thing about Arabic numerals is that it's basically the same but with different digit characters (i.e. decimal and left-to-right)

flows/routers/tests/tests.go
```diff
@@ -810,44 +802,79 @@ func hasOnlyPhraseTest(origHays []string, hays []string, pins []string) XTestRes
...
+	repl, found := altNumerals[r]
+	if found {
+		return repl
+	}
+	return r
+}
+
+// ParseDecimalFuzzy parses a decimal from a string
 func ParseDecimalFuzzy(val string, format *utils.NumberFormat) (decimal.Decimal, error) {
+	// must contain at least one real digit - prevents things like ll becoming 11
```
rowanseymour: seems like reasonable restriction?

nicpottier:
Seems fine. Why are we going Arabic exactly? What about Hindi? Chinese? I don't really favor adding that kind of thing unless we have a specific ask for it and then we should address it with a specific strategy. (IE, we will cover all numbers in the top 10 languages or whatnot)

rowanseymour:
Arabic and Persian are trivial to support because they're just different unicode chars for the same digits (other languages not necessarily so), and we know we have users using these languages, and we try to support these languages elsewhere in the codebase. Why not add that support and if anyone requests Chinese we can attempt that later?

rowanseymour:
I'm also not wedded to keeping that part in if you really don't like it - I just added it when I noticed unicode.IsDigit() matches things like ١.

nicpottier:
Ya I just want to be systematic about this not scatter shot, because then it is easier to explain to people what exists and why and why or why not we might accept a PR or whatnot. @ericnewcomer thoughts on this? 

I view this the same as turning dates to days of week etc.. a nice feature for sure, but we need to figure out what are we promising and what are the constraints.

rowanseymour:
@nicpottier I'm gonna pull that part out and put ther code snippet in an issue for future discussion

Issue #417, opened 2018-10-16
Validate all flows on startup


rowanseymour:
Validating a flow should validate each action, and in the case of StartFlow actions this will validate that the flow that they reference is also valid.

rowanseymour:
Added in #419

PR #419, opened 2018-10-16
🚨 Validate all flows which are referenced in the current flow

rowanseymour:
Addresses #417


extensions/transferto/action.go
```diff
@@ -46,7 +46,7 @@ func NewTransferAirtimeAction(uuid flows.ActionUUID, amounts map[string]decimal.
 }
 
 // Validate validates our action is valid and has all the assets it needs
-func (a *TransferAirtimeAction) Validate(assets flows.SessionAssets) error {
+func (a *TransferAirtimeAction) Validate(assets flows.SessionAssets, context *flows.ValidationContext) error {
```
rowanseymour: need some state to prevent loops - tried putting this in `FlowAssets` but turned out to be simpler to avoid race conditions if it's passed around as an arg

flows/definition/flow_test.go
```diff
@@ -20,103 +18,62 @@ import (
...
-	testInvalidFlow(t,
-		session.Assets(),
-		"testdata/flow_with_invalid_case_exit.json",
+	},
+	{
+		"flow_with_invalid_case_exit.json",
 		"validation failed for node[uuid=a58be63b-907d-4a1a-856b-0bb5579d7507]: validation failed for router: case exit 37d8813f-1402-4ad2-9cc2-e9054a96525b is not a valid exit",
-	)
-
-	assetsJSON, err := ioutil.ReadFile("testdata/flow_validation.json")
```
rowanseymour: all of these test conditions reworked into a table based tests above

PR #424, opened 2018-10-17
🔥 Fix not migrating text test args and errors blowing up switch routers

rowanseymour:


flows/routers/switch.go
```diff
@@ -133,7 +133,9 @@ func (r *SwitchRouter) PickRoute(run flows.FlowRun, exits []flows.Exit, step flo
 		// tests have to return either errors or test results
 		switch typedResult := result.(type) {
 		case types.XError:
-			return nil, flows.NoRoute, types.NewXErrorf("error calling %s: %s", strings.ToUpper(test), typedResult.(types.XError).Error())
+			// test functions can return an error
+			run.LogError(step, fmt.Errorf("error calling test %s: %s", strings.ToUpper(test), typedResult.Error()))
```
rowanseymour: If an expression error is passed a function besides `is_error` it is returned or a function may just return an error. In these cases we log it. 

Migrated split-by-expression rulesets will always have their operands wrapped in `@(if(is_error(<migrated-exp>), "<original-exp>", <migrated-exp>))` so that errors are swallowed and the original expression is rewritten out. This should match existing behavior which some users rely on to determine if run results are set.

Issue #427, opened 2018-10-19
Contact channel change action should generate URN changed events


rowanseymour:
The `set_contact_channel` action modifies a contact's URN's changing their order and affinity. It currently generates a `contact_channel_changed` event which just lists the channel from the action but this doesn't describe well what happened.

Maybe we need to introduce a `contact_urns_changed` event which could also be used by the `add_contact_urn` action? Difficulty is that unlike groups, URN order is important.

nicpottier:
Seems like `contact_urns_changed` which outputs the new ordered list in cases where things changed would make the most sense. Could then be used for the AddContactURN action as well. Bit of a pain for mailroom to then figure out smallest diff to apply that but not the end of the world. Only trick on engine level is that it is only output when it actually changes the contact's URNs. (IE, the action can be a noop)

rowanseymour:
So despite what I might have said previously, `AddContactURN` adds a URN to the end of the list (in old engine too). I would imagine in the future we might want to allow adding to the beginning or end, but either way, it does seem like it returning the complete list is appropriate here.

rowanseymour:
Fixed in #428

PR #428, opened 2018-10-22
🕵️ URN updates

rowanseymour:
* `add_contact_urn` action returns `contact_urns_changed` event with new URN list
 * `set_contact_channel` action also returns `contact_urns_changed` event if URNs change (#427)
 * don't expose URN query string params in expressions, e.g. users never see params like `id` or `channel`, only scheme+path+display


extensions/transferto/action.go
```diff
@@ -69,7 +69,7 @@ func (a *TransferAirtimeAction) Execute(run flows.FlowRun, step flows.Step) erro
 		run.LogEvent(step, events.NewErrorEvent(fmt.Errorf("can't transfer airtime to contact without a tel URN")))
 		return nil
 	}
-	recipient := telURNs[0].Path()
+	recipient := telURNs[0].URN().Path()
```
rowanseymour: `ContactURN` used to be a `urns.URN` but now the urn part is a field so we can update it when `ContactURN.SetChannel` is called

Issue #421, opened 2018-10-17
Waits for different media types


rowanseymour:
Needed for Surveyor 2

rowanseymour:
Currently RapidPro surveyor flows add 4 wait types to `RuleSet`:

```python
TYPE_WAIT_PHOTO = "wait_photo"
TYPE_WAIT_VIDEO = "wait_video"
TYPE_WAIT_AUDIO = "wait_audio"
TYPE_WAIT_GPS = "wait_gps"
```

Those are instructions to the surveyor client to record a photo/video/audio/gps but at some point they have to become messages, which presents 3 possibilities:

1. They are distinct wait types which require distinct resume types and remain distinct types  (e.g. `image_received` events) until RapidPro has to save things in the db and converts them to `Msg`. `@input` can be different types of things.
2. They are distinct wait types which require distinct resume types and the resume converts them to messages (i.e. the usual `msg_received` event). `@input` is always a msg.
3. They are a regular msg wait (meaning "wait for a message with an attachment of type image.. etc"). `@input` is always a msg.

I think I'm leaning toward 3 and it would mean just adding a field to our existing `msg` wait, e.g.

```json
"wait": {
    "type": "msg",
    "require_media": "image",
    "timeout": 60
}
```

It does make me wonder about making it easier to ask for the same thing in a regular messaging flow - currently users need to look in `@step.value` which contains attachment URLs and manually check for something looking like a correct image URL. It would be great to make this easier but I don't see any obvious way to consolidate asking for media in surveyor and asking for media over regular messaging. In the latter case you have to accept any msg and use flow logic to re-prompt the contact until you get the media you want. Maybe we could make that easier with new rule types?

nicpottier:
Ya, I don't think 1 or 2 work, because you can get resumed with something of a different type. IE, you could be waiting for an image, but they could send text, and that needs to properly resume, but fail through a router and get the retry message. Maybe the wait should be a wait for "input"? And then it is up for the router to determine whether the last input matches what is expected?

nicpottier:
> It would be great to make this easier but I don't see any obvious way to consolidate asking for media in surveyor and asking for media over regular messaging. In the latter case you have to accept any msg and use flow logic to re-prompt the contact until you get the media you want.

Are these that different really? I guess we could pass back the "hint" of what media we want back in the resume in the wait we pass back, but I think the engine needs to deal with being resumed with something that doesn't match just like any other input.

nicpottier:
[...]

PR #430, opened 2018-10-23
📸 Media wait hints

rowanseymour:
Addresses #421

### DIFF
--- a/legacy/testdata/rulesets.json
+++ b/legacy/testdata/rulesets.json
@@ -611,6 +611,63 @@
     },
     {
         "legacy_ruleset": {
+            "uuid": "bf8bcadf-7a26-4c4b-a5ab-577987ac3c20",
+            "x": 374,
+            "y": 668,
+            "label": "Selfie",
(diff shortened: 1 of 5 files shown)


legacy/testdata/rulesets.json
```diff
@@ -609,6 +609,63 @@
...
+                {
+                    "uuid": "b26f9640-93b4-409e-b65e-7d0ee1e886dd",
+                    "name": "All Responses",
+                    "destination_node_uuid": "5b977652-91e3-48be-8e86-7c8094b4aa8f"
+                }
+            ]
+        },
+        "expected_localization": {},
+        "expected_ui": {
+            "type": "wait_for_response",
```
rowanseymour: @ericnewcomer thoughts on what the UI part of this should be?

ericnewcomer: Since the UI needs to be in the business of creating media_hints it seems letting it switch on those is sufficient.

PR #433, opened 2018-10-25
🌐 Don't url encode msg attachment expressions automatically

rowanseymour:
Unlike webhook URLs, these aren't url encoded in the old engine, and the problem arises when an expression is used for the entire attachment URL and not just a query param.


rowanseymour:
@nicpottier made a little change there to encode spaces as `%20` rather than `+`. That matches old behavior and also means `url_encode` should also work in the URL path

PR #459, opened 2018-11-28
🧮 Don't do I>1 O>0 substitutions when parsing numbers

rowanseymour:
As discussed at https://github.com/rapidpro/rapidpro/issues/828


flows/routers/tests/tests_test.go
```diff
@@ -291,7 +290,6 @@ func TestParseDecimalFuzzy(t *testing.T) {
 		{"1234", decimal.RequireFromString("1234"), utils.DefaultNumberFormat},
 		{"1,234.567", decimal.RequireFromString("1234.567"), utils.DefaultNumberFormat},
 		{"1.234,567", decimal.RequireFromString("1234.567"), &utils.NumberFormat{DecimalSymbol: ",", DigitGroupingSymbol: "."}},
-		{"l00", decimal.RequireFromString("100"), utils.DefaultNumberFormat},
 		{"100.00", decimal.RequireFromString("100.00"), utils.DefaultNumberFormat},
```
nicpottier: Minor but seems like keeping the test but asserting it is NOT a decimal would be nice.

PR #460, opened 2018-12-11
🤙 IVR

rowanseymour:
Adds two new events:

 * `ivr_say` -> caller should issue a say IVR command with the given text
 * `ivr_play` -> caller should issue a play IVR command with the given audio file URL

And two new actions:

 * `say_msg` -> generates a `ivr_say` or a `ivr_play` depending on whether it has an audio URL or just text for the current language
 * `play_audio` -> generates an `ivr_play` for the given audio URL

MediaHint strings on waits become typed objects called Hints, and we add a new type of hint to model TYPE_WAIT_DIGIT and TYPE_WAIT_DIGITS on legacy flows.

Addresses #345


flows/waits/hints/digits.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// TypeDigits is the type of our digits hint
+const TypeDigits string = "digits"
+
+// DigitsHint requests a message containing one or more digits
+type DigitsHint struct {
+	baseHint
+
+	Count        *int   `json:"count,omitempty"`
```
rowanseymour: <gather num_digits= /> in TwiML

flows/waits/hints/digits.go
```diff
@@ -0,0 +1,36 @@
...
+	TerminatedBy string `json:"terminated_by,omitempty"`
```
rowanseymour: `<gather finishOnKey="..." />` in TwiML

Issue #464, opened 2018-12-11
Record contact changes in such a way that they can be replayed for surveyor submissions


rowanseymour:
Surveyor contacts can be matched against real contacts via URNs at submission time, so mailroom needs to apply contact changes to the actual contact. Events by themselves don't provide enough information to do this.

PR #469, opened 2018-12-14
👷 Contact modifiers

rowanseymour:
* All changes to the contact happen in modifiers
 * Modifiers log events *if* they make a change
 * Modifiers are responsible for dynamic group re-evaluation
 * Session `Start`/`Resume` now returns a `Sprint` which contains the new events and modifiers
 * Modifiers can be marshaled to and from JSON

Addresses #464


extensions/transferto/action.go
```diff
@@ -56,25 +56,25 @@ func (a *TransferAirtimeAction) AllowedFlowTypes() []flows.FlowType {
 }
 
 // Execute runs this action
-func (a *TransferAirtimeAction) Execute(run flows.FlowRun, step flows.Step) error {
+func (a *TransferAirtimeAction) Execute(run flows.FlowRun, step flows.Step, logModifier func(flows.Modifier), logEvent func(flows.Event)) error {
 	contact := run.Contact()
 	if contact == nil {
-		run.LogEvent(step, events.NewErrorEvent(errors.Errorf("can't execute action in session without a contact")))
+		logEvent(events.NewErrorEvent(errors.Errorf("can't execute action in session without a contact")))
```
rowanseymour: a bit more functional - the benefit of which is that now actions don't need to know where to log events and modifiers... they just log them to the passed in functions.

PR #476, opened 2019-01-08
🤙 IVR as messages

rowanseymour:
* Replaces `ivr_say` and `ivr_play` events with new `ivr_created` event which is IVR equivalent of `msg_created`
* Adds `connection` to trigger which provides the channel and URN for IVR messages
* Renames `StartFlowAction`  &rarr; `EnterFlowAction` and `flow_triggered` &rarr; `flow_entered` for clarity
* Disallow `EnterFlowAction` from entering flows of different type (i.e. modality)


flows/triggers/manual.go
```diff
@@ -36,9 +36,9 @@ type ManualTrigger struct {
 }
 
 // NewManualTrigger creates a new manual trigger
-func NewManualTrigger(env utils.Environment, flow *assets.FlowReference, contact *flows.Contact, params types.XValue, triggeredOn time.Time) flows.Trigger {
+func NewManualTrigger(env utils.Environment, flow *assets.FlowReference, contact *flows.Contact, connection *flows.Connection, params types.XValue, triggeredOn time.Time) flows.Trigger {
```
nicpottier: Not a big deal but general comment that this constructor is getting a bit complex for callers to understand what is required / optional etc.. haven't really settled on the right pattern here whether it is to have different contructors `NewManualIVRTrigger` or the like or to have modifiers `NewManualTrigger(..).WithConnection(..)` but we should figure out what our standard is and when we decide to start using it. Maybe any optional arguments warrant a different signature in some way?

rowanseymour: Yeah agree it's getting messy. The only saving grace is that this not being python, you can't muddle most of the arguments cuz they're different types. 

I think the builder pattern would be annoying cuz you'd have to build separate builder classes for each type. I think specialized constructors (or just leaning on set methods for not-so-common params) would be good way forward.

nicpottier: Ya, it is pointers that are still vague, as in, is this required or can I pass nil.

Builders don't need to be their own thing, the instance can just have a setter called `WithXXX` that returns itself.

rowanseymour: One argument against Set (or With on the instance) methods is that things like triggers and resumes are supposed to be immutable.

Issue #477, opened 2019-01-09
Replace complex constructors with simpler more specialized ones or set methods


rowanseymour:
Some constructor functions (e.g. `NewManualTrigger`) are getting a little unwieldy due to large number of arguments. Could try to make more specialized ones (`NewManualIVRTrigger`) or lean on `Set??` methods for not-so-common arguments.

See discussion at https://github.com/nyaruka/goflow/pull/476#discussion_r246418544

PR #480, opened 2019-01-14
🔫 Simpler trigger constructors

rowanseymour:
* Removes triggerOn from trigger constructors
* Adds two specialized constructors for IVR

Addresses #477


cmd/flowrunner/main.go
```diff
@@ -111,9 +111,9 @@ func RunFlow(assetsPath string, flowUUID assets.FlowUUID, initialMsg string, con
 
 	if initialMsg != "" {
 		msg := createMessage(contact, initialMsg)
-		repro.Trigger = triggers.NewMsgTrigger(env, flow.Reference(), contact, msg, nil, time.Now())
```
rowanseymour: I checked mailroom and this is always being set as now so.. goflow might as well set it itself like it does for resumes

PR #482, opened 2019-01-27
add full constructor for sprint

nicpottier:
Surveyor need a way of creating a sprint from events and modifiers, this just adds a constructor for that so I don't need to create my own Sprint class.

### DIFF
--- a/flows/engine/sprint.go
+++ b/flows/engine/sprint.go
@@ -9,16 +9,16 @@ type sprint struct {
 	events    []flows.Event
 }
 
-// NewSprint creates a new sprint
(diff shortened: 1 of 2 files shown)


flows/engine/sprint.go
```diff
@@ -17,6 +17,14 @@ func NewSprint() flows.Sprint {
 	}
 }
 
+// NewSprintWithValues creates a new sprint with the passed in events and modifiers
```
rowanseymour: can we call this `NewSprint` and the other one `NewEmptySprint`?

PR #483, opened 2019-01-28
Fix docs

norkans7:


norkans7:
Just markdown syntax is broken

https://github.com/nyaruka/goflow/blob/master/docs/md/flows.md#action

vs 

https://github.com/nyaruka/goflow/blob/fix-docs/docs/md/flows.md#action

rowanseymour:
ah ok - I never look at that - it's just a temporary file for making the final HTML, but let's fix it...

Issue #484, opened 2019-01-28
Session can't be read if channel assets no longer exist


rowanseymour:
`contact.urns` and `input.channel` have hard references to channels and err if channel doesn't exist.

rowanseymour:
@nicpottier exploring that last option in #486. Idea is that all ReadX methods for things in the session would report back missing assets as asset references instead of blowing up. To be clear this not just a surveyor problem as any time we resume a session (which starts by reading it), we might run into missing assets.

nicpottier:
I think that could work.. I guess then the flow would also still run but return errors when reaching said actions or are we still saying that at the time of execution all references need to be real? (I think I would prefer that honestly)

I'll still need to deal with missing items when dealing with events, as I can't throw out an event entirely if just one group is missing for example, but that should be fairly straightforward.

rowanseymour:
So there's two types of thing in goflow:

1. actions, events etc which only have references to assets and in the case of actions, only resolve those assets at runtime. In the case of actions, the references might be expressions which can only be resolved at runtime. Internally they have `asset.???Reference` structs. Goflow doesn't do anything with events.. just outputs them and considers them immutable.
2. contacts, inputs, modifiers, etc which resolve assets a read time. Internally they have pointers to real assets. Partly cuz cuz when they are involved in an expression, it's too late to start looking up assets.

There's also a question of correctness here - it might seem weird that the session contact loses fields and groups at read time - but if a group no longer exists.. then the contact is no longer part of that group etc.

I don't think we should change actions or events - in the case of the latter - it's up to the caller to decide at _apply_ time to ignore a group or field that it can't find. And we also don't want events to be dropped from the session because assets become missing later on - they're a historical record of what happened.

But I think for the latter class of things - we need to be flexible about missing assets at read time. If you call `ReadModifier` and it's an `add_groups` modifier with a group that exists and one that no longer exists, you should get back a modifier that still works for the existing group.

nicpottier:
Ya, just thinking aloud here as to how events are applied in mailroom and I think it works in that case too for two reasons:
  1. things like groups don't get hard deleted, just deactivated
  2. we aren't loading things like groups at that point, everything is already in our assets

[...]

PR #486, opened 2019-01-30
📣 Reading sessions should report missing assets instead of blowing up

rowanseymour:
Addresses #484


assets/references.go
```diff
@@ -11,12 +12,23 @@ func init() {
 	utils.Validator.RegisterStructValidation(LabelReferenceValidation, LabelReference{})
 }
 
+// Reference is interface for all reference types
+type Reference interface {
```
rowanseymour: this feels weird but maybe that's just the Java programmer in me. I have no way to explicitly make all the asset reference types be a particular interface.. anything that implements `String()` could pretend to be `asset.Reference`.. but what are the chances of that happening?

nicpottier: String() seems a bit broad for sure? Do all references have UUID()'s? Maybe they all have a Type()?

flows/inputs/base.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-type readFunc func(session flows.Session, data json.RawMessage) (flows.Input, error)
+type readFunc func(flows.SessionAssets, json.RawMessage, func(assets.Reference)) (flows.Input, error)
```
nicpottier: The `func(assets.Reference)` thing is a bit opaque. Would be better to make this a type so that it is a bit more descriptive. (the contact modifier application method suffers the same)

assets/references.go
rowanseymour: Unfortunately not (e.g. `FieldReference` has `key`)... they don't have any methods and if I start adding methods then they'll collide with the field names and I'll have to add envelopes for deserialization etc... argh.

Maybe the best way to think about this.. is what are we going to do with these? What functionality does the caller of `ReadSession` want on them. If it's just going to stringify them and log 'em, then maybe it's fine that it's essentially getting a bunch of Stringers? Honestly not sure what's the most go-ey approach here...

assets/references.go
```diff
@@ -11,12 +12,32 @@ func init() {
...
+	fmt.Stringer
+}
+
+// MissingCallback is callback to be invoked when an asset is missing
+type MissingCallback func(Reference)
+
+// PanicOnMissing panics if an asset is reported missing
+var PanicOnMissing MissingCallback = func(a Reference) { panic(fmt.Sprintf("unable to find asset %s", a.String())) }
```
rowanseymour: just using this for tests for convenience

assets/references.go
nicpottier: If References represent identity of an object, then it seems we would want the interface to be related to that somehow. Stringer says anything with that method matches which doesn't feel right. Maybe it is `Identity()` that returns a `string` that is the id in some way? (ie either UUID or label or key?) That feels less weird to me and gives us some hope of still having the compiler catching us doing stupid things.

rowanseymour: Added `Type()` and `Identity()`

Issue #475, opened 2018-12-21
Modifier loading


rowanseymour:
Modifiers may reference things that no longer exist. For example a surveyor submission handler might get a field modifier which references a field which no longer exists. It will try to load it.. it will fail.. it will move on because there's not much we can do about that at that point.

```go
mod, err := modifiers.ReadModifier(assets, modifierJSON)
log.debug("unable to load modifier: %s", err)
```

However the groups modifier is different.. it might reference some groups that do still exist and some that don't. It should return a modifier that will work for the groups that do still exist.. but what's the right way to communicate that from the read function?

My inclination is that it should quietly return a modifier for the groups that do exist, and only err if ends up with with zero groups that still exist. But that means no record of not being able to load particular groups which might not be ideal.

rowanseymour:
Fixed in #491

PR #491, opened 2019-01-31
🛂 Update modifier loading so missing assets are reported

rowanseymour:
Addresses #475. So now `ReadModifier` returns...

* a modifier and `nil` error if all assets exist
* a "partial" modifier and `nil` error if some assets exist (e.g. we still have some groups in a groups:add modifier) - with missing group assets reported via the callback
* a nil modifier and nil error if no assets exist and there is no partial modifier to return (with all missing assets reported via the callback)

The last one is the one I'm not sure about.. but I think it's consistent with saying "missing assets isn't actually an error case"

### DIFF
--- a/flows/actions/modifiers/base.go
+++ b/flows/actions/modifiers/base.go
@@ -3,6 +3,7 @@ package modifiers
 import (
 	"encoding/json"
 
+	"github.com/nyaruka/goflow/assets"
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/flows/events"
 	"github.com/nyaruka/goflow/utils"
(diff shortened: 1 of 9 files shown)


nicpottier:
So the parallel I can think of in go land is sqlx.Get(), which returns a special error and a nil result. The idea is that if you are asking for something then you really expect something back, but they use a constant error to let you deal with that case separately. So I think in Go land generally if it's a `thing, err` return, then when `err` is nil, `thing` is expected to be non-nill.

So case three would probably be more go-ish if it returned a constant `ErrNoModifier` instead. I'm honestly ok with it as you have it, but I think that would be more consistent with go.

rowanseymour:
@nicpottier that seems like a good approach.. lemme add...

PR #492, opened 2019-02-04
🔤 Add type to sessions

rowanseymour:
i.e. the type of the flow it was triggered with


flows/engine/session.go
```diff
@@ -69,8 +73,15 @@ func (s *session) SetContact(contact *flows.Contact) { s.contact = contact }
 func (s *session) Input() flows.Input         { return s.input }
 func (s *session) SetInput(input flows.Input) { s.input = input }
 
-func (s *session) CanEnterFlow(flow flows.Flow) bool {
-	return !s.flowStack.hasVisitedFlowSinceResume(flow.UUID())
+func (s *session) CanEnterFlow(flow flows.Flow) error {
+	if s.Type() != "" && s.Type() != flow.Type() {
```
rowanseymour: for now goflow will be flexible in case it is asked to resume a session that doesn't have type set..

Issue #493, opened 2019-02-07
Function to enumerate all asset references in a flow


rowanseymour:
Only complicated part here is that we also need to parse expressions for field references

ericnewcomer:
Groups can also be expressions among other things. Look here for the current checklist of where dependencies can live:
https://github.com/nyaruka/rapidpro/blob/27b5f975a761d878a55f84ecd31d96e322cff655/temba/flows/models.py#L2549

rowanseymour:
Added in #551

PR #496, opened 2019-02-07
🧪 Enumerable templates using reflection (WIP)

rowanseymour:
Bit of an experiment right but it does work. Idea is that we would be able to enumerate fields which are localizable and templatizable by looking at tags on the structs.

Would be used for...

- Getting all template values to extract field references
- Migrating expressions in template values ?
- Generate listings so we can more easily compare what the engine considers localizable and templatizable and what the editor considers ?
- Generating lists of translations for users https://github.com/rapidpro/rapidpro/issues/841 ?


nicpottier:
Nice, ya this seems like a great way to go!

rowanseymour:
@nicpottier I think the ultimate goal here could be the engine leaning on the engine tags to know which fields to localize and which to evaluate as templates - that's the only way to be sure that the engine behavior matches what the tags claim. But for now just want to be able to solve #493.

nicpottier:
Ya, no I think the approach is interesting.

Only caveats I can really think of are:

 1. the good old, reflection only works on public fields, so it kind of forces our hand to have everything be public. Go seems fine with that in general and it could be argued we go through way too many machinations to lock ourselves in that jail instead of just having the discipline not to use public fields

 2. tags don't work so well for discovery of subfields etc when those are dynamic since they are inherently static. In a way they are a bit of worst of both worlds in that there's no compile-time checking of anything with struct tags, things can fail at runtime, yet they don't provide the dynamic aspects of say, having an interface that hands back a set of valid subkeys on an object.

Not necessarily arguing against this approach, just voicing those thoughts that while neat they are pretty fragile as well.

rowanseymour:
Closing in favor of #547

PR #497, opened 2019-02-08
📏 Truncate contact field and run result values that are too long

rowanseymour:
I think we need a rethink about `EngineConfig` as all that was created when we had a `flowserver` cmd that read from a config file, but at the same time we wanted to override things at a request level for trialling it in RP... but for now this just honors some default length limits.


flows/engine/config.go
```diff
@@ -1,60 +1,38 @@
...
 import (
-	"encoding/json"
-
 	"github.com/nyaruka/goflow/flows"
 )
 
 // the configuration options for the flow engine
 type config struct {
+	maxFieldLength          int
+	maxResultLength         int
```
nicpottier: Is it useful for these ever being different? Results and contact values seem so closely tied that this seems potentially confusing to have as an option.

nicpottier:
I'm fine with this but ya this feels more like an env thing to me than an engine config. 

I guess it really could be either but in general I prefer the idea of settings being part of the env and the caller can decide if thats a 'global' thing vs an 'org' thing when calling into the engine. Would also make my mailroom life a lot easier since that's already captured there vs having to add a constructor for all engines being built. (not the end of the world, but yes, slightly easier)

nicpottier:
To add to this, I don't have to squint too hard to imagine longer field lengths as being a "premium feature" or the like, so that then becomes an org level thing, which feels more env like.

flows/engine/config.go
rowanseymour: Hmm maybe we should have a very generic `maxLength` value and apply it all over the place? Am thinking about other fields like `contact.name`, `result.category` and a nefarious user who wants to stuff huge data blobs somewhere.

nicpottier: Hrmm.. ya, how does this relate to say webhook results? That's part of the `extra` for that value right so wouldn't apply there?

It just seems like when these are different behavior can get really weird for users who love doing bizarre things like passing values through contact fields etc..

rowanseymour: Yeah for webhooks we read up to 10000 bytes and store that as extra on the result.

I could be persuaded that we get rid of `EngineConfig` completely at this point. Made a lot more sense having it when we had a stand alone flow server. We put `maxLength` (?) on `Environment` and we apply to contact.name, contact.field.value, result.value, result.category etc.

nicpottier: I could see `disableWebhooks` still being a thing on the engine config, at the very least for testing but ya, that'd be my preference.

nicpottier: Maybe just `maxValueLength` (contact values and result values?)

rowanseymour: If `disableWebhooks` and is the only thing left on `EngineConfig` then I think that can just become a static var in the module that uses it (we're not going to be using it) - and then we can stop passing `EngineConfig` instances around everywhere.

nicpottier: WFM

rowanseymour:
Replaced by #499

PR #510, opened 2019-02-13
📞 Verify parsed numbers in has_phone are valid

rowanseymour:


flows/routers/tests/tests.go
```diff
@@ -535,6 +535,10 @@ func HasPhone(env utils.Environment, text types.XText, args ...types.XValue) typ
 		return XFalseResult
 	}
 
+	if !phonenumbers.IsValidNumber(phone) {
```
rowanseymour: Tried `IsPossibleNumber` but it was accepting the error cases below

PR #511, opened 2019-02-13
📜 Better version support

rowanseymour:


flows/definition/flow.go
```diff
@@ -170,8 +173,26 @@ type flowEnvelopeWithUI struct {
 	UI flows.UI `json:"_ui,omitempty"`
 }
 
+// IsSpecVersionSupported determines if we can read the given flow version
+func IsSpecVersionSupported(ver *semver.Version) bool {
+	// major versions change flow schema
+	return ver.Major() <= CurrentSpecVersion.Major()
```
rowanseymour: Currently in the legacy world we use major minor versioning based on whether things will break Surveyor... but I think we should generalize and make it that the goflow engine itself will read minor versions ahead of itself.

Bump major versions for:
- Any schema changes (obviously these break flow reading)
- Any expression additions (new syntax, new functions)

Bump minor version for:
- Fixes of invalid state
- Rewriting expressions

nicpottier: Ya, that seems reasonable to me.

PR #513, opened 2019-02-14
🔍 Try to detect webhook response content type if header missing

### LINKED ISSUE #512: Detect JSON from webhooks rather than requiring content-type header
Use https://golang.org/pkg/net/http/#DetectContentType

rowanseymour:
Fixes #512

### DIFF
--- a/flows/webhook.go
+++ b/flows/webhook.go
@@ -77,6 +77,7 @@ type WebhookCall struct {
 	timeTaken     time.Duration
 	requestTrace  string
 	responseTrace string
+	bodyIgnored   bool
 }
 
 // MakeWebhookCall fires the passed in http request, returning any errors encountered. RequestResponse is always set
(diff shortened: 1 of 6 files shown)


flows/webhook.go
```diff
@@ -192,10 +216,13 @@ func newWebhookCallFromResponse(requestTrace string, response *http.Response, ma
...
 		}
 
+		if len(bodySniffed) > 0 {
+			bodyBytes = append(bodySniffed, bodyBytes...)
+		}
+
 		w.responseTrace += string(bodyBytes)
 	} else {
-		// no body for non-text responses but add it to our Response log so users know why
-		w.responseTrace += "Non-text body, ignoring"
```
rowanseymour: So instead of faking a response body `"Non-text body, ignoring"`, the event will have `body_ignored=true`. Up to the caller what they want to fake or not.

PR #515, opened 2019-02-15
🧺 Migrating expressions like flow.2factor using ["…"]

### LINKED ISSUE #514: Migrate properties starting with digits to use square bracket notation
`@flow.2factor` &rarr; `@(results["2factor"])`

rowanseymour:
Fixes #514


excellent/evaluator.go
```diff
@@ -50,9 +49,7 @@ func EvaluateTemplate(env utils.Environment, context types.XValue, template stri
 	// if we only have an identifier or an expression, evaluate it on its own
 	if nextTT == EOF {
 		switch tokenType {
-		case IDENTIFIER:
-			return ResolveValue(env, context, token), nil
-		case EXPRESSION:
+		case IDENTIFIER, EXPRESSION:
```
rowanseymour: Previously we've been parsing identifiers and expressions in subtly different ways - i.e. a whole identifier like `a.b.c` is passed to `ResolveValue` and it iterates through each property resolving them - but the expression `(a.b.c)` will be parsed by the visitor and `ResolveValue` will only be used to get the `b` of `a` and the `c` of `b`.

Consolidating these simplifies the code and means they have to play by the same parsing rules.

excellent/evaluator_test.go
```diff
@@ -85,7 +85,7 @@ func TestEvaluateTemplate(t *testing.T) {
 		{"@array1d", array1d},
 		{"@array1d.0", xs("a")},
 		{"@array1d.1", xs("b")},
-		{"@array2d.0.2", xs("c")},
+		{"@array2d.0.2", ERROR}, // need to use square brackets
```
rowanseymour: Previously `@(array2d.0.2)` was an error because the parser considers `0.2` to be a valid expression, but `@array2d.0.2` would parse because it's just splitting on `.`. Now they both go thru the parser.

flows/results_test.go
```diff
@@ -20,10 +20,10 @@ func TestResults(t *testing.T) {
 		expected types.XValue
 	}{
 		{[]byte(`{}`), "key", ERROR},
-		{[]byte(`{ "name": { "result_name": "Name", "value": "Ryan Lewis", "node": "uuid", "created_on": "2000-01-01T00:00:00.000000000-00:00"}}`), "key", ERROR},
-		{[]byte(`{ "name": { "result_name": "Name", "value": "Ryan Lewis", "node": "uuid", "created_on": "2000-01-01T00:00:00.000000000-00:00"}}`), "name", types.NewXText("Ryan Lewis")},
-		{[]byte(`{ "last_name": { "result_name": "Last Name", "value": "Lewis", "node": "uuid", "created_on": "2000-01-01T00:00:00.000000000-00:00"}}`), "last_name", types.NewXText("Lewis")},
-		{[]byte(`{ "last_name": { "result_name": "Last Name", "value": "Lewis", "node": "uuid", "created_on": "2000-01-01T00:00:00.000000000-00:00"}}`), "Last Name", types.NewXText("Lewis")},
+		{[]byte(`{ "name": { "result_name": "Name", "value": "Ryan Lewis", "node": "uuid", "created_on": "2000-01-01T00:00:00.000000000-00:00"}}`), `results.key`, ERROR},
```
rowanseymour: Having gotten rid of `ResolveValue` need to rewrite these tests to use actual expressions

excellent/evaluator_test.go
nicpottier: Maybe should add the `@array2d.0["2"]` as a test case?

nicpottier: `foo[123]` vs `foo["arst"]` vs `foo[foo.bar]` -> `foo.arst`, `foo.2auth` needs to be `foo["2auth"]`

nicpottier: does `foo[0]` match `{"0": "hi hi"}`?

Issue #516, opened 2019-02-15
Better time manipulation


rowanseymour:
Currently we don't have a good story for time manipulation which seems to be a fairly common use case. 

Currently something like `@(TODAY()+TIME(7, 30))`  becomes `@(datetime_add(today(), 27000, "s"))`

I.e. we don't have a time type so it becomes a date+seconds operation. The shortcomings of this are:

1. If the left-hand side might be a datetime that already has time information, there's no easy way to get just the date component of this.
2. The right-hand side has to be literals - we could fix this but the user would end up with a pretty complex looking expression.
3. No support for parsing a time value from text, like `TIMEVALUE` in legacy expressions.

Currently 3. means the expression `@(TODAY()+TIMEVALUE(flow.time))` becomes `@(datetime_add(today(), parse_datetime(results.time), "D"))` which is completely wrong.

I propose we make this usecase a lot simpler with the addition of 

- `time(num, num, num)` returns number of seconds-in-the-day as `XNumber`
- `parse_time(text)` returns number of seconds as `XNumber`
- `replace_time(datetime, num)` replaces all time data in the given datetime with the given time
- `format_time()` pretty prints a time value

But I could also be persuaded that time should be one of our primitive types... `XTime`

nicpottier:
I don't think we can represent time as an offset because of daylight savings time. IE, 3pm is not the same number of seconds from midnight depending on the day of the year. So we have to have a representation that really is the hour / minute / second.

So I think time probably has to be something closer to "11:30" as a string, as passing around tuple would be annoying.

How far could we get with just a `replace_time(datetime, str)` method along with a `has_time` that tests and returns something that is like `11:30pm` or `23:30`? I guess with the addition of a `get_time(datetime)` that returned the same.

That would make setting the time to today be `replace_time(today, "7:30pm")` or `replace_time(today, "19:30")`

I guess the question is then do we want a special type for that? It may not need to be something we expose to the user per say but may make reasoning on our side easier. Though we will probably have flexible parsing of string like times either way.

nicpottier:
So thinking about this a bit more I think we just adopt 8601 times, with an internal representation of the tersest `extended` format that includes both hour and minute.

IE. `22:30` or `22:30:30 or `22.30:30.500` but not just `22` and no internal representation of fractions for hour or minutes. (though we should support those and support commas for fractions)

We then add:
 * `PARSE_TIME(str) xtime, err` - parses things like `4:45pm` or `4pm` or just `7`
[...]

PR #519, opened 2019-02-18
⏰ Better Time Support

rowanseymour:
See #516


legacy/expressions/visitor.go
```diff
@@ -193,50 +178,35 @@ func (v *legacyVisitor) VisitAdditionOrSubtraction(ctx *gen.AdditionOrSubtractio
...
-		template := "datetime_add(%s, %s, \"%s\")"
-		if op == "-" {
-			template = "datetime_add(%s, -%s, \"%s\")"
-		}
+	if arg1Type == "number" && arg2Type == "number" {
+		// we are adding two numbers
+		return fmt.Sprintf("%s %s %s", arg1, op, arg2)
 
-		// determine the order of our parameters
-		replacements := []interface{}{value, next, dateUnit}
```
rowanseymour: order in legacy expressions has to be date+number so don't have to handle number+date

excellent/functions/functions.go
```diff
@@ -35,7 +35,8 @@ var XFUNCTIONS = map[string]XFunction{
 	"text":     OneArgFunction(Text),
 	"boolean":  OneArgFunction(Boolean),
 	"number":   OneArgFunction(Number),
-	"datetime": OneTextFunction(DateTime),
+	"datetime": OneArgFunction(DateTime),
+	"time":     OneArgFunction(Time),
```
rowanseymour: i could definitely be persuaded that these conversion funcs should called `to_xxx` and then `datetime_from_parts(...)` and `time_from_parts(...)` become `datetime(...)` and `time(...)`

legacy/expressions/visitor.go
```diff
@@ -193,50 +178,35 @@ func (v *legacyVisitor) VisitAdditionOrSubtraction(ctx *gen.AdditionOrSubtractio
...
-	} else {
-		// we are adding a field of unknown type with an integer
-		if op == "+" {
-			value = fmt.Sprintf("legacy_add(%s, %s)", value, next)
-		} else {
-			value = fmt.Sprintf("legacy_add(%s, -%s)", value, next)
-		}
+	} else if arg1Type == "datetime" && arg2Type == "time" && op == "+" {
+		// we are adding a date and a time
+		return fmt.Sprintf(`replace_time(%s, %s)`, arg1, arg2)
```
rowanseymour: making a bit of an assumption here that anyone trying to add a date and a time actually wants to replace rather than add the time value...

utils/timeofday.go
```diff
@@ -0,0 +1,67 @@
...
+// Format formats this time of day as a string
+func (t TimeOfDay) Format(layout string) string {
+	// upgrade us to a date time so we can use standard time.Time formatting
+	d := time.Date(1970, 1, 1, t.Hour, t.Minute, t.Second, t.Nanos, time.UTC)
+	return d.Format(layout)
+}
+
+// String returns the ISO8601 representation
+func (t TimeOfDay) String() string {
+	return t.Format(iso8601Time)
```
rowanseymour: Doesn't do anything clever with dropping nanoseconds if they're zero etc - but instead is consistent with formatting of time in datetimes.

flows/routers/tests/tests.go
```diff
@@ -488,6 +489,23 @@ func HasDateGT(env utils.Environment, text types.XText, date types.XDateTime) ty
 	return testDate(env, text, date, isDateGTTest)
 }
 
+// HasTime tests whether `text` contains a time.
+//
+//   @(has_time("the time is 10:30")) -> true
+//   @(has_time("the time is 10:30 PM").match) -> 22:30:00.000000
+//   @(has_time("the time is 10:30:45").match) -> 10:30:45.000000
+//   @(has_time("there is no time here, just a number 5")) -> false
```
rowanseymour: I think this is correct - we're strict about dates now (you need year+month+day).. so seems time needs to be more than just a single number

docs/functions.json
```diff
@@ -888,6 +934,25 @@
             }
         ]
     },
+    {
+        "signature": "replace_time(date)",
+        "summary": "Returns the a new date time with the time part replaced by the `time`.",
```
rowanseymour: I've slept on this and I still think `set_time` would be misleading.. can't think of any example where a set method returns a new value instead of modifying an argument 😬

excellent/functions/functions.go
nicpottier: I think the string versions are probably going to be used more so would rather keep those the short ones.

flows/routers/tests/tests.go
nicpottier: iso8601 says time can be simply `hh` so seems we should support just hour. (it also seems like a pretty common usecase really)

excellent/functions/functions.go
```diff
@@ -174,13 +181,29 @@ func Number(env utils.Environment, value types.XValue) types.XValue {
...
+}
 
-	return types.NewXDateTime(date)
+// Time tries to convert `value` to a time.
+//
+// An error is returned if the value can't be converted.
+//
+//   @(time("10:30")) -> 10:30:00.000000
+//   @(time(datetime("1979-07-18T10:30:45.123456Z"))) -> 10:30:45.123456
+//   @(time("what?")) -> ERROR
```
nicpottier: Looks like `10:30pm` will work here, so should probably call out that explicitly as an example.

utils/timeofday.go
nicpottier: I'm a bit sad about this but I suppose its consistent with our treatment of dates / locations.

excellent/functions/functions.go
```diff
@@ -1419,10 +1528,59 @@ func FormatDateTime(env utils.Environment, args ...types.XValue) types.XValue {
 		date = types.NewXDateTime(date.Native().In(location))
 	}
 
-	// return the formatted date
 	return types.NewXText(date.Native().Format(goFormat))
 }
 
+// FormatTime formats `time` as text according to the given `format`.
```
nicpottier: Docs should mention that if format is not specified then the default format is used.

utils/timeofday.go
rowanseymour: well not saying we shouldn't change it - just that times and dates should use the same formatting

docs/functions.json
nicpottier: You and your zillion character long expressions. So be it.

utils/timeofday.go
nicpottier: We are definitely training people in the new world that they need to use `format_` to get pretty things, so I think its consistent / fine.

For the record elastic doesn't have time support, so we would need to use a normalized time value for searching anyways, but really all that requires is zero padding the hour to guarantee lexical ordering. (IE, it doesn't really affect this decision either way, though from a purity point of view I agree having all our parsed times looks exactly the same sure feels right)

flows/routers/tests/tests.go
rowanseymour: I suppose my thinking here was that if you want to just ask for a number between 1 and 24 then we already have a rule for that... but yeah the common case is going to be people sending "2pm". I'll update our time parsing.

Should note that we don't currently support everything considered valid by ISO8601 - we don't support "hhmm" "hhmmss" or normalizing something like "24:00" to "00:00".

rowanseymour: https://github.com/nyaruka/goflow/issues/520

PR #522, opened 2019-02-19
🔂 Allow looping back into a flow that wasn't started in this sprint

rowanseymour:


cmd/flowrunner/main.go
```diff
@@ -27,11 +27,7 @@ var contactJSON = `{
 	"id": 1234567,
 	"name": "Ben Haggerty",
 	"created_on": "2018-01-01T12:00:00.000000000-00:00",
-	"fields": {
```
rowanseymour: otherwise your flow assets have to include this field

PR #524, opened 2019-02-20
🕰️ More flexible time parsing

rowanseymour:
* Accept hh, e.g. "12" or "2 PM"
 * Accept hhmm, e.g. "1400"
 * Accept "24:00" as midnight

### DIFF
--- a/excellent/types/time.go
+++ b/excellent/types/time.go
@@ -41,7 +41,7 @@ func (x XTime) String() string { return x.ToXText(nil).Native() }
 
 // Equals determines equality for this type
 func (x XTime) Equals(other XTime) bool {
-	return x.Native().Equals(other.Native())
+	return x.Native().Equal(other.Native())
 }
 
 // Compare compares this date to another
@@ -66,7 +66,7 @@ func ToXTime(env utils.Environment, x XValue) (XTime, XError) {
 		case XDateTime:
 			return typed.Time(), nil
 		case XText:
-			parsed, err := utils.TimeFromString(env, typed.Native())
+			parsed, err := utils.TimeFromString(typed.Native())
 			if err == nil {
 				return NewXTime(parsed), nil
 			}
--- a/utils/dates.go
+++ b/utils/dates.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"bytes"
+	"math"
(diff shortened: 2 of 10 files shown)


excellent/types/time.go
```diff
@@ -41,7 +41,7 @@ func (x XTime) String() string { return x.ToXText(nil).Native() }
 
 // Equals determines equality for this type
 func (x XTime) Equals(other XTime) bool {
-	return x.Native().Equals(other.Native())
+	return x.Native().Equal(other.Native())
```
rowanseymour: just making `utils.TimeOfDay.Equal` match `time.Time.Equal`

utils/dates_test.go
```diff
@@ -111,6 +111,56 @@ func TestDateFromString(t *testing.T) {
...
 }
 
+func TestTimeFromString(t *testing.T) {
+	timeTests := []struct {
+		value    string
+		expected utils.TimeOfDay
+		hasError bool
+	}{
+		{"it's 10 ok", utils.NewTimeOfDay(10, 0, 0, 0), false},
+		{"it's 10 PM ok", utils.NewTimeOfDay(22, 0, 0, 0), false},
```
nicpottier: `10:30pm` test case?

utils/dates.go
```diff
@@ -64,36 +65,40 @@ var isoFormats = []string{iso8601Format, iso8601NoSecondsFormat, iso8601Date}
 // ZeroDateTime is our uninitialized datetime value
 var ZeroDateTime = time.Time{}
 
-func dateFromFormats(env Environment, currentYear int, pattern *regexp.Regexp, d int, m int, y int, str string) (time.Time, error) {
+func dateFromFormats(env Environment, currentYear int, pattern *regexp.Regexp, d int, m int, y int, str string) (time.Time, string, error) {
 
-	matches := pattern.FindAllStringSubmatch(str, -1)
+	matches := pattern.FindAllStringSubmatchIndex(str, -1)
```
rowanseymour: previously time parsing happened on the same input as date parsing, but allowing single digits as hours means that the time regex matches against date components - so now we parse the date and return the remainder of the input - and only that is parsed for a time value

Issue #525, opened 2019-02-20
Add date as type


rowanseymour:
In legacy expressions, `today()` returns a `date` without time, and when that stringifies it becomes  an ISO date like `2019-02-20`

Currently in new expressions, we don't have a date type, so `today()` returns a full `XDateTime` which stringifies like `2019-02-20T00:00:00.000000Z`

We should add `XDate` as a type and update `today()` to return that to match existing behavior.

Side note: legacy context also includes `date.today`, `date.yesterday` and `date.tomorrow`. These have always been org formatted date strings, and they currently migrate to various `format_date(now() +/- 1)` calls to achieve the same result.

nicpottier:
Do we still do `+/-` on dates? Thought we removed that? (Or rather I think its `LEGACY_ADD` no?)

So would `today()` return an org formatted date or an iso date? If the latter then we would need to migrate `date.today` and `today()` to `FORMAT_DATE(NOW())` ya?

rowanseymour:
yes sorry not an actual +/-, anymore those migrations are:

`date.now` &rarr; `now()`
`date.today` &rarr;`format_date(now())`
`date.tomorrow` &rarr; `format_date(datetime_add(now(), 1, "D"))`
`date.yesterday` &rarr; `format_date(datetime_add(now(), -1, "D"))`

`today()` in old world returns `date` object and so would `today()` in new world

nicpottier:
by `date` object you mean `XDate` ya? So ISO8601 date would be returned by `today()`.

So maybe more natural translation for `date.today` is -> `format_date(today())` ?

rowanseymour:
Well in legacy world it's returning a Python `date` and in new world it'll be a `XDate`. To me it's confusing to talk about a returning an ISO8601 date - we're returning an `XDate` which happens to stringify as an ISO8601 date.

agree re `date.today` -> `format_date(today())`

nicpottier:
Sure sure, though remember user never sees `XDate` just ISO dates, and any expression that takes an `XDate` will also take an ISO string so I think its important to talk that way as well.

rowanseymour:
Sure sure, it just gets confusing when talking also about the legacy context because everything in that literally is a string.

nicpottier:
To the user everything is still a string. :)

PR #526, opened 2019-02-20
📅 Timeless dates

rowanseymour:
Addresses #525


excellent/functions/functions.go
```diff
@@ -1421,7 +1418,7 @@ func JSON(env utils.Environment, value types.XValue) types.XValue {
 //
 // @function format_date(date, [,format])
 func FormatDate(env utils.Environment, args ...types.XValue) types.XValue {
-	date, xerr := types.ToXDateTime(env, args[0])
+	date, xerr := types.ToXDate(env, args[0])
```
rowanseymour: this will currently get the date in the env timezone, which matches legacy behavior for things like `DAY()` etc

PR #528, opened 2019-02-21
📆 Date function tweaks

rowanseymour:
More tweaks in the quest for perfect compatibility with legacy expressions..

* `datetime_from_parts` &rarr; `date_from_parts` (`DATE` migrates to this )
* added `date` (`DATEVALUE` migrates to this )

### DIFF
--- a/docs/functions.json
+++ b/docs/functions.json
@@ -143,7 +143,7 @@
     {
         "signature": "date(value)",
         "summary": "Tries to convert `value` to a date.",
-        "detail": "If it is text then it will be parsed into a date using to the default date format.\nAn error is returned if the value can't be converted.",
+        "detail": "If it is text then it will be parsed into a date using the default date format.\nAn error is returned if the value can't be converted.",
         "examples": [
             {
                 "template": "@(date(\"1979-07-18\"))",
@@ -185,7 +185,7 @@
     {
         "signature": "datetime(value)",
         "summary": "Tries to convert `value` to a datetime.",
-        "detail": "If it is text then it will be parsed into a datetime using to the default date\nand time formats. An error is returned if the value can't be converted.",
+        "detail": "If it is text then it will be parsed into a datetime using the default date\nand time formats. An error is returned if the value can't be converted.",
         "examples": [
             {
                 "template": "@(datetime(\"1979-07-18\"))",
@@ -1164,7 +1164,7 @@
     {
         "signature": "time(value)",
         "summary": "Tries to convert `value` to a time.",
-        "detail": "If it is text then it will be parsed into a time using to the default time format.\nAn error is returned if the value can't be converted.",
+        "detail": "If it is text then it will be parsed into a time using the default time format.\nAn error is returned if the value can't be converted.",
         "examples": [
             {
                 "template": "@(time(\"10:30\"))",
--- a/docs/functions.json
+++ b/docs/functions.json
@@ -142,8 +142,8 @@
     },
     {
         "signature": "date(value)",
-        "summary": "Tries to convert `value` to a date. If it is text then it will be",
-        "detail": "parsed into a date using to the default date format.\n\nAn error is returned if the value can't be converted.",
+        "summary": "Tries to convert `value` to a date.",
+        "detail": "If it is text then it will be parsed into a date using to the default date format.\nAn error is returned if the value can't be converted.",
         "examples": [
             {
                 "template": "@(date(\"1979-07-18\"))",
@@ -184,8 +184,8 @@
     },
     {
         "signature": "datetime(value)",
-        "summary": "Tries to convert `value` to a datetime. If it is text then it will be",
-        "detail": "parsed into a datetime using to the default date and time formats.\n\nAn error is returned if the value can't be converted.",
+        "summary": "Tries to convert `value` to a datetime.",
+        "detail": "If it is text then it will be parsed into a datetime using to the default date\nand time formats. An error is returned if the value can't be converted.",
         "examples": [
             {
                 "template": "@(datetime(\"1979-07-18\"))",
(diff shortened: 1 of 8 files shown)


docs/functions.json
```diff
@@ -141,9 +141,51 @@
         ]
     },
     {
-        "signature": "datetime(text)",
-        "summary": "Parses `text` into a date using to the default date format.",
-        "detail": "An error is returned if the value can't be converted.",
+        "signature": "date(value)",
+        "summary": "Tries to convert `value` to a date.",
+        "detail": "If it is text then it will be parsed into a date using to the default date format.\nAn error is returned if the value can't be converted.",
```
nicpottier: typo.. `using the`

nicpottier: does this work for MM-DD-YYYY? and DD-MM-YYYY? depending on org format? might be good to have an example there.

docs/functions.json
```diff
@@ -141,9 +141,51 @@
...
+            {
+                "template": "@(date_from_parts(2017, 13, 15))",
+                "output": "ERROR"
+            }
+        ]
+    },
+    {
+        "signature": "datetime(value)",
+        "summary": "Tries to convert `value` to a datetime.",
+        "detail": "If it is text then it will be parsed into a datetime using to the default date\nand time formats. An error is returned if the value can't be converted.",
```
nicpottier: same typo

docs/md/expressions.md
```diff
@@ -393,15 +393,46 @@ It is the inverse of [char](expressions.html#function:char).
...
-<a name="function:datetime"></a>
+<a name="function:date"></a>
 
-## datetime(text)
+## date(value)
 
-Parses `text` into a date using to the default date format.
+Tries to convert `value` to a date.
 
+If it is text then it will be parsed into a date using to the default date format.
```
nicpottier: "if it is text" may be a bit confusing to people.. its all text to them :)

docs/functions.json
rowanseymour: problem is all these examples run against the same default environment which happens to be YYY-MM-DD - could change that to DD-MM-YYYY

docs/md/expressions.md
rowanseymour: I think (at least some of our) users understand that when they do `1 + 2` or `x != "bar"` they are working with non-text types

nicpottier: `1+2` sure, but I think its a lot to expect for them to reason about something that is stringy looking not being a string. Whatever, keep it if you want, but I think we should be careful with this kind of language and remember their point of view.

nicpottier: I would just rather see something like "if it is not an ISO8601 date" which would provide a bit more clue as to how this is going to behave.

rowanseymour: But that's not even accurate - we'll parse lots of things that aren't ISO8601 dates

nicpottier: isn't saying "If this isn't an ISO 8601 date then this will be parsed according to your org settings" accurate?

rowanseymour: No because `DATE(2019)` won't try to parse anything. It'll say sorry we can't convert a number to a date.

nicpottier: But `DATE("2019")` does work? Does that feel right to you?

rowanseymour: `DATE("2019")` will return an error saying `"2019"` couldn't be parsed as a date... yes that feels right

nicpottier: I feel like that is way too subtle. Remember the parameter may be a contact field or a flow value that is an integer and to the user that is going to be really weird that those behave differently. When they echo that value it is going to look identical. `2019` and `2019` Why wouldn't we try to cast the argument to a string so we are consistent in both?

rowanseymour: I honestly don't get why it's a bad thing to give the user some insight into what types are being created by different expressions, and supposed I disagree with your fundamental premise that users are oblivious to intermediate types :):):):)

`DATE(2019+2+19)` returning an error saying that a number couldn't be converted to a `DATE` gives them insight into what `2019+2+19` does. 

If they're using a contact field it might remind them that fields can be `TEXT` or `NUMBER` etc

nicpottier: You keep using literals, but that's not going to be the typical case.

Getting that error for `DATE(contact.birth_year)` is going to be super weird to a user, especially when they echo `@contact.birth_year` and `@contact.birth_year_string` and they look the same.

We are creating shit work for a user to make them cast into a string, why would we do that when we can try it ourselves. I don't understand the argument against it, can you explain that, I must be missing something.

rowanseymour: If `contact.fields.birth_year` is a NUMBER like `1980`.. then I think the ideal error message for `DATE(contact.fields.birth_year)` is 

`the number 1980 couldn't be converted to a date`

It reminds the user that `contact.fields.birth_year` is a number field.

"We are creating shit work for a user to make them cast into a string"

I'm not following.. what to do they need to cast to a string?

rowanseymour: (not trying to kill this conversion but I need to merge this to fix a ticket...)

nicpottier: Ok, maybe this isn't such a big deal for `DATE` since it doesn't accept a single year either way, but what happens in the `TIME` case where we do? Do we accept both `"11"` and `11`? (or any other function for that matter)

My concern is places where we act one way vs another for a type that the user can't really see. I can be convinced different errors are maybe ok though I don't love it, but I'm not ok with one succeeding for `"10"` and failing for `10` or vice versa.

rowanseymour: We don't currently accept `TIME(10)` but that's only cuz I forgot to update that when we decided a single hour was a valid time.

I think that's actually a useful principle moving forward - that `foo(text(bar))` should work if `foo(bar)` works. I just think it's beneficial to let the user see the actual type of thing passed to `foo` if it fails (and certainly more beneficial to one of us trying to debug a flow).

nicpottier: Ok, then we are in agreement. I think the inverse should also be true `foo(bar)` should succeed if `foo(int(bar))` does.

PR #531, opened 2019-02-21
🧯 Fix calling length on a complex object that needs to be reduced to a primitive

rowanseymour:
e.g. `contact.fields.gender` resolves to a `FieldValue` object which has no intrinsic length - so it needs to be first reduced to a primitive (it's `.value`)

### DIFF
--- a/excellent/functions/functions.go
+++ b/excellent/functions/functions.go
@@ -1716,14 +1716,14 @@ func FormatURN(env utils.Environment, arg types.XText) types.XValue {
 func Length(env utils.Environment, value types.XValue) types.XValue {
 	// argument must be a value with length
 	lengthable, isLengthable := value.(types.XLengthable)
-	if isLengthable {
+	if isLengthable && !utils.IsNil(lengthable) {
 		return types.NewXNumberFromInt(lengthable.Length())
 	}
 
-	// or reducable to something with length
+	// or reduceable to something with length
 	value = types.Reduce(env, value)
 	lengthable, isLengthable = value.(types.XLengthable)
-	if isLengthable {
+	if isLengthable && !utils.IsNil(lengthable) {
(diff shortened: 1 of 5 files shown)


excellent/functions/functions.go
```diff
@@ -1719,6 +1720,13 @@ func Length(env utils.Environment, value types.XValue) types.XValue {
 		return types.NewXNumberFromInt(lengthable.Length())
 	}
 
+	// or reducable to something with length
+	value = types.Reduce(env, value)
+	lengthable, isLengthable = value.(types.XLengthable)
```
nicpottier: can value be nil?

rowanseymour: calling `types.Reduce(..)` instead of `value.Reduce()` for that reason because it handles `nil`.. but now you've got me worried about `value.(types.XLengthable)` returning an interface to a nil...

nicpottier: Ya, that was my question there.

nicpottier: Sounds like for interfaces, nil is never true.

rowanseymour: I think if you passed it an `XValue` interface to a `nil` that implemented `XLengthable`, then `isLengthable` would be true - the same problem we fixed with `@parent.contact`.. tho in this case `types.Reduce(..)` returns a real `nil` if you pass it an interface to a `nil`

nicpottier: I fee l like we really need to figure out a strategy for this.. feels like we must be doing something wrong or approaching things wrong to get us in a state where it is so subtle. Just smells like endless bug squashing.

rowanseymour: maybe that strategy should be to never create a non-nil interface to a nil.. I need to dig into this more and figure out where they're coming from in the first place.

Issue #401, opened 2018-10-09
Formatting for @results


rowanseymour:
In old engine, `@flow` renders a list like 

```
key1: value1
key2: value2
...
```

which you sometimes see in things like `EmailAction`. In new world `@results` renders as raw JSON. Maybe we can add function like `@(report(results))`...

rowanseymour:
So think we have to add this since we have at least one user wondering why `@flow` in his SendEmail is now becoming raw JSON. I think the solution is a function that can take something map like, i.e.

- `@contact.fields`
- `@results`, `@parent.results` or `@child.results`
- An `XMap`
- An `XJSONObject` fragment 

Namewise probably should be something like `format_results` or `format_map` to match the other `format_` functions

nicpottier:
Ya, I like this. `_map` feels too techy and `_report` seems too vague.

What about just a `format()` that you pass anything to and we do our best?

rowanseymour:
By anything you mean anything or anything map like? I just mean does `format()` work for numbers and dates etc ?

ericnewcomer:
I think it should be valid for anything. Worst case it’s a noop. Users don’t necessarily know what the type is. It’s basically pretty print for whatever.

nicpottier:
Ya, I think if you pass in something that we know the type of, then we do our best.

IE `format(DATE("arst"))` is roughly same as `format_date("arst")`

nicpottier:
same with URN etc..

rowanseymour:
Yeah like that. Having a think about how to make this work with any complex object. Right now all XValues have a `ToXJSON`. That's typically implemented by returning a map of `string` -> `XValue` and then recursively JSONifying everything in that map - but if you took each value in that map and stringified it instead, you'd get a nice string -> string map.

```
@(format(contact)) =>

name: Bob
language: eng
...

```

nicpottier:
Yep yep.. though how is nesting done? tab / space to show that?

rowanseymour:
Well I was thinking no nesting - i.e. every value is just the TEXT version of that value, i.e. so you get

```
gender: male
age: 24 
```

and not

```
gender:
  text: male
age:
  text: 24
  number: 24
```

nicpottier:
What's the JSON version? Maybe they should have the same rough structure?
What happens with arbitrary JSON when you pass it to this?
What about lists? IE urns?

rowanseymour:
[...]

PR #538, opened 2019-02-22
💅 Pretty maps and arrays

rowanseymour:
Addresses #401


cmd/docgen/renderers.go
```diff
@@ -258,7 +259,8 @@ func checkExample(session flows.Session, line string) error {
 	}
 
 	test := strings.TrimSpace(pieces[0])
-	expected := strings.Replace(strings.TrimSpace(pieces[1]), "\\n", "\n", -1)
+	expected := strings.Replace(strings.TrimSpace(pieces[1]), `\n`, "\n", -1)
+	expected = strings.Replace(expected, `\x20`, " ", -1)
```
rowanseymour: need a way of encoding a trailing space in an example output

ericnewcomer:
+1 for emoji

flows/engine/session_test.go
```diff
@@ -79,12 +79,12 @@ func TestEvaluateTemplateAsString(t *testing.T) {
...
 		{"@input", "Hi there\nhttp://s3.amazon.com/bucket/test.jpg\nhttp://s3.amazon.com/bucket/test.mp3", ""},
 		{"@input.text", "Hi there", ""},
-		{"@input.attachments", `["http://s3.amazon.com/bucket/test.jpg","http://s3.amazon.com/bucket/test.mp3"]`, ""},
+		{"@input.attachments", `http://s3.amazon.com/bucket/test.jpg, http://s3.amazon.com/bucket/test.mp3`, ""},
 		{"@(input.attachments[0])", "http://s3.amazon.com/bucket/test.jpg", ""},
 		{"@input.created_on", "2017-12-31T11:35:10.035757-02:00", ""},
 		{"@input.channel.name", "My Android Phone", ""},
 
-		{"@results", `{"2factor":"34634624463525","favorite_color":"red","phone_number":"+12344563452"}`, ""},
+		{"@results", "2Factor: 34634624463525\nFavorite Color: red\nPhone Number: +12344563452", ""},
```
rowanseymour: note this is using result name as key

nicpottier:
So urns are displaying their `format_urn` version ya? not their `string(urn)` version? So in this case stringed version of the higher element is showing formatted of sub-elements. Maybe that's ok just want to voice it.

rowanseymour:
`text(x)` will always be what `x` stringifies to.. one is just explicit, other implicit

Currently for a URN that looks like...

```
@(contact.urns[0]) == "tel:+1234567890"
@(text(contact.urns[0])) == "tel:+1234567890"
@(format_urn(contact.urns[0])) == "(234) 567-890"
```

nicpottier:
But `@(contact)` -> something with urns that are formatted right?

nicpottier:
What I'm doing a poor job explaining is I'm wondering if having `text()` on a meta item turn inner items into non `text` is right.

IE, I think I would expect `text` to call `text` on everything inner.

Just as I would expect a `format()` to call `format()` on everything inner.

rowanseymour:
In the case of maps and arrays, `text()` is a concatenation of `text()` of each item in that thing

For complex objects there's no concrete rule, `text(contact)` is the name, `text(run)` is the UUID, `text(results.response_1)` is the value.

Where are you seeing an inner item turned into non `text`?

Then we currently have a limited set of `format_` methods for humanzing/customizing stringification of some primitive types but we haven't generalised that to all types.

nicpottier:
Ok cool.. I think we're on the same page then. For some reason I thought I saw somewhere where URNs were being turned into displays but I must have misread the HTML diff there.

Issue #539, opened 2019-02-22
Date migration error


ericnewcomer:
Given the following user legacy expression:
`@(DATEVALUE(date.today) + TIMEVALUE(CONCATENATE(flow.time_input, ":00")))`

Errors with the following:
`error evaluating @(legacy_add(date(format_date(today())), time(results.time_input))): error calling LEGACY_ADD: unable to convert date to a number`

PR #540, opened 2019-02-22
🚁 Fix migration of datevalue() + time

rowanseymour:
So cool we found the last migration of legacy date arithmetic bug (#539) 😬


nicpottier:
Just think Monday we could come in to 2 days on the sign!

PR #542, opened 2019-02-23
🔧 Fix remove_first_word so that punctuation is preserved

rowanseymour:


excellent/functions/functions.go
```diff
@@ -466,15 +466,23 @@ func Word(env utils.Environment, text types.XText, args ...types.XValue) types.X
...
-	if len(words) > 1 {
-		return types.NewXText(strings.Join(words[1:], " "))
+	firstWordVal := Word(env, text, types.XNumberZero)
+	firstWord, isText := firstWordVal.(types.XText)
+	if !isText || firstWord == types.XTextEmpty {
+		return types.XTextEmpty
 	}
 
-	return types.XTextEmpty
+	firstWordStart := strings.Index(text.Native(), firstWord.Native())
```
rowanseymour: I'm copying the legacy implementation.. except it looks its wrong when the first word doesn't start at index 0.. so instead I get the first word, find the end of it, return everything after that

PR #544, opened 2019-02-25
🚀 Refactor expression parsing to better support different kinds of visitors

rowanseymour:
And adds 2 new utilities that use custom visitors:

1. `tools.RefactorTemplate` - for now just rewrites and tidies a template but this would be the starting point for supporting flow spec migrations that rewrites expressions.
2. `tools.FindContextRefsInTemplate` - for now just provides a callback for context references in expressions but we'd use this to find contact field references during flow dependency calculation.


cmd/docgen/renderers.go
```diff
@@ -263,7 +263,7 @@ func checkExample(session flows.Session, line string) error {
 	expected = strings.Replace(expected, `\x20`, " ", -1)
 
 	// evaluate our expression
-	val, err := session.Runs()[0].EvaluateTemplateAsString(test)
+	val, err := session.Runs()[0].EvaluateTemplate(test)
```
rowanseymour: this is the 99% case so it gets the simple name

excellent/tools/context_refs_test.go
```diff
@@ -0,0 +1,43 @@
...
+		paths    []string
+		hasError bool
+	}{
+		{``, []string{}, false},
+		{`Hi @foo @foo.bar`, []string{`foo`, `foo`, `foo.bar`}, false},
+		{`@foo.bar.x.y`, []string{`foo`, `foo.bar`, `foo.bar.x`, `foo.bar.x.y`}, false},
+		{`@(foo) @(foo.bar)`, []string{`foo`, `foo`, `foo.bar`}, false},
+		{`@((foo))`, []string{`foo`}, false},
+		{`@(lower(foo.bar))`, []string{`foo`, `foo.bar`}, false},
+		{`@(foo["bar"])`, []string{`foo`, `foo.bar`}, false},
```
rowanseymour: would support `contact.fields["age"]` just in case

excellent/tools/context_refs.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// VisitAtomReference deals with visiting a single atom in our expression
+func (v *auditContextVisitor) VisitAtomReference(ctx *gen.AtomReferenceContext) interface{} {
+	return v.Visit(ctx.Atom())
+}
+
+// VisitFunctionCall deals with function calls like TITLE(foo.bar)
+func (v *auditContextVisitor) VisitFunctionCall(ctx *gen.FunctionCallContext) interface{} {
+	return v.VisitChildren(ctx)
```
rowanseymour: go's lack of polymorphism hurts a bit here.. would be great if there was a base class that called `VisitChildren` by default on every visit function you don't override.. but that base wouldn't be able to call the functions that you have overridden

Issue #534, opened 2019-02-21
Investigate creation of non-nil interfaces to nils


rowanseymour:
There are various places where we use `utils.IsNil(..)` to check for non-nil interfaces to nils. Should figure out if we can avoid these being created at all

rowanseymour:
I think this is happening when a nil struct is being returned from a function that returns an interface type, e.g. https://play.golang.org/p/9NfbUUV-zw8

rowanseymour:
Closing this as it's just going to be a fact of life. Trying to write code that explicitly returns nil interfaces but not sure I'll ever find all the places we don't do this.

rowanseymour:
Tried and failed again to fix this but we have too much code like this...

```golang
type Contextable interface {
	Context(env envs.Environment) map[string]types.XValue
}

func Context(env envs.Environment, contextable Contextable) types.XValue {
	// contextable becomes a non-nil interface boxing a nil *Contact
}

var contact *Contact
Context(contact)
```

Argh.. the biggest foot-gun in Go.

PR #551, opened 2019-03-01
🤓 Record dependencies and result names in the definition after flow validation

rowanseymour:


extensions/transferto/action.go
```diff
@@ -45,8 +45,8 @@ func NewTransferAirtimeAction(uuid flows.ActionUUID, amounts map[string]decimal.
 	}
 }
 
-// Validate validates our action is valid and has all the assets it needs
-func (a *TransferAirtimeAction) Validate(assets flows.SessionAssets, context *flows.ValidationContext) error {
+// Validate validates our action is valid
+func (a *TransferAirtimeAction) Validate() error {
```
rowanseymour: actions now only do structural validation on themselves - they're not responsible for determining if their dependencies exist - we do that centrally

flows/actions/testdata/add_contact_groups.json
```diff
@@ -11,7 +11,7 @@
                 }
             ]
         },
-        "validation_error": "no such group with UUID '33382939-babf-4982-9395-8793feb4e7c6'"
+        "validation_error": "missing dependencies: group[uuid=33382939-babf-4982-9395-8793feb4e7c6,name=Climbers]"
```
rowanseymour: this should never end up being a user facing error.. assuming we track dependencies correctly

flows/definition/dependencies.go
```diff
@@ -0,0 +1,85 @@
...
+			missing(ref)
+		} else {
+			d.Channels[i] = a.Reference()
+		}
+	}
+	for i, ref := range d.Fields {
+		a, err := sa.Fields().Get(ref.Key)
+
+		if err != nil {
+			// TODO for now if a field reference came from an expression (i.e. no name), we don't blow up if it's missing
```
rowanseymour: chatted with @nicpottier and we can look at addressing this in the old world but there are just too many flows out there that likely to have expressions referencing fields that don't exist

flows/definition/flow.go
```diff
@@ -95,12 +105,38 @@ func (f *flow) Validate(assets flows.SessionAssets, context *flows.ValidationCon
...
+		depStrings := make([]string, len(missingAssets))
+		for d := range missingAssets {
+			depStrings[d] = missingAssets[d].String()
+		}
+		return errors.Errorf("missing dependencies: %s", strings.Join(depStrings, ","))
+	}
+
 	f.valid = true
+	f.dependencies = deps
+	f.resultNames = f.ExtractResultNames()
```
rowanseymour: I think this is threadsafe... it's possible that multiple threads validate this flow object at the same time, but it shouldn't matter

flows/definition/flow.go
```diff
@@ -173,10 +209,12 @@ func (f *flow) ExtractDependencies() []assets.Reference {
 	dependencies := make([]assets.Reference, 0)
 	dependenciesSeen := make(map[string]bool)
 	addDependency := func(r assets.Reference) {
-		key := fmt.Sprintf("%s:%s", r.Type(), r.Identity())
-		if r != nil && !r.Variable() && !dependenciesSeen[key] {
-			dependencies = append(dependencies, r)
-			dependenciesSeen[key] = true
+		if !utils.IsNil(r) && !r.Variable() {
```
rowanseymour: See https://github.com/nyaruka/goflow/issues/534

flows/actions/start_session.go
```diff
@@ -53,15 +53,9 @@ func NewStartSessionAction(uuid flows.ActionUUID, flow *assets.FlowReference, ur
...
-// Validate validates our action is valid and has all the assets it needs
-func (a *StartSessionAction) Validate(assets flows.SessionAssets, context *flows.ValidationContext) error {
-	// check the flow exists and that it's valid
-	if err := a.validateFlow(assets, a.Flow, context); err != nil {
-		return err
-	}
-
-	// finally check that all the groups exist
-	return a.validateGroups(assets, a.Groups)
+// Validate validates our action is valid
```
nicpottier: Maybe no-op Validate should be part of BaseAction?

PR #557, opened 2019-03-04
🗺️ Store a map of result keys to result names in a validated flow definition

rowanseymour:
Upgrades the list of result names returned from flow validation to... a dict of result keys to the result names that map to that key.

I took quick look at the UReport code to see what it needs and whether if this sort of thing was included on the flows.json API endpoint it would be sufficient - but it's using things like rule UUIDs... so I'm thinking UReport might need to continue fetching flow definitions.

@norkans7 @nicpottier

### DIFF
--- a/flows/definition/results_info.go
+++ b/flows/definition/results_info.go
@@ -4,18 +4,28 @@ import (
 	"github.com/nyaruka/goflow/utils"
 )
 
+type resultInfo struct {
+	Names []string `json:"names"`
+}
+
 // holds information about what results a flow can generate, as a map of result
 // keys to slices of result names, e.g.
 //
-//  { "age": ["Age"], "response_1": ["Response 1", "Response-1"] }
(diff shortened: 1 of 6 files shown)


flows/definition/results_info.go
```diff
@@ -0,0 +1,21 @@
+package definition
+
+import (
+	"github.com/nyaruka/goflow/utils"
+)
+
+// holds information about what results a flow can generate, as a map of result
+// keys to slices of result names, e.g.
+//
+//  { "age": ["Age"], "response_1": ["Response 1", "Response-1"] }
```
nicpottier: From a slightly more work now but a lot simpler in the future, what do you think of this being a key -> map instead? That would let us add keys in the future (`type`?) without changing the structure of the results for all flows.

IE, something more like:
```json
{ 
  "age":  { "labels": ["Age"] },
  "response_1": { "labels": ["Response 1", "Response-1"] }
}
```

PR #563, opened 2019-03-07
🌐 Absolutify attachment URLs during flow migration

rowanseymour:


legacy/definition.go
```diff
@@ -440,6 +440,17 @@ func migrateAction(baseLanguage utils.Language, a Action, localization flows.Loc
 			quickReplies = TransformTranslations(legacyQuickReplies)
 		}
 
+		for lang, attachment := range media {
+			parts := strings.SplitN(attachment, ":", 2)
+			mediaType := parts[0]
+			mediaURL := parts[1]
+
+			if strings.Contains(mediaType, "/") {
```
ericnewcomer: Shouldn't we be looking for `@` ?

rowanseymour: this is differentiating between _foo/bar:relativeurl_ and _foo:template_ .. the thing about the latter is it doesn't necessarily contain a `@` expression. This matches how `ReplyAction` determines whether or not to add `STORAGE_URL` in the legacy engine.

legacy/definition.go
```diff
@@ -440,6 +440,17 @@ func migrateAction(baseLanguage utils.Language, a Action, localization flows.Loc
 			quickReplies = TransformTranslations(legacyQuickReplies)
 		}
 
+		for lang, attachment := range media {
+			parts := strings.SplitN(attachment, ":", 2)
+			mediaType := parts[0]
+			mediaURL := parts[1]
```
ericnewcomer: Should we enforce that it doesn't (or alternatively does) end in a `/` so we don't double slash below?

legacy/definition.go
ericnewcomer: Ah, makes sense

legacy/definition.go
rowanseymour: The legacy engine assumes they don't start with `/` but I'll add this for good measure

PR #565, opened 2019-03-07
🔧 Return results as list of name/key objects during flow validation

rowanseymour:
Cuz order is import for things like result exports


flows/definition/results_info.go
```diff
@@ -4,32 +4,16 @@ import (
 	"github.com/nyaruka/goflow/utils"
 )
 
+// holds information about a possible result in the flow
 type resultInfo struct {
-	Names []string `json:"names"`
+	Name string `json:"name"`
+	Key  string `json:"key"`
```
rowanseymour: @nicpottier doing no merging here across same labels etc in case in future we want to add node specific stuff like uuid or inferred type

PR #566, opened 2019-03-08
🔀 Change random router to return full random value as result value

rowanseymour:
And add some router specific tests!


flows/routers/base.go
```diff
@@ -46,7 +51,9 @@ func (r *BaseRouter) EnumerateDependencies(localization flows.Localization, incl
 }
 
 // EnumerateResultNames enumerates all result names on this object
-func (r *BaseRouter) EnumerateResultNames(include func(string)) {}
+func (r *BaseRouter) EnumerateResultNames(include func(string)) {
+	include(r.ResultName())
+}
```
rowanseymour: previously only returning result names for switch routers which was a bug... funny how adding tests always turns up bugs 😬

PR #567, opened 2019-03-08
🙏 More forgiving flow validation

rowanseymour:
Option to pass a missing assets callback into `Flow.ValidateRecursively` which can then be called instead of returning a validation error for missing dependencies

### DIFF
--- a/cmd/flowrunner/main.go
+++ b/cmd/flowrunner/main.go
@@ -114,7 +114,7 @@ func RunFlow(assetsPath string, flowUUID assets.FlowUUID, initialMsg string, con
 	session := eng.NewSession(sessionAssets)
 
 	// start our session
-	sprint, err := session.Start(repro.Trigger)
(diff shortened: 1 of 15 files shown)


rowanseymour:
Problem with this PR is that validation happens inside a trigger or in `Session.Resume`. Seems preferable if the caller could do asset validation before creating a session. Easier for the caller to get a list of missing assets. So maybe starts and resumes work like...

```
err, missingAssets := flow.Validate(sa)
// if err, don't start session
// if missing assets log to sentry
```

```
flow := sa.Flows().Get(session.WaitingRun().Flow().UUID)
err, missingAssets := flow.Validate(sa)
// if err, don't resume session
// if missing assets log to sentry
```

nicpottier:
Ya, I prefer that honestly.

rowanseymour:
Replaced by #576

PR #568, opened 2019-03-10
🔥 Fix migration of datetime + time in legacy expressions

rowanseymour:
One last fix for migration of legacy date arithmetic expressions before I go start selling 🍢

Fixes https://app.intercom.io/a/apps/l6e7tr01/inbox/inbox/155936/conversations/20906198371

### DIFF
--- a/legacy/expressions/visitor.go
+++ b/legacy/expressions/visitor.go
@@ -210,7 +210,7 @@ func (v *legacyVisitor) VisitAdditionOrSubtraction(ctx *gen.AdditionOrSubtractio
 		// we are adding a datetime and a time
 
 		// create expression which converts arg2 to minutes
-		asMinutes := fmt.Sprintf(`format_time(%[1]s, "h") * 60 + format_time(%[1]s, "m")`, arg2)
+		asMinutes := fmt.Sprintf(`format_time(%[1]s, "tt") * 60 + format_time(%[1]s, "m")`, arg2)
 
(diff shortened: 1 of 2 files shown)


legacy/expressions/visitor.go
```diff
@@ -210,7 +210,7 @@ func (v *legacyVisitor) VisitAdditionOrSubtraction(ctx *gen.AdditionOrSubtractio
 		// we are adding a datetime and a time
 
 		// create expression which converts arg2 to minutes
-		asMinutes := fmt.Sprintf(`format_time(%[1]s, "h") * 60 + format_time(%[1]s, "m")`, arg2)
+		asMinutes := fmt.Sprintf(`format_time(%[1]s, "tt") * 60 + format_time(%[1]s, "m")`, arg2)
```
rowanseymour: curiously golang date formatting doesn't support away to get a non-zero padded 24 hour hour-of-day value

nicpottier: Seems this is a bug in `FormatTime` then as well no? Or are we removing the option to use `h`? Seems we need to support it but we'll need to do it ourselves.

rowanseymour: `FormatTime` currently supports `h`, `hh`, and `tt` but not `t` (which is also not listed as an option in the docs). Agree we should find a way to add support for `t` tho.

rowanseymour: https://github.com/nyaruka/goflow/issues/569

PR #574, opened 2019-03-14
👮 Step limit instead of loop detection

rowanseymour:


putorti:
@nicpottier or @rowanseymour Is this configurable? We're running into loads of problems right now as our letter writing flow is complicated and it's being interrupted in a majority of cases before completing now.

rowanseymour:
@putorti not currently but just created https://github.com/nyaruka/mailroom/issues/131

chris-erickson:
What is the fix in the meantime?  This is breaking production flows for a lot of people...

rowanseymour:
First thing I'd check is that this isn't an unintended loop. If not, can you re-organize the flow? Now that we have `set_run_result` actions, you might be able to replace some rulesets with actions and group those actions in a single node to reduce the number of nodes. If not, then you make quick adjustment in your own mailroom fork... and failing all that.. try to implement https://github.com/nyaruka/mailroom/issues/131 cuz it seems like a pretty straightforward tweak.

chris-erickson:
This is a series of flows that are very mature but we can look into it.  Reorganizing isn't the best option, especially since it was operating well prior to this new limit.

- Can you point us to docs on how the set run result will help these types of problems?
- We've been following guidance on building things for quite some time (perhaps imperfectly) - how should we structure flows that's different than in the past?
- How do we know when we'll hit this limit other than running traffic through it?
- Can we deploy the version prior to the limit addition to temporarily get past this?

Thanks for the help!

nicpottier:
1. No docs, but basically because in new flows you can have an action for setting a result (instead of doing a split) you can probably reduce a great number of steps by 3 or more. IE, if you were using a split by expression to update a result or calculate something that can now just be in the action, so that will reduce a series of action -> split -> action to just one action and that can obviously continue to simplify. You can really only do this in new editor so probably not going to help you short term though, but I'd be really surprised if you can't get under 100 steps per sprint with those new tools.
2. See 1), addition of new set result should really help. Note that the reason this limit exists is because we are unlocking other things. You can do looping on a conditional expression now without waiting for a message, this lets you build for/while loops in a flow which is super powerful but then you have a halting problem. The solution to that is putting an upper bound on number of steps per sprint, otherwise you could generate a zillion messages with an infinite loop you build or fill your DB etc..
3. Just like an infinite loop, you don't, test test test.
4. I don't think so no, that change was made a while back. That's a possible argument for keeping more up to date in that you would have discovered this earlier and at that point yes a revert probably would have been possible.

chris-erickson:
What is a sprint, and what is a step?  You know how we use nested flows to call web hooks and start building the users state - can you relate this to how we should target our efforts?

nicpottier:
A step is a block in a flow. That's either a split or a set of actions. Note that one step may contain multiple actions but that will only be counted as one step.

A sprint is what we call a resume or start of a flow before it reaches a wait (for a reply). So on a very simple flow that asks for your age, waits for a response, then replies with it, that is three steps. The first sprint executes the first step to send the message, then moves to the "wait for response", that's a second step, and those two steps consist of the first sprint. When the person responds with their age, that's a new sprint, the first step will record the response, then the next step will echo back that age.

putorti:
Is there a way to migrate existing flows to the new editor? It seems that an export/re-import, even if I rename, will not work. It appears I will have to recreate each flow one at a time, and swap them into place.

nicpottier:
No official way just yet, that's on the docket for the next few days.

putorti:
One more oddity I've noticed in the new editor, it doesn't seem that you can "create" flow results. You have to somehow introduce the flow result name into the namespace ahead of using that action, since it uses a dropdown for the name. I can do this now by doing an expression split like in the old engine. But for example if I want to send the variable `Mixpanel Event` to a child metrics flow, there's no natural way to set that without faking it with a split and then deleting the split later. Am I missing something?

For example this was impossible without doing a split first, then deleting:

image

PR #576, opened 2019-03-14
🔏 Add validation at session assets level which returns missing assets

rowanseymour:
Replaces #567.

### DIFF
--- a/flows/actions/set_contact_field.go
+++ b/flows/actions/set_contact_field.go
@@ -64,9 +64,11 @@ func (a *SetContactFieldAction) Execute(run flows.FlowRun, step flows.Step, logM
 	fields := run.Session().Assets().Fields()
 	field := fields.Get(a.Field.Key)
 
-	newValue := run.Contact().Fields().Parse(run.Environment(), fields, field, rawValue)
+	if field != nil {
+		newValue := run.Contact().Fields().Parse(run.Environment(), fields, field, rawValue)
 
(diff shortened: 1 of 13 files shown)


flows/actions/set_contact_field.go
```diff
@@ -64,9 +64,11 @@ func (a *SetContactFieldAction) Execute(run flows.FlowRun, step flows.Step, logM
 	fields := run.Session().Assets().Fields()
 	field := fields.Get(a.Field.Key)
 
-	newValue := run.Contact().Fields().Parse(run.Environment(), fields, field, rawValue)
+	if field != nil {
```
rowanseymour: this was only the only place where an action would blow up on a missing asset

flows/definition/dependencies.go
```diff
@@ -47,13 +47,8 @@ func (d *dependencies) refresh(sa flows.SessionAssets, missing assets.MissingCal
 	}
 	for i, ref := range d.Fields {
 		a := sa.Fields().Get(ref.Key)
-
 		if a == nil {
-			// TODO for now if a field reference came from an expression (i.e. no name), we don't blow up if it's missing
```
rowanseymour: since caller will now get back list of missing assets, it's up to them how strictly to deal with contact fields

nicpottier:
So to be clear events will only be created for assets that exist in the session is that right? (just need to think through whether that needs to change in mailroom)

rowanseymour:
yeah - if a group/label/field doesn't exist at runtime, we're going to noop that

PR #579, opened 2019-03-15
Fix "multiple-value uuid.NewV4() in single-value context"

johncordeiro:
When you try to build the current project, this error happens:

```
# github.com/nyaruka/goflow/utils
../goflow/utils/uuid.go:22:24: multiple-value uuid.NewV4() in single-value context
```

This PR replaces the current library to a more consistent one for generating UUIDs.


nicpottier:
Build is passing so not sure how that would be the case. What version of go are you on? Are you outside your $GOPATH? (you should be, this is built with modules)

But this is a problem with your environment not goflow.

PR #580, opened 2019-03-19
📅 Fix HasDate tests to compare dates in env timezone

rowanseymour:


legacy/testdata/rulesets.json
```diff
@@ -1175,6 +1175,118 @@
             }
         }
     },
+    {
```
rowanseymour: just testing existing migration of date rules - we can add to this when we figure out what should populated in the `_ui` section

PR #585, opened 2019-03-22
🔍 Extract and save result categories during validation

rowanseymour:


utils/misc.go
```diff
@@ -39,3 +39,13 @@ func DeriveCountryFromTel(number string) string {
 	}
 	return phonenumbers.GetRegionCodeForNumber(parsed)
 }
+
+// StringSliceContains determines whether the given slice of strings contains the given string
+func StringSliceContains(slice []string, str string) bool {
```
rowanseymour: ah golang where everybody writes their own contains function for each type...

nicpottier:
What about my exit uuids!?

PR #586, opened 2019-03-22
Match characters  intended to be combined with another character to s…

norkans7:
Addresses https://github.com/rapidpro/rapidpro/issues/898

### DIFF
--- a/utils/text.go
+++ b/utils/text.go
@@ -8,7 +8,7 @@ import (
 var snakedChars = regexp.MustCompile(`[^\p{L}\d_]+`)
 
 // treats sequences of letters/numbers/_/' as tokens, and symbols as individual tokens
-var wordTokenRegex = regexp.MustCompile(`[\pL\pN_']+|\pS`)
+var wordTokenRegex = regexp.MustCompile(`[\pM\pL\pN_']+|\pS`)
 
 // Snakify turns the passed in string into a context reference. We replace all whitespace
 // characters with _ and replace any duplicate underscores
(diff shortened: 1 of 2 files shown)


utils/text_test.go
```diff
@@ -42,7 +42,10 @@ func TestTokenizeString(t *testing.T) {
 		{"math+=×÷√∊", []string{"math", "+", "=", "×", "÷", "√", "∊"}},     // math symbols treated as individual tokens
 		{"emoji😄🏥👪👰😟🧟", []string{"emoji", "😄", "🏥", "👪", "👰", "😟", "🧟"}},   // emojis treated as individual tokens
 		{"👍🏿 👨🏼", []string{"👍", "🏿", "👨", "🏼"}},                            // tone modifiers treated as individual tokens
-		{"ℹ︎ ℹ️", []string{"ℹ", "ℹ"}},                                      // variation selectors ignored
+		{"ℹ ℹ️", []string{"ℹ", "ℹ️"}},                                      // variation selectors ignored
```
norkans7: FWIW It seems the first i had some non printable character next to it, I had to remove that as well both here and in expressions

nicpottier: What does this tokenize to? `หอบาเบลทั่วแผ่นดินโลกมีภาษาเดียวและมีสำเนียงเดียวกัน`

norkans7: that หอบาเบลทั่วแผ่นดินโลกมีภาษาเดียวและมีสำเนียงเดียวกัน

utils/text.go
```diff
@@ -8,7 +8,7 @@ import (
 var snakedChars = regexp.MustCompile(`[^\p{L}\d_]+`)
 
 // treats sequences of letters/numbers/_/' as tokens, and symbols as individual tokens
-var wordTokenRegex = regexp.MustCompile(`[\pL\pN_']+|\pS`)
+var wordTokenRegex = regexp.MustCompile(`[\pM\pL\pN_']+|\pS`)
```
rowanseymour: FYI this includes...

https://www.fileformat.info/info/unicode/category/Mn/list.htm
https://www.fileformat.info/info/unicode/category/Mc/list.htm
https://www.fileformat.info/info/unicode/category/Me/list.htm

There's some weird stuff in `Me`.. but maybe for simplicity best to just match all of `M` like you have it above.

utils/text_test.go
nicpottier: Ya, so we kind of have to accept that we can't really tokenize these languages. Still think this PR is the right thing but we aren't really breaking up words with our tokenize here, rather just NOT breaking up individual characters. (that sentence above translates to "The Babel tower throughout the earth has one language and has the same accent.")

norkans7: Yep

PR #587, opened 2019-03-22
🚪 Record exit UUIDs coming from waits in validated flow definition

rowanseymour:


flows/definition/flow.go
```diff
@@ -258,6 +262,21 @@ func (f *flow) ExtractResults() []*flows.ResultSpec {
...
 
+// ExtractExitsFromWaits extracts all exits coming from nodes with waits
+func (f *flow) ExtractExitsFromWaits() []flows.ExitUUID {
+	exitUUIDs := make([]flows.ExitUUID, 0)
+	include := func(e flows.ExitUUID) { exitUUIDs = append(exitUUIDs, e) }
+
+	for _, n := range f.nodes {
+		if n.Wait() != nil {
+			for _, e := range n.Exits() {
+				include(e.UUID())
```
rowanseymour: So this isn't perfect but it matches the existing behavior in RP. If we wanted to exclude timeouts we'd have to work back from exits to categories to cases to tests.. and well... I do think the correct solution here is changing RP to look at message counts over time and not this path data and having mailroom record that instead.

nicpottier: Is that really so much more code to have less confusing stats?

rowanseymour: Well yes..  and this whole thing feels like a hack to avoid properly tracking responses over time. Aren't we going to add something like `flows_flowmsgcount` at some point?

rowanseymour: And also if the graph the user is looking at is labeled "responses over time"... then stats shouldn't be disappearing because certain exits have been removed from the flow.

PR #589, opened 2019-03-25
🔥 Switch router should use category from first matching rule

rowanseymour:
Bug introduced in categories vs exits refactor

### DIFF
--- a/flows/routers/switch.go
+++ b/flows/routers/switch.go
@@ -131,11 +131,7 @@ func (r *SwitchRouter) PickExit(run flows.FlowRun, step flows.Step, logEvent flo
 		run.LogError(step, err)
 	}
 
-	// the ingredients for a result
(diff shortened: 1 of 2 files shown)


flows/routers/switch.go
```diff
@@ -179,34 +196,18 @@ func (r *SwitchRouter) PickExit(run flows.FlowRun, step flows.Step, logEvent flo
...
+				resultAsStr, xerr := types.ToXText(run.Environment(), typedResult.Match())
 				if xerr != nil {
-					return "", xerr
+					return "", "", nil, xerr
 				}
 
-				match = resultAsStr.Native()
-				categoryUUID = c.CategoryUUID
-				extra = typedResult.Extra()
-				break
```
rowanseymour: breaks out of the `switch` but not the `for`

Issue #592, opened 2019-03-26
Make location parsing a little more forgiving


rowanseymour:
Currently when parsing a value as a location, the rule is that if the value contains `>`, the lookup has be an exact match. If the value contains a period* at the end, no match.

We could normalize the value being looked up so that

- Tokenize and rejoin the value around `>` so the separators are always `\x20>\x20 `
- Normalize spacing within tokens (`\w+` becomes `\x20`)

* The final terminating period is specific and common problem with the old editor

PR #594, opened 2019-03-26
Normalize value looked up to match locations

norkans7:
Addresses https://github.com/nyaruka/goflow/issues/592

### DIFF
--- a/utils/location.go
+++ b/utils/location.go
@@ -14,6 +14,8 @@ const (
 	LocationPaddedPathSeparator = " > "
 )
 
+var spaceRegex = regexp.MustCompile(`\s+`)
+
 // Location represents a single Location
 type Location struct {
 	level    LocationLevel
@@ -165,7 +167,6 @@ func (h *LocationHierarchy) FindByName(name string, level LocationLevel, parent 
 // FindByPath looks for a location in the hierarchy with the given path
 func (h *LocationHierarchy) FindByPath(path string) *Location {
 	tokens := strings.Split(path, LocationPathSeparator)
-	spaceRegex := regexp.MustCompile(`\s+`)
 	for i := range tokens {
 		tokens[i] = spaceRegex.ReplaceAllString(strings.ToLower(strings.TrimSpace(tokens[i])), " ")
 	}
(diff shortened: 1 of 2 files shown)


utils/location.go
```diff
@@ -163,7 +163,12 @@ func (h *LocationHierarchy) FindByName(name string, level LocationLevel, parent
 
 // FindByPath looks for a location in the hierarchy with the given path
 func (h *LocationHierarchy) FindByPath(path string) *Location {
-	return h.pathLookup.lookup(strings.ToLower(path))
+	tokens := strings.Split(path, ">")
+	for i := range tokens {
+		tokens[i] = strings.TrimSpace(tokens[i])
+	}
+	normalizedPath := strings.ToLower(strings.TrimRight(strings.Join(tokens, "\x20>\x20"), "."))
```
rowanseymour: don't have to use `\x20` here instead of just a space - i just put that in the in the issue cuz it isn't obvious that ` > ` has spaces

utils/location.go
```diff
@@ -163,7 +163,12 @@ func (h *LocationHierarchy) FindByName(name string, level LocationLevel, parent
 
 // FindByPath looks for a location in the hierarchy with the given path
 func (h *LocationHierarchy) FindByPath(path string) *Location {
-	return h.pathLookup.lookup(strings.ToLower(path))
+	tokens := strings.Split(path, ">")
```
rowanseymour: there are already constants for this and " > " as used below

utils/location.go
```diff
@@ -163,7 +163,12 @@ func (h *LocationHierarchy) FindByName(name string, level LocationLevel, parent
...
+	for i := range tokens {
+		tokens[i] = strings.TrimSpace(tokens[i])
```
rowanseymour: I think we can do a little bit more normalization here of each location token within the path
  1. trim whitespace
  2. replace `\w+` with a single space
  3. lowercase (don't need to do it again below)

so ` Kigali\x20\t\x20City ` becomes `kigali city`

utils/location.go
```diff
@@ -163,7 +164,13 @@ func (h *LocationHierarchy) FindByName(name string, level LocationLevel, parent
 
 // FindByPath looks for a location in the hierarchy with the given path
 func (h *LocationHierarchy) FindByPath(path string) *Location {
-	return h.pathLookup.lookup(strings.ToLower(path))
+	tokens := strings.Split(path, LocationPathSeparator)
+	spaceRegex := regexp.MustCompile(`\s+`)
```
rowanseymour: best to make this a global

PR #595, opened 2019-03-26
🌿 Add @fields

rowanseymour:
1. Adds `@fields`, `@parent.fields`, `@child.fields`
2. Exposes field values as simple map
3. Removes `@contact.fields.age.text` as a resolveable thing (no legacy equivalent)


flows/field.go
```diff
@@ -137,7 +126,6 @@ func (v *FieldValue) ToXJSON(env utils.Environment) types.XText {
 }
 
 var _ types.XValue = (*FieldValue)(nil)
```
rowanseymour: `FieldValue` is still a complex object that needs to reduce itself to a primitive... I want to get rid of this... but I'm a bit stumped on what to do with locations and generally whether locations and URNs are types at an excellent level or just formatting behaviors of `XText`

flows/engine/session_test.go
```diff
@@ -70,9 +70,18 @@ func TestEvaluateTemplate(t *testing.T) {
 		{"@contact.fields.activation_token", "AACC55", ""},
 		{"@contact.fields.age", "23", ""},
 		{"@contact.fields.join_date", "2017-12-02T00:00:00.000000-02:00", ""},
-		{"@contact.fields.favorite_icecream", "", "error evaluating @contact.fields.favorite_icecream: no such contact field 'favorite_icecream'"},
+		{"@contact.fields.favorite_icecream", "", "error evaluating @contact.fields.favorite_icecream: map has no property 'favorite_icecream'"},
```
rowanseymour: This is a bit of a lame error message coming from the fact that these complex objects are now just `XMap` instances instead of having their own `Resolve` methods with custom error handling. But I think we can find generic solution like `Context()` methods also returning a name to be used for error messages, or leaning on the expression itself, e.g. `contact.fields has no property`

nicpottier: Ya I like `contact.fields has no property 'favorite_icecream'`

PR #596, opened 2019-03-26
📋  Templates in goflow

nicpottier:


assets/interfaces.go
```diff
@@ -181,6 +182,53 @@ type Resthook interface {
...
+
+// TemplateTranslation represents a single translation for a specific template and channel
+type TemplateTranslation interface {
+	Content() string
+	Language() utils.Language
+	VariableCount() int
+	Channel() ChannelReference
+}
+
+// TemplateAssets represents our interface to get a template
```
rowanseymour: don't think this is used?

assets/static/types/template_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/nyaruka/goflow/assets"
+	"github.com/nyaruka/goflow/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestChannel(t *testing.T) {
```
rowanseymour: name

assets/references.go
```diff
@@ -193,6 +194,33 @@ func (r *LabelReference) String() string {
 
 var _ Reference = (*LabelReference)(nil)
 
+// TemplateReference is used to reference a Template
```
rowanseymour: this isn't used anywhere

docs/md/flows.md
```diff
@@ -741,7 +741,14 @@ A [msg_created](sessions.html#event:msg_created) event will be created with the
 {
     "type": "send_msg",
     "uuid": "8eebd020-1af5-431c-b943-aa670fc74da9",
-    "text": "Hi @contact.name, are you ready to complete today's survey?"
+    "text": "Hi @contact.name, are you ready to complete today's survey?",
+    "template": {
+        "uuid": "3ce100b7-a734-4b4e-891b-350b1279ade2",
+        "name": "revive_issue",
```
rowanseymour: maybe this is where a `TemplateReference` could be used?

```
"templating": {
    "template": {
        "uuid": "3ce100b7-a734-4b4e-891b-350b1279ade2",
        "name": "revive_issue",
    },
    "variables": [
        "@contact.name"
    ]
}
```

flows/template.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return &TemplateAssets{
+		templates: templates,
+		byName:    byName,
+	}
+}
+
+// FindTranslation looks through our list of templates to find the template matching the passed in name
+// If no template or translation is found then empty string is returned
+func (l *TemplateAssets) FindTranslation(name string, channel *assets.ChannelReference, langs []utils.Language) *TemplateTranslation {
```
rowanseymour: wouldn't it be better to lean on UUID here instead of name? I'm just trying think where we might go in the future with templates, and it seems possible in the future that they are editable in RP and thus the UUID is the immutable identifier of a template, and name is the thing that might be stale in a flow definition.

nicpottier:
going to tweak some naming so don't merge this just yet.

assets/interfaces.go
nicpottier: Oh right, ya, started with `send_msg` having both `template` and `template_variables` attributes but then decided it made more sense to have a single attribute with everything.

docs/md/flows.md
nicpottier: Sure if you prefer, let references spread far and wide!

flows/template.go
nicpottier: Man, I feel we are just making work for ourselves with these uuids but sure I will change this.

rowanseymour: Maybe.. or maybe we're future proofing ourselves to using templates differently

docs/md/flows.md
rowanseymour: well it also helps with enumerating asset references in actions.. if this uses a template reference.. the action can just return that

flows/msg.go
```diff
@@ -27,7 +27,8 @@ type MsgIn struct {
 type MsgOut struct {
 	BaseMsg
 
-	QuickReplies_ []string `json:"quick_replies,omitempty"`
+	QuickReplies_ []string       `json:"quick_replies,omitempty"`
+	Templating_   *MsgTemplating `json:"templating,omitempty"`
```
nicpottier: Find the naming slightly awkward here but seems consistent with the action so I can convince myself it is ok.

nicpottier:
all better @rowanseymour

assets/interfaces.go
rowanseymour: I think this interface still isn't used 

(and comment was meant for https://github.com/nyaruka/goflow/pull/596#discussion_r270184407)

flows/actions/testdata/send_msg.json
```diff
@@ -196,5 +196,84 @@
...
+                    "text": "Hi Ryan Lewis, who's a good boy?",
+                    "uuid": "59d74b86-3e2f-4a93-aece-b05d2fdcde0c"
+                },
+                "step_uuid": "e7187099-7d38-4f60-955c-325957214c42",
+                "type": "msg_created"
+            }
+        ]
+    },
+    {
+        "description": "Msg with a matching template",
```
rowanseymour: seems worth adding a `inspection` section to one of these test cases (see first test case in file) to check that template dependences.. templates within templates are being enumerated correctly... darn that word is so overloaded now

nicpottier: Sure.. and what's the magic to get it to recognize the template as a dependency?

nicpottier: nm, figured it out I think, PR updated.

flows/msg.go
```diff
@@ -86,3 +88,31 @@ func (m *MsgIn) SetExternalID(id string) { m.ExternalID_ = id }
...
 func (m *MsgOut) QuickReplies() []string { return m.QuickReplies_ }
+
+// Templating returns the templating to use to send this message (if any)
+func (m *MsgOut) Templating() *MsgTemplating { return m.Templating_ }
+
+// MsgTemplating represents any substituted message template that should be applied when sending this message
+type MsgTemplating struct {
+	Template_  assets.TemplateReference `json:"template"`
+	Language_  utils.Language           `json:"language"`
+	Variables_ []string                 `json:"template_variables"`
```
rowanseymour: this could be just `variables` like the action since it's now inside in a bigger thing called templating

nicpottier: Doh ya.. damn renames! Thanks for all the stupid catches.

rowanseymour:
looks good.. kinda wish there was a different word for template assets vs excellent templates.. but I can't think of anything obvious.. so 🤷‍♂️

nicpottier:
Should be good to go now @rowanseymour

PR #600, opened 2019-03-27
🔬 Router test results as dicts in expressions

rowanseymour:


flows/engine/session_test.go
```diff
@@ -71,7 +71,7 @@ func TestEvaluateTemplate(t *testing.T) {
 		{"@contact.fields.age", "23", ""},
 		{"@contact.fields.join_date", "2017-12-02T00:00:00.000000-02:00", ""},
 		{"@contact.fields.favorite_icecream", "", "error evaluating @contact.fields.favorite_icecream: dict has no property 'favorite_icecream'"},
-		{"@(is_error(contact.fields.favorite_icecream))", "true", ""},
+		{"@(is_error(contact.fields.favorite_icecream))", "{match: dict has no property 'favorite_icecream'}", ""},
```
rowanseymour: this makes me wonder if there shouldn't be a `is_error` excellent function which just returns true/false and this becomes `has_error` router test which returns the dict. Have we decided if router test functions are going to be treated like regular function in terms of auto-completion etc ?

nicpottier: I think having the test functions available (and in autocomplete) is desirable, they expose lots of interesting things that people may want to use in more advanced splits by expression etc.

I'm ok with an `is_error` that is separate from that though.

PR #606, opened 2019-03-29
🧮 Functional programming

rowanseymour:


antlr/Excellent2.g4
```diff
@@ -1,71 +1,71 @@
 grammar Excellent2;
 
-// rebuild with % antlr4 -Dlanguage=Go Excellent2.g4 -o ../excellent/gen -package gen -visitor
+// rebuild with % antlr -Dlanguage=Go Excellent2.g4 -o ../excellent/gen -package gen -visitor
 
 import LexUnicode;
 
-COMMA      : ',';
```
rowanseymour: gave up fighting with vscode and let it reformat this file...

excellent/evaluator_test.go
```diff
@@ -221,7 +223,7 @@ func TestEvaluateTemplate(t *testing.T) {
 		"dec1":    types.RequireXNumberFromString("1.5"),
 		"dec2":    types.RequireXNumberFromString("2.5"),
 		"words":   types.NewXText("one two three"),
-		"array":   types.NewXArray(types.NewXText("one"), types.NewXText("two"), types.NewXText("three")),
+		"array1":  types.NewXArray(types.NewXText("one"), types.NewXText("two"), types.NewXText("three")),
```
rowanseymour: now you can't have (top-level) names in the context that conflict with function names

antlr/Excellent2.g4
```diff
@@ -1,71 +1,69 @@
...
+// variable names, e.g. contact.name or function names, e.g. SUM
+
+WS: [ \t\n\r]+ -> skip; // ignore whitespace
+
+ERROR: .;
+
+parse: expression EOF;
+
+atom:
+	atom LPAREN parameters? RPAREN		# functionCall
```
rowanseymour: this is still stricter than it needs to be, for example `(upper)()` won't work.. will investigate if it can be loosened more another time tho

excellent/functions/functions.go
```diff
@@ -21,107 +21,105 @@ import (
...
-	"format_urn":      OneTextFunction(FormatURN),
-
-	// utility functions
-	"length":     OneArgFunction(Length),
-	"default":    TwoArgFunction(Default),
-	"legacy_add": TwoArgFunction(LegacyAdd),
-	"read_chars": OneTextFunction(ReadChars),
-	"extract":    InitialArrayFunction(1, -1, Extract),
+func init() {
+	std := map[string]types.XFunction{
```
rowanseymour: apparently i was creating an `initialization loop` so moving this into `init()` avoids that

excellent/functions/functions.go
```diff
@@ -1677,10 +1683,6 @@ func FormatTime(env utils.Environment, args ...types.XValue) types.XValue {
 //
 // @function format_number(number, places [, humanize])
 func FormatNumber(env utils.Environment, args ...types.XValue) types.XValue {
-	if len(args) < 1 || len(args) > 3 {
```
rowanseymour: unrelated - just noticed don't need this when function is wrapped with `ArgCountCheck`

antlr/Excellent2.g4
```diff
@@ -1,71 +1,69 @@
...
+	| atom DOT prop = (NAME | NUMBER)	# dotLookup
```
rowanseymour: for now needes to make this a bit more strict, but to be addressed further in https://github.com/nyaruka/goflow/issues/607

docs/functions.json
```diff
@@ -341,23 +341,15 @@
     },
     {
         "signature": "extract(array, properties...)",
-        "summary": "Takes an array of objects and returns a new array by extracting named properties from each item.",
-        "detail": "If a single property is specified, the returned array is a flat array of values. If multiple properties\nare specified then each item is a dict of with those properties.",
+        "summary": "Takes a dict and returns a new dict by extracting only them named properties.",
```
nicpottier: sig parameter should probably be `map`

nicpottier: the named properties.

docs/functions.json
```diff
@@ -341,23 +341,15 @@
...
+        "detail": "If a single property is specified, the function returns that single value. If multiple properties\nare specified the returned value is a new dict with those properties.",
```
nicpottier: For the single case, doesn't that violate our rule about the type of functions being known? Or not because we know the # of parameters? May be good to call out behavior if a key isn't there (null)

nicpottier: Do any of our migrations require this? If not, I would vote towards always returning a dict and punting this for now. (until we have the need)

rowanseymour: unfortunately yes (group names, attachment urls) - it's the returning a dict case that we don't currently need... `select` vs `extract` ? Or maybe compile time predictability is fine. i'm trying to convince myself of that right now.

nicpottier: Hah. Maybe they should be more similar than different? 
`extract` and `extract_one`? 
`extract_dict` vs `extract`?

rowanseymour: I see single value being the common case so I'd lean toward `extract` vs `extract_dict`

PR #610, opened 2019-04-01
⌨️ Input as dict in expressions

rowanseymour:


legacy/expressions/context.go
```diff
@@ -294,8 +294,8 @@ func newMigrationVars() map[string]interface{} {
 		"step": &varMapper{
 			substitutions: map[string]interface{}{
 				"contact":     contact,
-				"__default__": "input",
-				"value":       "input",
+				"__default__": "input.text",
+				"value":       `input.text`,
```
rowanseymour: So this isn't quite right for anyone who is checking for attachments, but not sure what to do. In legacy expressions `@step` and `@step.value` are the text and the attachment URLs joined by `\n` Replicating that in an expression would require a crazy expression, and then the operand for every switch router would be a crazy expression. Not sure if the new editor would even recognize those as "wait for response".

So maybe we need to add a key to the new `@input` dict that replicates that?

rowanseymour: Does kinda suck that out story for checking for the presence of attachments is checking if `@step.value` contains "http"

flows/engine/testdata/timeout_test.json
```diff
@@ -41,8 +41,17 @@
                             }
                         ],
                         "default_category_uuid": "78ae8f05-f92e-43b2-a886-406eaea1b8e0",
-                        "operand": "@input",
+                        "operand": "@input.text",
                         "cases": [
+                            {
```
rowanseymour: for now need to make this first test

PR #617, opened 2019-04-03
🔧 Operand of group split should be @contact.groups

rowanseymour:


excellent/functions/builtin.go
```diff
@@ -37,7 +37,7 @@ func init() {
 		"char":              OneNumberFunction(Char),
 		"code":              OneTextFunction(Code),
 		"split":             TwoTextFunction(Split),
-		"join":              ArgCountCheck(2, 2, Join),
+		"join":              TwoArgFunction(Join),
```
rowanseymour: unrelated change.. just remembered there's a better wrapper for a function that takes a fixed number of args

test/testdata/runner/router_tests.test.json
```diff
@@ -13,11 +13,11 @@
                 {
                     "category": "Other",
                     "created_on": "2018-07-06T12:30:11.123456789Z",
-                    "input": "Ben Haggerty",
+                    "input": "[]",
```
rowanseymour: result input is the operand stringified.. need to think about implications of removing operand on switch routers here

PR #618, opened 2019-04-04
⌚ Timeouts as properties of the wait instead of a switch case

rowanseymour:
* Splits `Wait` into `Wait` (the flow definition part) and `ActivatedWait` (the session part)
 * `Wait` gets new timeout object with category UUID
 * Removes `has_wait_timed_out` test

### DIFF
--- a/cmd/flowrunner/testdata/two_questions.json
+++ b/cmd/flowrunner/testdata/two_questions.json
@@ -21,7 +21,10 @@
                         "type": "switch",
                         "wait": {
                             "type": "msg",
-                            "timeout": 600
+                            "timeout": {
+                                "seconds": 600,
+                                "category_uuid": "1024833c-91aa-4873-a3b5-3bac1ef55812"
+                            }
                         },
                         "result_name": "Favorite Color",
                         "categories": [
@@ -39,6 +42,11 @@
                                 "uuid": "78ae8f05-f92e-43b2-a886-406eaea1b8e0",
                                 "name": "Other",
                                 "exit_uuid": "84696f43-07b5-4fde-9991-73d10f8406a5"
+                            },
+                            {
+                                "uuid": "1024833c-91aa-4873-a3b5-3bac1ef55812",
+                                "name": "No Response",
+                                "exit_uuid": "f0649239-6ab2-4903-b5c5-f813beb5539d"
                             }
                         ],
                         "default_category_uuid": "78ae8f05-f92e-43b2-a886-406eaea1b8e0",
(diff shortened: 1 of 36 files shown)


cmd/flowrunner/testdata/two_questions.json
```diff
@@ -94,8 +105,7 @@
                     "router": {
                         "type": "switch",
                         "wait": {
-                            "type": "msg",
-                            "timeout": 600
```
rowanseymour: thanks to copy and pasting flow defs there were lots of cases of example flows with timeouts set but case or exit for them etc

flows/routers/base.go
```diff
@@ -49,6 +50,11 @@ func (r *BaseRouter) Type() string { return r.type_ }
 // Wait returns the optional wait on this router
 func (r *BaseRouter) Wait() flows.Wait { return r.wait }
 
+// AllowTimeout returns whether this router can be resumed at with a timeout
+func (r *BaseRouter) AllowTimeout() bool {
+	return r.wait != nil && !utils.IsNil(r.wait.Timeout())
```
rowanseymour: punting on solving the interface to a nil problem for now with our usual solution

flows/routers/switch.go
```diff
@@ -26,17 +26,15 @@ type Case struct {
 	UUID         utils.UUID         `json:"uuid"                   validate:"required"`
 	Type         string             `json:"type"                   validate:"required"`
 	Arguments    []string           `json:"arguments,omitempty"`
-	OmitOperand  bool               `json:"omit_operand,omitempty"`
```
rowanseymour: this was a workaround for `has_wait_timed_out`

flows/engine/session_test.go
```diff
@@ -278,5 +278,5 @@ func TestWaitTimeout(t *testing.T) {
 	result := run.Results().Get("favorite_color")
 	require.Equal(t, "Timeout", result.Category)
 	require.Equal(t, "2018-04-11T13:34:40.123456Z", result.Value)
-	require.Equal(t, "", *result.Input)
+	require.Nil(t, result.Input)
```
rowanseymour: i had to change this from `nil` when we changed operands to `@input.text` because that errors in the case when input is null

flows/interfaces.go
```diff
@@ -206,22 +206,33 @@ type Router interface {
...
 
-	Timeout() *int
-	TimeoutOn() *time.Time
+	Timeout() Timeout
+
+	Begin(FlowRun, EventCallback) ActivatedWait
+}
 
-	Begin(FlowRun, EventCallback) bool
+type ActivatedWait interface {
```
rowanseymour: I think think we can get rid of this long term and callers use events instead:
  * But we need to go update callers like the simulator
  * We still lean on this in the engine to know if caller is trying to resume with a timeout before timeout_on... but I feel like timeout_on should be advisory to the caller, which would then allow things like the simulator having a "timeout" button that the user can click that any time

PR #619, opened 2019-04-04
🦉 Use relative timeout value in activated wait and msg_wait event

rowanseymour:


flows/engine/session.go
```diff
@@ -184,8 +184,12 @@ func (s *session) tryToResume(sprint flows.Sprint, waitingRun flows.FlowRun, res
 		return err
 	}
 
+	if node.Router() == nil || node.Router().Wait() == nil {
+		return errors.New("can't resume from node without a router or wait")
+	}
+
 	// try to end our wait which will return and log an error if it can't be ended with this resume
-	if err := s.wait.End(resume, node); err != nil {
+	if err := node.Router().Wait().End(resume); err != nil {
```
rowanseymour: this feels a lot cleaner - now we only look at the wait in the definition - not both that and the wait on the session

PR #622, opened 2019-04-05
🤓 Expressions refactor

rowanseymour:
* Convert all remaining things in the context to be an X primitive (#591)
* Dot vs square bracket notation cleanup (#607)
* Rework results so `@results` exposes value/category/category_localized as single values, `@run.results` exposes them as arrays
* Adjust grammar so literals aren't atoms (i.e. can't invoke them as functions, index into them), but expression surrounded by parentheses is, e.g. `@( (contact).name )`

PR #623, opened 2019-04-07
🐶 Context reference migration via regexes

rowanseymour:
Running into complications in #622 and increasingly having to lean on different kinds of substitutions.. got me thinking maybe it would be easier to just use regexes.


antlr/Excellent1.g4
```diff
@@ -1,71 +1,60 @@
 grammar Excellent1;
```
rowanseymour: this is the grammar file from rapidpro/expressions so it should work fine with any legacy expression

legacy/expressions/migrate_test.go
```diff
@@ -91,7 +91,7 @@ func TestMigrateTemplate(t *testing.T) {
 		{old: `@parent.contact.tel.scheme`, new: `@(urn_parts(parent.urns.tel).scheme)`},
 		{old: `@parent.contact.tel.path`, new: `@(urn_parts(parent.urns.tel).path)`},
 		{old: `@parent.contact.tel.urn`, new: `@parent.urns.tel`},
-		{old: `@parent.contact.tel_e164`, new: `@(urn_parts(urns.tel).path)`},
+		{old: `@parent.contact.tel_e164`, new: `@(urn_parts(parent.urns.tel).path)`},
```
rowanseymour: found a bug in old migrations

antlr/Excellent1.g4
```diff
@@ -1,71 +1,60 @@
...
+	| expression AMPERSAND expression					# concatenation
+	| STRING											# stringLiteral
+	| DECIMAL											# decimalLiteral
+	| TRUE												# true
+	| FALSE												# false
+	| NAME												# contextReference
+	| LPAREN expression RPAREN							# parentheses;
 
-atom       : fnname LPAREN parameters? RPAREN             # functionCall
-           | atom DOT atom                                # dotLookup
```
rowanseymour: important difference is getting rid of this so that context references are evaluated as the whole path and not a series of property lookups

rowanseymour:
Ran benchmarks to compare this and the existing resolver based approach and this is marginally faster (~2.3 secs vs ~2.4 secs)

PR #635, opened 2019-04-16
🎨 Format function

rowanseymour:
* Every X type gets a `Format()` method
* Some get an additional `FormatCustom(...)`
* All `format_` functions now lean on that
* Adds `format` which just calls `Format()`

Doesn't include any inspection of text values for other types, but that requires more thought and a different PR. Want to get this part out to leverage in #401.

See https://github.com/nyaruka/goflow/compare/format?expand=1#diff-0541bd492dde63f4df5efafa13ff86fa for what this does with complex objects


docs/functions.json
```diff
@@ -528,20 +547,20 @@
...
-                "template": "@(format_number(31337))",
-                "output": "31,337.00"
+                "template": "@(format_number(1234))",
+                "output": "1,234"
             },
             {
-                "template": "@(format_number(31337, 2))",
-                "output": "31,337.00"
+                "template": "@(format_number(1234.5678))",
+                "output": "1,235"
```
rowanseymour: Have changed `format_number` so it defaults to zero decimal places instead of 2, so that we don't end up adding .00 to everybody's contact field's for age and other whole numbers. `FIXED(x)` in legacy expressions which defaults to 2 places, migrates to `format_number(x, 2)`.

rowanseymour: I guess on further thought this could be weird for people with results that actually are decimals.. like currency.. is that common? Maybe it should be adaptive by default?

nicpottier: Seems like default should be number of significant places?

nicpottier: IE not `0` places. And if you want to force places then you pass in an argument?

rowanseymour: so basically avoid trailing zeros, right? I think that's reasonable.

rowanseymour: updated so now..

```
format_number(1234) -> 1234
format_number(1234.000) -> 1234
format_number(1234.500) -> 1234.5
format_number(1234.567) -> 1234.567
```

nicpottier: Yep, I think that makes the most sense. When you want fixed then you can pass a parameter for that.

PR #651, opened 2019-04-23
💁 Defaults are back

rowanseymour:
Re-adds default values for some things in the context. Things with no default still render like dicts.

| Object           | Default |
|--------------|---------|
|`contact` | name or urn or id |
|`results`| Name: value\n... same as `@flow` in legacy |
|`results.foo` | value |
|`input`| text\nattachments same as `@step.value` in legacy |
|`channel`| name |
|`flow`|  name |
|`test_result`|  true/false |
|`run`|  _none_ |
|`trigger`|  _none_ |
|`fields`| _none_ |
|`urns`|  _none_ |
|`legacy_extra`|  _none_ |

Not sure how I feel about the inconsistency but some things just don't have a natural default.


docs/functions.json
```diff
@@ -511,21 +511,6 @@
             }
         ]
     },
-    {
-        "signature": "format_input(urn)",
```
rowanseymour: this function was definitely weird and annoying that it had to be the operand for all migrated rulesets on @step.value

nicpottier:
Is address the right thing for channel? Perhaps name is more flexible? (think shortcodes where we have more than one channel with the same address)

docs/functions.json
nicpottier: yes, win!

rowanseymour:
@nicpottier `@channel` in legacy world is address... but I have no strong feelings here.. all feels a bit arbitrary... if it is name, maybe same is true for `@flow`...

nicpottier:
Ya, name just feels most natural. Also maps to what we do for contact where name is the highest priority thing.

Issue #654, opened 2019-04-24
Defaults for run and trigger in context


rowanseymour:
We currently have these defaults in the context:

| Object           | Default |
|--------------|---------|
|`contact` | name or urn or id |
|`results`| Name: value\n... same as `@flow` in legacy |
|`results.foo` | value |
|`input`| text\nattachments same as `@step.value` in legacy |
|`channel`| name |
|`flow`|  name |
|`test_result`|  true/false |
|`run`|  _none_ |
|`trigger`|  _none_ |
|`fields`| _none_ |
|`urns`|  _none_ |
|`legacy_extra`|  _none_ |

`fields`, `urns` and `legacy_extra` have no obvious default. But seems like `run` and `trigger` could have them. That would have the added benefit of `run` not dumping out a huge blob unless you explicitly want that i.e. `@(json(run))`

Maybe for run it's a contrived value like `<contact-name>@<flow-name>`.. `Bob@Registration`

nicpottier:
Honestly my leaning is towards no defaults unless we have a specific need and/or the shortcut if obvious. I guess I don't share the same concern over someone including `@run` in a flow, that's going to be pretty obvious pretty quickly and is going to be truncated anyways.

I could see `fields` having a similar default as `results` though.

Happy to be outvoted on this.

rowanseymour:
I'm just trying to establish what the logic is for something having a default and another thing not, as we've spent the previous couple of weeks talking about how great it was that everything in the context now looks and behaves the same.

`@run` dumping out 10s of messages worth of data isn't the end of the world until someone starts building logic around what that looks like (maybe they start posting it to a webhook 😬) and then suddenly we can't change it so easily.

If for argument's sake `@run` defaulted to the UUID, then it would be easy in future to change that and migrate `@run` to `@run.uuid`

So I think I would lean toward reducing the complexity of these defaults where appropriate.

nicpottier:
I actually think the default being simpler and being able to be remapped later is a strong arguments, so I'm on board with more defaults.

PR #656, opened 2019-04-24
🎁 More context defaults

rowanseymour:
Addresses #654

### DIFF
--- a/docs/context.html
+++ b/docs/context.html
@@ -79,7 +79,7 @@
 <a class="sourceLine" id="cb2-7" title="7">@(contact.urns[<span class="dv">0</span>]) → tel:+<span class="dv">12065551212</span></a>
 <a class="sourceLine" id="cb2-8" title="8">@contact.urn → tel:+<span class="dv">12065551212</span></a>
 <a class="sourceLine" id="cb2-9" title="9">@(foreach(contact.groups, extract, <span class="st">&quot;name&quot;</span>)) → [Testers, Males]</a>
-<a class="sourceLine" id="cb2-10" title="10">@contact.fields → {activation_token: AACC55, age: <span class="dv">23</span>, gender: Male, join_date: <span class="dv">2017-12</span>-02T00:<span class="dv">00</span>:<span class="fl">00.000000</span><span class="dv">-02</span>:<span class="dv">00</span>, not_set: }</a>
+<a class="sourceLine" id="cb2-10" title="10">@contact.fields → Activation Token: AACC55\nAge: <span class="dv">23</span>\nGender: Male\nJoin Date: <span class="dv">2017-12</span>-02T00:<span class="dv">00</span>:<span class="fl">00.000000</span><span class="dv">-02</span>:<span class="dv">00</span></a>
 <a class="sourceLine" id="cb2-11" title="11">@contact.fields.activation_token → AACC55</a>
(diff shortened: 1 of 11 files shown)


docs/context.html
```diff
@@ -79,7 +79,7 @@ <h2 id="contact">Contact</h2>
 <a class="sourceLine" id="cb2-7" title="7">@(contact.urns[<span class="dv">0</span>]) → tel:+<span class="dv">12065551212</span></a>
 <a class="sourceLine" id="cb2-8" title="8">@contact.urn → tel:+<span class="dv">12065551212</span></a>
 <a class="sourceLine" id="cb2-9" title="9">@(foreach(contact.groups, extract, <span class="st">&quot;name&quot;</span>)) → [Testers, Males]</a>
-<a class="sourceLine" id="cb2-10" title="10">@contact.fields → {activation_token: AACC55, age: <span class="dv">23</span>, gender: Male, join_date: <span class="dv">2017-12</span>-02T00:<span class="dv">00</span>:<span class="fl">00.000000</span><span class="dv">-02</span>:<span class="dv">00</span>, not_set: }</a>
+<a class="sourceLine" id="cb2-10" title="10">@contact.fields → Activation Token: AACC55\nAge: <span class="dv">23</span>\nGender: Male\nJoin Date: <span class="dv">2017-12</span>-02T00:<span class="dv">00</span>:<span class="fl">00.000000</span><span class="dv">-02</span>:<span class="dv">00</span></a>
```
nicpottier: Huh, so this is going to show all unset fields too ya? Wonder since we have full control of that whether we should exclude those.

nicpottier: Oh hm, that is weird stying.

flows/field.go
```diff
@@ -284,11 +288,21 @@ func (f FieldValues) Parse(env utils.Environment, fields *FieldAssets, field *Fi
...
-		entries[string(k)] = v.ToXValue(env)
+		val := v.ToXValue(env)
+		entries[string(k)] = val
+
+		if v != nil {
+			lines = append(lines, fmt.Sprintf("%s: %s", v.field.Name(), types.Render(val)))
+		}
 	}
+
+	sort.SliceStable(lines, func(i, j int) bool { return strings.Compare(lines[i], lines[j]) < 0 })
```
nicpottier: https://golang.org/pkg/sort/#Strings ?

flows/field_test.go
```diff
@@ -39,8 +39,17 @@ func TestFieldValues(t *testing.T) {
...
 	assert.Nil(t, ageVal.ToXValue(env)) // doesn't have a value in the right type
+
+	test.AssertXEqual(t, types.NewXObject(map[string]types.XValue{
+		"__default__":      types.NewXText("Age: \nGender: Male"),
+		"activation_token": nil,
+		"age":              nil,
+		"gender":           types.NewXText("Male"),
+		"join_date":        nil,
+		"not_set":          nil,
+	}), flows.Context(env, fieldVals))
```
nicpottier: Ok now I'm confused, why is age in `__default__` but not the other nil values?

nicpottier: Oh because the typed value is nil? Seems like we should just be consistent there no? If `@fields.age` is nil then it shouldn't be in `__default__`

rowanseymour: Yeah mistake, updated now

PR #660, opened 2019-04-25
🧞 Add more tests for comparing text using `!=` operator

dodobas:


contactql/parser_test.go
```diff
@@ -94,6 +96,7 @@ func TestEvaluateQuery(t *testing.T) {
 		{`gender = "female"`, false},
 		{`gender != "female"`, true},
 		{`gender != "male"`, false},
+		{`empty != "male"`, true}, // this is true because "" is not "male"
```
rowanseymour: let's also test the reverse: 
```
{`gender != ""`, true},
```

dodobas: added...

PR #664, opened 2019-04-30
🧮 Match number like .5 and 1OO (o's) as 1)

### LINKED ISSUE #657: has_number test fails with values like `.25`
Seems like that should work no?

rowanseymour:
Fixes #657

### DIFF
--- a/flows/routers/cases/tests.go
+++ b/flows/routers/cases/tests.go
@@ -836,8 +836,10 @@ func hasOnlyPhraseTest(origHays []string, hays []string, pins []string) types.XV
 
 // ParseDecimalFuzzy parses a decimal from a string
 func ParseDecimalFuzzy(val string, format *utils.NumberFormat) (decimal.Decimal, error) {
+	cleaned := strings.TrimSpace(val)
+
 	// remove digit grouping symbol
-	cleaned := strings.Replace(val, format.DigitGroupingSymbol, "", -1)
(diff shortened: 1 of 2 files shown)


flows/routers/cases/tests.go
```diff
@@ -849,7 +851,7 @@ type decimalTest func(value decimal.Decimal, test1 decimal.Decimal, test2 decima
 
 func testNumber(env utils.Environment, str types.XText, testNum1 types.XNumber, testNum2 types.XNumber, testFunc decimalTest) types.XValue {
 	// create a number finding regex based on current environment
-	pattern := regexp.MustCompile(fmt.Sprintf(`[-+]?[\pNlO\%s]+(\%s[\pNlO]+)?`, env.NumberFormat().DigitGroupingSymbol, env.NumberFormat().DecimalSymbol))
+	pattern := regexp.MustCompile(fmt.Sprintf(`[-+]?([\pN\%[1]s]+(\%[2]s[\pN]+)?|(\W|^)\%[2]s[\pN]+)`, env.NumberFormat().DigitGroupingSymbol, env.NumberFormat().DecimalSymbol))
```
rowanseymour: note removal of `lO` (lowercase L, uppercase O) here - we ripped out doing substitutions of these but I evidentally forgot to remove them from this regex, meaning `1OO` would match here but then fail in `ParseDecimalFuzzy`. Now we match that as just `1`

Issue #663, opened 2019-04-29
Awkward to get at webhook body


nicpottier:
So we currently only include `extra` on the full context for results, meaning you have to do something like `@run.results.webhook.extra.response[0]` to get at something vs `@results.webhook.extra.response[0]`

Do we want that to be easier? I suppose the argument against having that in `@results` is that these are the same simplified results we see on `@parent` and `@child` so being consistent on what is available there makes sense. (I agree with that)

Should be have something similar to our old `@extra` though that is just the last webhook body? IE: `@webhook`? That would shorten the above to just `@webhook.response[0]` which sure seems more reasonable.

nicpottier:
@ericnewcomer

rowanseymour:
It should definitely be on `@results` the same as `@run.results` - the reason I punted on adding it there previously is because that would change the default webhook payload which is using `json(results)`, so we have to decide if 1) we want potentially big webhook payloads (payloads in payloads!) appearing there 2) we want a func like `extract_dict` that excludes a named field.

Re having a `@webhook.response[0]` - I don't love treating webhooks as a special case here since extra can come from various places, so if we did go that route I think I'd lean toward `@extra.foo` where `foo` is the last extra. But I don't hate `@results.webhook.extra.foo`. `@results.webhook` is the natural place to look for that since that's where everything else relating to that webhook result is.

nicpottier:
I though we already had an `extract`, did we lose that? It feels slightly nutty to be able to access the webhook results of your parents / children but I guess that's all really just captured in the giant session blob anyways so 🤷‍♂ 

To be clear I wasn't proposing putting all extras in `@webhook` just the webhook ones. I could go either way as to whether resthooks populate that though likely would lean against it.

It just seems to be that this is another case where we are making things quite a bit more awkward for users who used to just be able to do `@extra.foo` and now need to do `@results.my_webhook.extra.foo` (if we did add it to `@results`).

Calling out to external APIs is a pretty big thing in RapidPro, I'm not bothered by that being a special / optimized case.

I guess I could be convinced to have an `@extra` that basically is always the last `@extra` that is written. Would that get cleared when a result is written with no `extra`? Nice thing there would be that in most cases people wouldn't have to relearn how `@extra` works and that we'd be consistent in the naming of those attributes in `@run.results` and elsewhere.

nicpottier:
@nyaruka/nyaruka thoughts?

rowanseymour:
[...]

PR #669, opened 2019-05-02
🎣 Add @webhook as shortcut to .extra of last webhook result

rowanseymour:
Addresses #663

### DIFF
--- a/flows/runs/run.go
+++ b/flows/runs/run.go
@@ -206,12 +206,25 @@ func (r *flowRun) RootContext(env utils.Environment) map[string]types.XValue {
 		"results": flows.ContextFunc(env, r.Results().SimpleContext),
 		"urns":    urns,
 		"fields":  fields,
+		"webhook": r.lastWebhookResponse(),
 
 		// other
 		"trigger":      flows.Context(env, r.Session().Trigger()),
 		"input":        flows.Context(env, r.Session().Input()),
 		"legacy_extra": r.legacyExtra.ToXValue(env),
 	}
+}
+
+func (r *flowRun) lastWebhookResponse() types.XValue {
+	for e := len(r.events) - 1; e >= 0; e-- {
+		switch typed := r.events[e].(type) {
+		case *events.WebhookCalledEvent:
+			return types.JSONToXValue(flows.ExtractResponseBody(typed.Response))
+		default:
+			continue
+		}
+	}
+	return nil
 }
 
 // Context returns the properties available in expressions
--- a/flows/actions/base.go
+++ b/flows/actions/base.go
@@ -246,18 +246,7 @@ func (a *BaseAction) saveWebhookResult(run flows.FlowRun, step flows.Step, name 
 	input := fmt.Sprintf("%s %s", webhook.Method(), webhook.URL())
 	value := strconv.Itoa(webhook.StatusCode())
 	category := webhookStatusCategories[webhook.Status()]
-
-	body := []byte(webhook.Body())
-	var extra json.RawMessage
-
-	// try to parse body as JSON
-	if utils.IsValidJSON(body) {
-		// if that was successful, the body is valid JSON and extra is the body
-		extra = body
-	} else {
-		// if not, treat body as text and encode as a JSON string
-		extra, _ = json.Marshal(string(body))
-	}
+	extra := flows.ExtractResponseBody(webhook.Response())
 
 	a.saveResult(run, step, name, value, category, "", input, extra, logEvent)
 }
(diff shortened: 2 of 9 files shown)


flows/runs/run.go
```diff
@@ -214,6 +215,18 @@ func (r *flowRun) RootContext(env utils.Environment) map[string]types.XValue {
 	}
 }
 
+func (r *flowRun) lastWebhookResponse() types.XValue {
+	for e := len(r.events) - 1; e >= 0; e-- {
```
nicpottier: minor obviously but `e` is an odd choice for index no?

rowanseymour: Interesting, I guess go's convention of calling the `this` variable a single letter has taken me down a path of using the first letter of thing being looped, rather than the traditiona `i`, `j`, `k`,.. so for example:

```go
func (i *Input) foo() {
  s, session := range i.sessions {
     e, event := range s.events {
```

nicpottier: `i` in sacrosanct in my world, but ya maybe I'm wrong. Perhaps worth investigating what the stdlib does.

nicpottier: Perhaps just the strings library, but first place I looked seems to agree:
  https://golang.org/src/strings/strings.go?s=9816:9869#L373

ericnewcomer: Iterating a list of objects is one thing, but a counter? oof.

rowanseymour: I see some `n` and `p` but yeah `i` is most common. I don't mind sticking to `i`'s but if stdlib is the authority then you can't get upset when I do `if err := foo(); err != nil` which is all over the stdlib 😸

nicpottier: touche

nicpottier: I don't see any `n` or `p` loop vars though, where? Just out of curiosity.

nicpottier: And on the `if` construct, it isn't that I disagree that that is most common it is just that I am too dumb to parse it. It literally takes me a minute to figure out what is happening. I guess I'm just not smart enough for that cleverness is all, so I find the likelihood of me writing a bug that way much higher that if I do the simpler construct.

rowanseymour: Here's some `i`, `n`, `x` and `y` all together https://github.com/golang/go/blob/50bd1c4d4eb4fac8ddeb5f063c099daccfb71b26/src/runtime/memmove_test.go#L31

You can certainly make some ugly `if` statements but the exact same thing is true of `for` loops. I'm not sure `if err := foo(); err != nil` really is so much harder to parse than `for i := foo(); i < something;`

nicpottier: Those are while loops, but noted.

Hey, not arguing the construct, maybe I'll get used to it, just saying for me personally as much as I would love to be able to parse them no problem, I just can't. It isn't a style thing for me at all, its a not smart enough thing.

nicpottier: Apparently I can't tell what a while loop is either, nm that previous comment :)

rowanseymour: I don't think anyone buys you not being smart enough but anyways. We should write a style guide one day. 

One thing I'm noticing just flicking thru stdlib is they really love their single char var names.. not just for counters etc

nicpottier: I don't know what to say, I'm being 100% honest there, I just find it super hard to parse that construct. Given how much code we read, I just think it's best to choose the easier to read construct, if others find it easy to read and obvious all the more reason. Like I said, maybe I'd get used to it with time.

rowanseymour: Yeah but that has nothing to do with "smartness" so I'm not sure how helpful it is to keep claming that. It's simple that you don't use that construct so you don't recognize it as easily.

nicpottier: Perhaps what I'm trying to say is that if I don't find it super easy then maybe others won't either. Totally willing to acknowledge that may change with use. To me it just isn't worth the downsides to save one line.

Super happy to have the team discuss and be outvoted.

Did you merge without changing the `i`? I thought we came to agreement there.

rowanseymour: I was going to fix the loop variables afterwards. That style of using the first letter of the thing being enumerated is everywhere in goflow.

nicpottier: Ah ok.

rowanseymour: Btw looks like for maps `k` (key) and `v` (value) is the norm, e.g. https://github.com/golang/go/blob/master/src/archive/tar/common.go#L457

PR #671, opened 2019-05-03
properly return template dependencies for flows

nicpottier:
Wasn't being included in the dependencies json as they are included by type.

### DIFF
--- a/flows/definition/dependencies.go
+++ b/flows/definition/dependencies.go
@@ -6,12 +6,13 @@ import (
 )
 
 type dependencies struct {
-	Channels []*assets.ChannelReference `json:"channels,omitempty"`
-	Contacts []*flows.ContactReference  `json:"contacts,omitempty"`
-	Fields   []*assets.FieldReference   `json:"fields,omitempty"`
-	Flows    []*assets.FlowReference    `json:"flows,omitempty"`
-	Groups   []*assets.GroupReference   `json:"groups,omitempty"`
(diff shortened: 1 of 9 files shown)


flows/definition/dependencies.go
```diff
@@ -6,12 +6,13 @@ import (
...
-	Flows    []*assets.FlowReference    `json:"flows,omitempty"`
-	Groups   []*assets.GroupReference   `json:"groups,omitempty"`
-	Labels   []*assets.LabelReference   `json:"labels,omitempty"`
+	Channels  []*assets.ChannelReference  `json:"channels,omitempty"`
+	Contacts  []*flows.ContactReference   `json:"contacts,omitempty"`
+	Fields    []*assets.FieldReference    `json:"fields,omitempty"`
+	Flows     []*assets.FlowReference     `json:"flows,omitempty"`
+	Groups    []*assets.GroupReference    `json:"groups,omitempty"`
+	Labels    []*assets.LabelReference    `json:"labels,omitempty"`
+	Templates []*assets.TemplateReference `json:"templates,omitempty"`
```
nicpottier: It feels like a shame that references aren't self typing in JSON so that we could avoid this. Maybe worth considering at some point?

rowanseymour: you mean a single merged list and they have a type field, or a dynamic way of building these lists? The latter doesn't sound impossible

nicpottier: The former.

Issue #674, opened 2019-05-09
Text should be required for send_msg and send_broadcast


rowanseymour:
However, see #673

Issue #673, opened 2019-05-09
Struct-level vs semantic validation


rowanseymour:
There are parts of the flow spec that we lean on struct validation with `validate:".."` tags for validation, e.g. `uuid` is required, `node.router.categories` must have at least 1 value. This kind of validation is invoked within `definition.ReadFlow` so you can't actually read a flow that violates one of those constraints. The mailroom validation endpoint will return 400 (bad request) for such a flow definition.

There are other parts where we lean on custom `.Validate` methods on flow objects, e.g. the `remove_contact_groups` action can't both specify `all_groups=true` and have specific groups in `groups`. This kind of flow validation is invoked from `Flow.Validate` and can include checking that dependencies exist for a given org. The mailroom validation endpoint will return 400 for such a flow definition. The mailroom validation endpoint will return 422 (unprocessable entity) for such a flow definition.

Yet from the pov of caller like RapidPro trying to determine if a flow definition is valid... those concepts of validity aren't so different.

So maybe we should defer doing struct level validation until `Flow.Validate` ?  Or leave things as they are and just make mailroom validation return 422 if it can't read the flow?

rowanseymour:
Having thought about this a bit more I think..

* `definition.ReadFlow` should always return a "valid" flow definition _in the current engine version_, i.e. you can pass it v13.x flow JSON and get back a 14.y flow if that's the latest version - you don't have explicitly migrate the flow.
* The returned flow is "valid" in the sense that it is guaranteed to be runnable so this includes tag-based struct-level validation as well as higher-level object validation (i.e. a node's exit's point to actual nodes)
* Checking of dependencies (against session assets) and augmenting the definition with inspection results should be a separate operation - maybe called something different to `Validate`

@nicpottier @ericnewcomer ?

nicpottier:
Ya that all seems reasonable to me

PR #677, opened 2019-05-10
🔬 Better validation

rowanseymour:
* Structural validation now occurs within `definition.ReadFlow` - this includes tag based stuff as well as custom .Validate methods on flow entities
* Dependency checking, result extraction is a separate thing, methods for that renamed to `Inspection*` for clarity

Addresses #673 and #674


flows/interfaces.go
```diff
@@ -171,9 +171,6 @@ type Node interface {
 	Router() Router
 	Exits() []Exit
 
-	AddAction(Action)
-	SetRouter(Router)
```
rowanseymour: never liked having these.. assets should be immutable

flows/definition/dependencies.go
```diff
@@ -59,10 +60,11 @@ func (d *dependencies) refresh(sa flows.SessionAssets, missing assets.MissingCal
 	for i, ref := range d.Flows {
 		a, err := sa.Flows().Get(ref.UUID)
 		if err != nil {
-			missing(ref)
-		} else {
-			d.Flows[i] = a.Reference()
+			// flows are the one thing that aren't allowed to be missing as we wouldn't know how to
+			// route from a subflow split with a missing flow
+			return errors.Wrapf(err, "unable to read %s", ref)
```
rowanseymour: this means we won't start a flow A which contains a subflow B if
* B doesn't exist
* B is unreadable (missing non-flow assets within flow B still ok)

@nicpottier

flows/definition/flow.go
```diff
@@ -32,18 +32,18 @@ type flow struct {
 	// optional properties not used by engine itself
 	ui json.RawMessage
 
-	// properties set after validation
+	// properties set after inspection
```
rowanseymour: feeling less sure about this approach of adding unofficial fields to the flow definition, especially now that in RP we save these off to Flow.metadata.. maybe the validate endpoint should just return this stuff separately

rowanseymour: U-Report will use this information but also not from the definition directly - instead we'll expose parts of `Flow.metadata` on the flows.json endpoint

flows/definition/dependencies.go
nicpottier: Having references to flows that don't exist is a thing right now that isn't super uncommon. I think we currently run these just fail when we reach that node. I think we should probably still do the same or we are going to get into trouble.

flows/definition/flow.go
nicpottier: Other question may be what this means from an import / export perspective. IE, do we include metadata in those exports? (I think there was some thought at some point that the exports may contain additional information such as the type of contact fields we depend on so that we can recreate them appropriately, but that may be separate)

rowanseymour: Hmm contact field types is an interesting one - thus far everything we've called metadata is stuff that we can re-calculate from the flow definition - so there's no reason to trust incoming metadata when the goflow validate call is going to return refreshed metadata. If we did add something like contact field types then it would be more analogous to `_ui` in that the engine doesn't want/need it but RapidPro does.

So I think that's maybe (?) an argument for treating these re-calculable things ( `_results`/`_waiting_exits`/`_dependencies` as something different which are not kept in the flow definition and are not included in exports. They're just a thing you get back from mailroom/validate and RP can put them wherever it wants. 

I guess the thing that's irking me most is having a growing number of top-level "unofficial" parts of the spec with no clear demarcation of who owns and needs what. `_ui` is nice because the editor owns that and can work with any flow that has just that.

flows/definition/dependencies.go
rowanseymour: Just checked the old engine and it's behavior in this scenario looks a bit undefined, but also it's looking up the subflow without filtering by `is_archived` or `is_active` so it was way less likely to run into this scenario. The new engine currently lets the flow pass validation when running (it's treated like any other missing asset) - but then hard errors (i.e. `Resume` returns `error`) if a `EnterFlow` action tries to load it and can't.

Agree best to be forgiving for now so I think we can have a missing subflow result in a fatal error event.

rowanseymour: Updated so that flows treated like any other missing asset at inspection time - and at run time we generate a fatal error if we can't get the subflow.

flows/definition/flow.go
rowanseymour: Since this PR doesn't change that stuff, am punting this conversation to a new issue #679

PR #678, opened 2019-05-13
🤕 Simplify the expression used to emulate legacy webhook payloads

rowanseymour:


docs/md/flows.md
```diff
@@ -214,16 +214,47 @@ through `extra` on the result.
         "step_uuid": "644592ee-11ad-4bc4-9566-6fb2598c32d6",
         "resthook": "new-registration",
         "payload": {
+            "channel": {
```
rowanseymour: a lot of these tests are sensitive to whitespace or ordering changes so i added a new one in master which isn't and you can see it isn't affected by this change https://github.com/nyaruka/goflow/pull/678/files#diff-20e5eb1c65c375c018ca6cb5bb957efcL147

flows/webhook.go
```diff
@@ -15,16 +15,28 @@ import (
...
-	"flow": @(json(run.flow)),
-	"path": @(json(run.path)),
-	"results": @(json(results)),
-	"run": {"uuid": "@run.uuid", "created_on": "@run.created_on"},
-	"input": @(json(if(input, object("attachments", foreach(input.attachments, attachment_parts), "channel", input.channel, "created_on", input.created_on, "text", input.text, "type", input.type, "urn", if(input.urn, object("display", default(format_urn(input.urn), ""), "path", urn_parts(input.urn).path, "scheme", urn_parts(input.urn).scheme), null), "uuid", input.uuid), null))),
-	"channel": @(json(if(input, input.channel, null)))
-}`
+// LegacyWebhookPayload is a template that matches the JSON payload sent by legacy webhooks
+var LegacyWebhookPayload = `@(json(object(
+	"contact", object("uuid", contact.uuid, "name", contact.name, "urn", contact.urn),
```
rowanseymour: I've simplified 

```
if(
  default(input.urn, default(contact.urns[0], null)), 
  text(default(input.urn, default(contact.urns[0], null))), 
  null
)
``` 

to just `contact.urn`... I think we needed the previously more complex expression because `contact.urn` was formatted and I can't think of any realistic scenario where these would be different.

flows/webhook.go
```diff
@@ -15,16 +15,28 @@ import (
...
+			"created_on", input.created_on,
+			"text", input.text,
+			"type", input.type,
+			"urn", if(input.urn, object("display", default(format_urn(input.urn), ""), "path", urn_parts(input.urn).path, "scheme", urn_parts(input.urn).scheme), null),
+			"uuid", input.uuid
+		),
+		null
+	),
+	"channel", default(input.channel, null)
+)))`
```
ericnewcomer: Would definitely prefer this stretched vertically instead of horizontally. Also, feel like `if` is just easier to grok when on different lines. What about 2 space indent?

rowanseymour: Took more new lines and changed indentation to 2 spaces.. which I find a little harder to read but only place people will see this is the editor so your call

Issue #679, opened 2019-05-13
Re-think "unofficial" data in the flow definition?


rowanseymour:
Currently we have 4 "unofficial" top-level objects in a flow definition:

```
_ui
_dependencies
_results
_waiting_exit_uuids
```

`_ui` is ignored by the engine and owned/maintained entirely by the editor. It's simple to look at a flow and if it has `_ui` we know it works with the editor. So not suggesting any change to that cuz that seems like an example of doing this right.

The others are set after inspection/validation so may or may not exist, and are read by RapidPro and saved to `Flow.metadata`. If we were importing a flow we'd ignore them since they are re-calculated by validation. So that raises the question of should they even be in the flow definition. They could just be returned by mailroom's validate endpoint and saved directly to `Flow.metadata`.

Also them being individual top-level items doesn't feel so scalable as the definition top-level will feel pretty cluttered if we add more.

PR #680, opened 2019-05-16
🕵️‍♂️ (Re)allow inspecting without session assets

rowanseymour:
Forgot when I made #677 that we sometimes need to inspect a flow (get it's dependencies) without actually checking they exist

### DIFF
--- a/flows/definition/flow.go
+++ b/flows/definition/flow.go
@@ -38,8 +38,7 @@ type flow struct {
 	waitingExits []flows.ExitUUID
 
 	// internal state
-	nodeMap   map[flows.NodeUUID]flows.Node
-	inspected bool
+	nodeMap map[flows.NodeUUID]flows.Node
 }
 
 // NewFlow creates a new flow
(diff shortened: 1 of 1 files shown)


flows/definition/flow.go
```diff
@@ -102,68 +101,72 @@ func (f *flow) validate() error {
...
 
 type brokenDependency struct {
 	dependency assets.Reference
 	reason     error
 }
 
-func (f *flow) doInspect(sa flows.SessionAssets, recursive bool, missing func(assets.Reference)) error {
-	// if this flow has already been inspected, don't need to do it again - avoid unnecessary work
-	// but also prevents looping if recursively inspecting flows
-	if f.inspected {
```
rowanseymour: need to be able to `Inspect` the same flow more than once - with and without assets

flows/definition/flow.go
```diff
@@ -102,68 +101,72 @@ func (f *flow) validate() error {
...
+					}
+				}
+				return errors.Errorf("missing dependencies: %s", strings.Join(depStrings, ","))
 			}
-			return errors.Errorf("missing dependencies: %s", strings.Join(depStrings, ","))
 		}
 	}
 
-	f.inspected = true
 	f.dependencies = deps
```
rowanseymour: really starting to feel like these things should be just a return value of `Inspect` and not embedded in the flow as per #679

PR #683, opened 2019-05-22
👬 Add support for cloning flows

rowanseymour:


assets/references.go
```diff
@@ -22,6 +22,12 @@ type Reference interface {
 	Variable() bool
 }
 
+// UUIDReference is interface for all reference types that contain a UUID
+type UUIDReference interface {
+	Reference
+	GenericUUID() utils.UUID
```
rowanseymour: having different types for every kind of UUID is mostly good but sometimes hurts

flows/definition/flow.go
```diff
@@ -280,6 +280,43 @@ func (f *flow) ExtractExitsFromWaits() []flows.ExitUUID {
 	return exitUUIDs
 }
 
+// Generic returns this flow's data modelled as a hierarchy of generic maps and slices
+func (f *flow) Generic() map[string]interface{} {
+	marshaled, err := json.Marshal(f)
+	if err != nil {
+		panic(err.Error())
```
rowanseymour: there's really no way this can happen.. thus handled as a panic

PR #684, opened 2019-05-23
🎊 Split up flow inspection and dependency validation

rowanseymour:
..and don't embed inspection results in the flow definition. Addresses #679.

So now we have these flow operations and hopefully slighter clearer terminology:

- **Read**: deserializes a flow from JSON and errors if it isn't structurally valid
- **Validate**: extracts dependencies from a flow, checking if they, exist, and then recursively doing the same for any flow dependencies
- **Inspect**: extracts results/dependencies/waitingexits as a separate thing, can't error


flows/actions/base.go
```diff
@@ -67,7 +67,7 @@ func (a *BaseAction) EnumerateDependencies(localization flows.Localization, incl
 }
 
 // EnumerateResults enumerates all potential results on this object
-func (a *BaseAction) EnumerateResults(include func(*flows.ResultSpec)) {}
+func (a *BaseAction) EnumerateResults(include func(*flows.ResultInfo)) {}
```
rowanseymour: renamed for consistency with `FlowInfo`

flows/inspect.go
```diff
@@ -9,6 +9,82 @@ import (
 	"github.com/nyaruka/goflow/utils"
 )
 
+type FlowInfo struct {
+	Dependencies *Dependencies `json:"_dependencies"`
+	Results      []*ResultInfo `json:"_results"`
+	WaitingExits []ExitUUID    `json:"_waiting_exits"`
+}
```
nicpottier: Where is this object used exactly? Should the default json keys here not have the leading `_` since that is handled in the flow envelope case?

rowanseymour: Ah yes.. probably should have written a test for that! This is what `/mr/flow/inspect` would return. We can add whatever we want in here.

Issue #702, opened 2019-06-18
`@results.foo.extra` isn't a thing


nicpottier:
`extra` only lives on `run.results`, which seems a bit lame. Do these need to be different really? (now that we have `__default__`?)

nicpottier:
(note this is also what was giving Jason trouble, he COULD access the parent / child extra via `parent.run.results.extra.foo`, I thought we didn't pass that through)

rowanseymour:
Schema of `@results`, `@child.results` and `@parent.results` is currently...

- `__default__` => value
- name
- value
- category
- category_localized
- input
- node_uuid
- created_on

whereas schema of `@run.results`,  `@child.run.results`, `@parent.run.results` is currently...

- `__default__` => value
- name
- values
- categories
- categories_localized
- input
- extra
- node_uuid
- created_on

I do feel like those being subtly different is always going to cause some confusion.. two options:

 1.  Just add `@extra` and for now always steer users to `@results`, `@child.results` and `@parent.results`
 2. Merge these to give them the same mega schema..

nicpottier:
My vote is just mega schema it.. having them be subtly different is obviously so confusing that even we are having trouble keeping them straight. We could have auto-complete not advertise the plural cases in the editor to keep some of the simplicity.

nicpottier:
@ericnewcomer ?

ericnewcomer:
I think mega schema, so `value` and `values` are siblings. I think we probably surface all of it though. Would add to the confusion to just omit values if it is a valid thing. We have descriptions right there at least. But ya, can play with how we display it.

I'm all for one MEGASCHEMA to rule them all.

rowanseymour:
Addressed in #708

PR #708, opened 2019-06-19
👯 Make @results and @run.results the same

rowanseymour:
Addresses #702


flows/webhook_test.go
```diff
@@ -227,30 +227,42 @@ func TestLegacyWebhookPayload(t *testing.T) {
 		"results": {
 			"2factor": {
 				"category": "",
+				"categories": [""],
```
rowanseymour: I looked at trying to keep the legacy webhook payload the same but we don't have the right functions in excellent to pull it off - `foreach` only works on arrays, not objects like `@results`

So... maybe we just accept a certain amount of duplication here 🤷‍♂. We're only adding fields so this shouldn't break anyone's stuff.

nicpottier: I can't believe you didn't add foreach support for dicts! But ya, I'm ok with this.

Issue #703, opened 2019-06-18
Should `parent` and `child` be anything more than the run?`


nicpottier:
Currently only thing that seems useful is our treatment of `urns` at the root level.

rowanseymour:
FYI the last discussion which unfortunately we had offline https://github.com/nyaruka/goflow/issues/627

rowanseymour:
To aid discussion, here's a simplified map of run stuff in the context currently:

- run
  - uuid
  - contact
  - flow
  - status
  - results
  - path
  - created_on
  - exited_on
- child
  - run
    - uuid
    - contact
    - flow
    - status
    - results
  - contact (shortcut to child.run.contact)
  - urns (simplified urn access for child.run.contact)
  - fields (shortcut to child.run.contact.fields)
  - results (shortcut to child.run.results)
- parent (same schema as child)
- contact (shortcut to run.contact)
- results (shortcut to run.results)
- urns (simplified urn access for run.contact)
- fields (shortcut to run.contact.fields)
- webhook
- trigger
- input
- legacy_extra

rowanseymour:
After discussion with @nicpottier, we could "deprecate" `parent.run.foo` and `child.run.foo` for now, make those fields accessible at `parent.foo` and `child.foo`, and remove them from the auto-complete. Then we need to write a migration all `(parent|child).run.*` references.

Thus we have...

- run
  - uuid
  - contact
  - flow
  - status
  - results
  - path
  - created_on
  - exited_on
- child
  - run
    - uuid (to be migrated to child.uuid)
    - contact (to be migrated to child.contact)
    - flow (to be migrated to child.flow)
    - status (to be migrated to child.status)
    - results (to be migrated to child.results)
  - uuid (new!)
  - flow (new!)
  - status (new!)
  - contact
  - urns (simplified urn access for child.contact)
  - fields (shortcut to child.contact.fields)
  - results
- parent (same schema as child)
- contact (shortcut to run.contact)
- results (shortcut to run.results)
- urns (simplified urn access for run.contact)
- fields (shortcut to run.contact.fields)
- webhook
- trigger
- input
- legacy_extra

I'm old enough to remember when this was what the context looked like but then we made `results` and `run.results` different and needed an equivalent for `parent` and `child`

PR #710, opened 2019-06-20
👩‍👦 Deprecate parent.run and child.run in the context

rowanseymour:
And move those fields onto parent and child themselves. Step 1 of #703. Next steps:

1. Update editor auto-complete and how it constructs subflow routers
2. Write migration for 13.1 flows which rewrites `(parent|child).run.*` context references


ericnewcomer:
ace emoji work

Issue #715, opened 2019-06-28
Expose external_id of incoming message in the flow


erikh360:
We'd like to have access to the `external_id` of an incoming message as a variable in the new flow designer. We need it to initiate actions on the incoming message in external apps.

For example if a user needs help, we can label it in the external app and action accordingly:
POST `v1/messages/<external-id>/labels`
{
    "labels": ["needs_help"]
}

I see external_id was already added to the MsgIn object in `goflow` here: https://github.com/nyaruka/goflow/pull/437

nicpottier:
I'm not opposed to this, obviously it may be `null` but seems we can expose it when available and can definitely see the use case. What do you think @rowanseymour ?

rowanseymour:
Yeah I'll make a PR for this shortly

erikh360:
Thanks! @rowanseymour @nicpottier

miltontony:
@nicpottier @rowanseymour appreciate the quick turnaround on this. When is this likely to be released and deployed to Textit.in?

rowanseymour:
This is deployed now on the engine side - but `@input.external_id` won't appear in auto-complete until we update editor https://github.com/nyaruka/floweditor/issues/665

PR #716, opened 2019-07-01
🏡 Include external ID of msg on input expression context

rowanseymour:
Addresses #715


flows/inputs/msg.go
```diff
@@ -90,6 +94,7 @@ type msgInputEnvelope struct {
 	URN         urns.URN           `json:"urn" validate:"omitempty,urn"`
 	Text        string             `json:"text"`
 	Attachments []utils.Attachment `json:"attachments,omitempty"`
+	ExternalID  string             `json:"external_id,omitempty"`
```
nicpottier: Should this be a `null.String` maybe? Just seems like differentiating null from empty may make sense in this case.

rowanseymour: We've thus far avoided nullable strings in the engine and I can't imagine a case where that needs to be differentiated in expressions.

nicpottier: No reason to be afraid of nullable strings but fair point that expressions probably don't care about the difference between the two.

PR #719, opened 2019-07-02
tweak to goflow interfaces to allow introspection into contactql

nicpottier:
Just adding some getters so that mailroom can get at what it needs.

### DIFF
--- a/contactql/parser.go
+++ b/contactql/parser.go
@@ -14,17 +14,20 @@ import (
 	"github.com/shopspring/decimal"
 )
 
-type boolOp string
+// BoolOperator is a boolean operator (and or or)
+type BoolOperator string
 
 const (
-	boolOpAnd boolOp = "and"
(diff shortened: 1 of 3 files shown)


contactql/parser.go
```diff
@@ -14,17 +14,20 @@ import (
...
+type BoolOperator string
 
 const (
-	boolOpAnd boolOp = "and"
-	boolOpOr  boolOp = "or"
+	// BoolOperatorAnd is our constant for an AND operation
+	BoolOperatorAnd BoolOperator = "and"
+
+	// BoolOperatorOr is our constant for an OR operation
+	BoolOperatorOr BoolOperator = "or"
```
nicpottier: FYI, only renamed this because it seemed weird having just an `Op()` accessor or having an `Operator()` accessor return a `BoolOp`. Otherwise would have left it.

PR #720, opened 2019-07-05
📱 Allow setting channel on a non-tel URN if it doesn't have a channel

rowanseymour:


rowanseymour:
How about lazy lazy lazy channel resolution by courier where it tries the first channel with the right scheme, if that fails the next channel... 🤯

nicpottier:
Problem with that is knowing what type of failure is taking place, lots of false negatives due to network, throttling, etc.. But ya, maybe there's a more structured approach that we could take hitting APIs for some of these channels that support them. (someday)

PR #721, opened 2019-07-08
🗺️ Map of the context

rowanseymour:


rowanseymour:
@ericnewcomer @nicpottier if we're ok with the general structure of this I'd like to merge this and then we can tweak and re-order the docstrings as needed later

PR #726, opened 2019-07-12
🧪 Reflection based template and dependency extraction

rowanseymour:
Having a second stab at this as it would also lend itself to extracting the "paths" of the templates so for expression migrations - we could have a set of paths that represent all the locations of templates in a particular version of the flow spec.

Those paths would be using the JSON names of fields which come from the `json:` tags on fields so tags seem like the obvious direction to go there.


rowanseymour:
@nicpottier so this doesn't include the whole template path extraction part that we'd need for migrations but I figure this is a good point to pause, take a step back and discuss if this is the direction we want to go. This is just replacing the existing inspection stuff at the level of actions with new stuff that uses struct tags.

Obviously there's some complexity in this PR but the important question is whether this is more or less complicated than the code it's replacing. I _think_ it is but I also don't want to merge this just because I've spent 2 days on it.

flows/actions/send_email.go
```diff
@@ -109,11 +110,6 @@ func (a *SendEmailAction) Inspect(inspect func(flows.Inspectable)) {
...
 
 // EnumerateTemplates enumerates all expressions on this object and its children
-func (a *SendEmailAction) EnumerateTemplates(include flows.TemplateIncluder) {
-	include.String(a.Subject)
-	include.String(a.Body)
-	include.Slice(a.Addresses)
-
-	include.Translations(a, "subject")
-	include.Translations(a, "body")
+func (a *SendEmailAction) EnumerateTemplates(localization flows.Localization, include func(string)) {
```
nicpottier: Can we make this a mixin so we don't have all this repetition?

rowanseymour: Alas no because a golang "mixin" only knows about itself.. it can't point to the wider struct that contains it.

Right now the `Inspect()` stuff is going as deep as the actions, and then the new reflection stuff goes into the actions themselves. We need all this boilerplate because we need the reflection stuff to start on the actual struct itself and not just an embedded part of each action struct.

But if we got rid of the old inspection stuff all together we wouldn't need this because from the node, you can see the the actual action stucts...

nicpottier: There must be some other way no? What if the mixin instance is constructed with the outer struct reference? Makes construction more complicated.

rowanseymour: You could make something like that work, but as you say it makes construction more complicated, and wouldn't it be good to use the same reflection-based system for the whole flow tree?

nicpottier: Ok, maybe I'm not understanding that right. Just smells to have all that repetition and seems prone to implementations drifting in ways that break things. (there would be constructor repetition yes, but implementation could be done in one place that would affect all) Up to you.

rowanseymour: I'm not arguing against removing the repetition - I'm arguing for getting rid of it by making nodes use the same reflection-based inspection. That feels cleaner and saner.

Issue #478, opened 2019-01-09
Add support for flow spec migrations


rowanseymour:
Because 12.0 is unlikely to be the last flow version...

rowanseymour:
The simplest version of this would be like how we do it in Python currently - the old flow is loaded into generic map of maps, and that's run through a series of transformation functions, e.g.

```go
func migrateTo_12_1(jsonFlow map[string]interface{}) (map[string]interface{}, error) {...}
```

However, working with `map[string]interface{}` in go isn't anywhere as easy as working with a dict in Python. Pondering on ways to use reflection to create a "snapshot of structs" of each flow version.

Would also be great if there was an easy way to visit all fields in a definition which are excellent templates. Maybe that could be achieved with struct tags and would save any migration which needs to rewrite expressions from having to know which fields can contain expressions.

nicpottier:
Definitely agree that having an expression shortcut migration would be a neat thing.

Ya, go is pretty shit at this kind of random map manipulation. I think the other bit there is that many past migrations have actually required us to do db lookups which then gets us squarely out of the goflow world. (though maybe that's captured in assets actually)

Maybe `jsonparser` is good enough for a while with some utility methods to let us more easily define 'ActionMigrator' or `SwitchMigrator` or whatnot. (you specify a list of migrators and it takes care of walking the tree for you and calling your migrator)

To be clear I don't think it necessarily makes sense to start on this until we have a real migration, we just want to make sure we have everything we need in place such that if we DID need a migration we could do it without too much work. (so just the stub for 'ensure_current_version' equiv?)

PR #729, opened 2019-07-19
🚚 First flow migration: adding UUID to send_msg templating

rowanseymour:
This PR is three things:

1. Adding a framework for doing migrations (#478)
2. Using it to define 13.1 of the flow spec that adds UUID to `templating` on `send_msg` objects
3. Using that to implement localization of templating variables on `send_msg` actions


flows/definition/migrations/13_1.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/Masterminds/semver"
+)
+
+func init() {
+	registerMigration(semver.MustParse("13.1"), AsParsed(Migrate13_1))
+}
+
+// Migrate13_1 adds UUID to send_msg templating
+func Migrate13_1(f Flow) Flow {
+	for _, node := range f.Nodes() {
```
nicpottier: We can totally do this later but it seems this type of pattern (find me all the types of this node, let me tweak it and return it) is going to be super popular so having a utility method in time will be wanted.

IE, I can imagine something where you just give it the type of action you want and a receiver method. That method returns the modified thing (or nil) and the utility function takes care of fixing things up appropriately in the case of removal etc..

But this is a good baby step.

flows/actions/base_test.go
```diff
@@ -124,7 +124,7 @@ func testActionType(t *testing.T, assetsJSON json.RawMessage, typeName string) {
 
 		// if we have a localization section, inject that too
 		if tc.Localization != nil {
-			localizationPath := []string{"flows", fmt.Sprintf("[%d]", flowIndex), "localization", "spa", "ad154980-7bf7-4ab8-8728-545fd6378912"}
```
rowanseymour: the localization section for each test now needs to be the entire flow localization rather than just the action since `templating` is a separate localizable object

flows/actions/send_msg.go
```diff
@@ -91,18 +97,20 @@ func (a *SendMsgAction) Execute(run flows.FlowRun, step flows.Step, logModifier
 		if a.Templating != nil {
 			translation := sa.Templates().FindTranslation(a.Templating.Template.UUID, channelRef, []envs.Language{run.Contact().Language(), run.Environment().DefaultLanguage()})
 			if translation != nil {
+				localizedVariables := run.GetTextArray(uuids.UUID(a.Templating.UUID), "variables", a.Templating.Variables)
```
rowanseymour: potential for language mismatches here?

rowanseymour:
@nicpottier reworked some stuff in this PR so that `migrations.MigrateToVersion` which will do the heavy lifting of moving between legacy and new versions takes a `to` argument. I think with that I can merge this but have mailroom do it's migrations with that set to `13.0.0`.

Internally goflow will be running `13.1.0` flows, but the editor can continue writing `13.0.0` flows for now.

PR #737, opened 2019-07-23
📅 Add week_number expression function

rowanseymour:


excellent/functions/builtin.go
```diff
@@ -1337,6 +1338,19 @@ func Epoch(env envs.Environment, date types.XDateTime) types.XValue {
 	return types.NewXNumber(nanos.Div(nanosPerSecond))
 }
 
+// WeekNum returns the week number (1-53) of `date`.
```
rowanseymour: original ticket suggested the return should be 0-52 but turns out there's an ISO standard and no-one walks about week 0

nicpottier: Makes sense but are you sure it is only 1-53? Seems could be 54 possible values (if Jan 1 is a Sunday for example). Looks like WEEKNUM in excel returns 1-54. May be useful to call out which day the week starts in the docs too. (assume it is Monday?)

rowanseymour: So digging into this a bit more I'm not sure if the ISO way of doing this is what we want because:

- It's based on Monday being the start of the week - our WEEKDAY function copies Excel and uses Sunday = 0
- It would require returning year+week since Jan 01 can be returned as week num 53 of previous year.

Since our weekday function works like Excel's, I think I should make WEEKNUM work like Excel's too

nicpottier: Seems likely that most wanting to use this want an Excel like implementation so that works for me.

docs/functions.json
```diff
@@ -1448,6 +1448,25 @@
             }
         ]
     },
+    {
+        "signature": "week_number(date)",
```
rowanseymour: we have other functions `number()` and `format_number()` so changed this from `week_num`

excellent/functions/builtin.go
rowanseymour: if the masses start demanding ISO week numbers we can do as Excel do and have a separate ISOWEEKNUM function

PR #740, opened 2019-07-26
🎌 Add UUID field to sessions

rowanseymour:
Note that this doesn't add UUIDs to old sessions being read. Am trying to avoid at situation where a session ends up with a different UUID in the database than in the JSON, and I imagine we're going to want to backfill db UUIDs.


flows/actions/base_test.go
```diff
@@ -48,7 +49,14 @@ func TestActionTypes(t *testing.T) {
 
 	server := test.NewTestHTTPServer(49996)
 
+	typeNames := make([]string, 0)
 	for typeName := range actions.RegisteredTypes() {
+		typeNames = append(typeNames, typeName)
+	}
+
+	sort.Strings(typeNames)
```
rowanseymour: just needed these tests to run in the same order each time

PR #743, opened 2019-07-29
Set redaction policy in visitor constructor

nicpottier:
Not being set right now which means nothing is ever redacted.


contactql/visitor.go
```diff
@@ -27,7 +27,7 @@ type Visitor struct {
 
 // NewVisitor creates a new ContactQL visitor
 func NewVisitor(redaction envs.RedactionPolicy) *Visitor {
-	return &Visitor{}
+	return &Visitor{redaction: redaction}
```
rowanseymour: oops 😬

PR #746, opened 2019-07-30
🗄️ Parsed contact query should include property types

rowanseymour:
Also error if URN scheme used in contact query when URN redaction is enabled


contactql/visitor.go
```diff
@@ -69,42 +80,56 @@ func (v *Visitor) VisitCondition(ctx *gen.ConditionContext) interface{} {
...
+	if attributeKeys[propKey] {
+		propType = PropertyTypeAttribute
+	} else if urns.IsValidScheme(propKey) {
+		propType = PropertyTypeScheme
+
+		if v.redaction == envs.RedactionPolicyURNs {
+			v.errors = append(v.errors, errors.New("URN scheme not allowed"))
+		}
+	} else {
+		propType = PropertyTypeField
```
rowanseymour: so could add validation of field keys here.. just trying to decide what should be passed in from the caller.. slice of keys? slice of field assets?

contactql/visitor.go
```diff
@@ -69,7 +80,21 @@ func (v *Visitor) VisitCondition(ctx *gen.ConditionContext) interface{} {
...
-	return &Condition{key: key, comparator: comparator, value: value}
+	var propType PropertyType
+
+	if attributeKeys[propKey] {
+		propType = PropertyTypeAttribute
+	} else if urns.IsValidScheme(propKey) {
+		propType = PropertyTypeScheme
+
+		if v.redaction == envs.RedactionPolicyURNs {
+			v.errors = append(v.errors, errors.New("URN scheme not allowed"))
```
nicpottier: maybe `cannot query on redacted URNs` just to make it super obvious as to WHY this is an error.

contactql/visitor.go
nicpottier: I'm partial to that being an interface to look things up (see FieldRegistry on my PR) just because then you don't have to create these temp arrays everywhere constantly when most of them won't ever be used.

IE: https://github.com/nyaruka/mailroom/pull/144/files#diff-bd28d383d170f63012daebf08f68b901R427

rowanseymour: Yeah it's funny how if this was Python one wouldn't think twice about creating a new set of field keys on the fly

rowanseymour: I see in the mailroom evaluator you have access to `OrgAssets`, so seems we could lean on `assets.Field` here instead of me creating a new field type to hold key + data type ? I.e. so add an interface or function type which resolves a key to a `assets.Field`  and store that on the condition?

rowanseymour: Or maybe not best idea since we also want data type stored on the condition for attributes and schemes

rowanseymour: Or maybe storing a field on the condition is not best idea since we also want data type stored on the condition for attributes and schemes

nicpottier: What about pulling in the fieldish type that I used (or something similar):
  https://github.com/nyaruka/mailroom/pull/144/files#diff-b1682053a47b7b53d624c6c4e45d2c83R87   

Schemes could also be treated the same way as I do, IE, the parser does the lookup by key and from that figures out if it is a scheme/field/system attribute etc..

I like that approach because every is totally decoupled then and then ya I get to pass in the same registry you use etc.. which would be nice. The mailroom search stuff knows nothing of assets or anything that shouldn't be in gocommon.

contactql/parser.go
```diff
@@ -204,7 +212,7 @@ func (q *ContactQuery) String() string {
 }
 
 // ParseQuery parses a ContactQL query from the given input
-func ParseQuery(text string, redaction envs.RedactionPolicy) (*ContactQuery, error) {
+func ParseQuery(text string, redaction envs.RedactionPolicy, fieldResolver func(string) assets.Field) (*ContactQuery, error) {
```
nicpottier: I'm fine with this but can we declare a type for this instead of having it inline? IE:
`type FieldResolverFunc func(string) assets.Field`

contactql/visitor.go
rowanseymour: So I already had reworked it to something similar to what you have in that PR.. but still using `assets.Field` (the most basic interface) as the type for look ups. I think this dependency makes sense - you can't really do contactql without knowledge of fields, and it felt pretty weird in goflow to make a new field type since we already have `assets.Field` and `flows.Field`, and all the field types defined.

So the end result is the condition knows whether its an attribute, scheme or field, and also what data type is... 

What ya think?

contactql/parser.go
rowanseymour: updated

PR #757, opened 2019-08-13
🚨 Rename errored statuses to failed, replace fatal error events with failure events

rowanseymour:


rowanseymour:
Yeah the separation of non-fatal and fatal problems is a win.. I just think `warning/error` would be a natural language here and a more accurate reflection of what those things mean to us. But meh `error/failure` can mean whatever we need them to mean.

rowanseymour:
@nicpottier just wanna check before hitting merge.. not seeing any complications here with existing sessions. Mailroom just needs to updated to handle different statuses from goflow... but we don't ask goflow to read errored/failed sessions... we don't read error/failure events in goflow... 

only other thing would be updating the simulator to render failure events but that's trivial

nicpottier:
Yep, that sounds right.

PR #764, opened 2019-09-11
🏬 Rework services architecture

rowanseymour:
So now...

**Service**: takes a session and returns a suitable provider (or nil)
**Provider**: an implementation of service, perhaps tied to a 3rd party product

In the case of mailroom, it can implement services which resolve an org from a session, determine which type provider is right and return an instance of that to goflow.

And dtone/transferto airtime transfer provider moved here from mailroom


flows/actions/transfer_airtime.go
```diff
@@ -73,23 +73,26 @@ func (a *TransferAirtimeAction) Execute(run flows.FlowRun, step flows.Step, logM
 		sender, _ = urns.Parse("tel:" + channel.Address())
 	}
 
-	transfer, err := run.Session().Engine().Services().Airtime().Transfer(run.Session(), sender, recipient, a.Amounts)
+	airtimeSvc := run.Session().Engine().Services().Airtime(run.Session())
+	if airtimeSvc == nil {
+		logEvent(events.NewError(errors.Errorf("no airtime provider available")))
+		return nil
```
rowanseymour: what kind of error is it if the org doesn't have an airtime provider configured?

nicpottier: Seems like call above `Airtime` likely needs an error to be returned as well, would capture both that case and the one of say, config values being invalid in some way and us logging more details there.

rowanseymour: I'm thinking that either the org config has the right keys and it returns a provider, or it doesn't have the keys and it returns nil... but that might be just wanting to avoid adding another error case here

nicpottier: A  funny hill do die on in Go! :)

Fine with either just suspect knowing WHY a provider doesn't return something could be useful.

rowanseymour: In our implementation there is only one WHY.. that the config keys don't exist or they do exist. I suppose another hypothetical implementation might have other reasons but... 1) that isn't necessarily the sort of error that needs to be in the session so they can log that however they want.. 2) we can change this in the future.

PR #766, opened 2019-09-20
🔎 Add contact_query field to start_session actions

rowanseymour:
Diff is a bit bigger than you'd expect cuz I wanted to bake this functionality into the base class of `start_session` and `send_broadcast`.

### DIFF
--- a/docs/md/sessions.md
+++ b/docs/md/sessions.md
@@ -303,14 +303,14 @@ Events are created when an action wants to send a message to other contacts.
         }
     },
     "base_language": "eng",
-    "urns": [
-        "tel:+12065551212"
-    ],
     "contacts": [
         {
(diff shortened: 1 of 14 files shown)


ericnewcomer:
Are we planning to add this to send_broadcast in mailroom or introducing this as drift?

rowanseymour:
Well obviously the intention isn't to introduce drift... from the engine pov these two actions have a bunch of stuff in common because they are the two actions which are directed at other contacts. The contact query addition seemed to fit in that and well.. the editor has always offered a select subset of what is possible at an engine level.

I was tho also assuming we would be adding it to `send_broadcast`.. since that feels like a special case of `start_session` anyway. It would hopefully solve the same problems on that action re legacy variables that we're trying to solve on `start_session`.

PR #770, opened 2019-09-25
🈷️ Fix migration of localization when flow has unused base translations

rowanseymour:


legacy/testdata/flows.json
```diff
@@ -816,5 +816,198 @@
...
+                    "exit_uuid": "2ddd0916-06bc-4d59-983d-c17754d3f32a",
+                    "actions": [
+                        {
+                            "type": "reply",
+                            "uuid": "cf86e381-438a-434b-98d6-b77e23c7a094",
+                            "msg": {
+                                "fra": "Bonjour",
+                                "eng": "Hello",
+                                "spa": "Hola",
+                                "base": "G'day"
```
rowanseymour: This is the case we were seeing with the broken Swahili flow - it has `base` translations which aren't exposed in the editor - in English - and these were then sometimes overwriting the `swc` translations to become the base text in the migrated flow

So seems like we can ignore `base` translations if the flow base_language isn't `base`..? Instead of putting `base` in the flow localization section.

Does that seem correct @ericnewcomer @nicpottier ?

rowanseymour: (whether the `swc` translation got overwritten depended on the order of items pulled from that map of translations - i.e. non-deterministic)

nicpottier: > and these were then sometimes overwriting the swc translations to become the base text in the migrated flow

What are the cases where that is true? (or is that the bug?) Seems that `base` should never overwrite an explicit language but not sure about totally ignoring `base` in general.

nicpottier: In theory the editor should never be allowing the switch to a non-base language unless it is completely translated so the current fix seems valid. From a flow engine perspective I could see the argument being that the priority should be `contact language` -> `org language` -> `flow base language` -> `base`? But really I do think the flows base language probably should be the only one honored so this seems right.

rowanseymour: To be clear - if the flow base language is still `base` - we use the `base` translations as the base text in the migrated flow.

I think this scenario happens when you switch the base language of a flow from `base`.. in the editor those then become inaccessible but they hang around in the definition... I think.

nicpottier: Yep yep.

PR #775, opened 2019-10-03
🦜 NLU

rowanseymour:
Adds..

- Classifier assets
- `has_intent`  and `has_top_intent` case tests
- a `call_classifier` action which calls a..
- NLU service which provides classification
- a wit.ai implementation of that service


flows/actions/classify_nlu.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/pkg/errors"
+)
+
+func init() {
+	registerType(TypeClassifyNLU, func() flows.Action { return &ClassifyNLUAction{} })
+}
+
+const (
+	classificationCategorySuccess = "success"
+	classificationCategorySkipped = "skipped"
```
rowanseymour: this for cases like the input evaluating to empty.. no there's nothing to classify.. but it doesn't quite feel like a failure.. cuz we reserve that for actual problems vs flow design issues.

flows/actions/classify_nlu.go
```diff
@@ -0,0 +1,130 @@
...
+	classificationCategoryFailure = "failure"
+)
+
+var classificationCategories = []string{classificationCategorySuccess, classificationCategorySkipped, classificationCategoryFailure}
+
+// TypeClassifyNLU is the type for the classify NLU action
+const TypeClassifyNLU string = "classify_nlu"
```
nicpottier: bike sheddy but do we need the `_nlu`? Why not just `classify`?

flows/classifiers.go
```diff
@@ -0,0 +1,37 @@
+package flows
+
+import "github.com/nyaruka/goflow/assets"
+
+// Classifier represents an NLU classifier.
+type Classifier struct {
```
nicpottier: should this be `ClassifierNLU` if the action is `ClassifyNLU`? (to be clear my preference is to drop the nlu in both)

flows/engine/engine.go
```diff
@@ -75,7 +75,13 @@ func (b *Builder) WithWebhookService(svc WebhookService) *Builder {
 	return b
 }
 
-// WithAirtimeService sets the airtime transfer service
+// WithNLUService sets the NLU service
+func (b *Builder) WithNLUService(svc NLUService) *Builder {
```
nicpottier: `ClassifierService`? Just wondering whether we really need to pull in NLU anywhere at all. Is it adding anything?

flows/actions/classify_nlu.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		logEvent(events.NewError(err))
+	}
+
+	classification, skipped, err := a.classify(run, step, input, classifier, logEvent)
+	if err != nil {
+		logEvent(events.NewError(err))
+
+		if skipped {
+			a.saveSkipped(run, step, input, logEvent)
```
rowanseymour: all code paths write a result.. even if we never do any classification

flows/actions/classify_nlu.go
```diff
@@ -0,0 +1,130 @@
...
+	if nluSvc == nil {
+		return nil, false, errors.Errorf("no NLU provider available")
+	}
+
+	classification, err := nluSvc.Classify(run.Session(), input)
+	return classification, false, err
+}
+
+func (a *ClassifyNLUAction) saveSuccess(run flows.FlowRun, step flows.Step, input string, classification *flows.NLUClassification, logEvent flows.EventCallback) {
+	// value is name of top ranked intent if there is one
```
rowanseymour: this differs from what I had in the examples in the issue but feels more natural (also not actually used by the router tests)

flows/engine/services.go
```diff
@@ -1,23 +1,29 @@
...
 import (
+	"github.com/nyaruka/goflow/assets"
 	"github.com/nyaruka/goflow/flows"
 )
 
 // WebhookService resolves a session to a webhook provider
 type WebhookService func(flows.Session) flows.WebhookProvider
 
+// NLUService resolves a session and classifier to an NLU provider
+type NLUService func(flows.Session, assets.Classifier) flows.NLUProvider
```
nicpottier: is `Service` the right word here? This feels like a `Factory` to me in this case. IE, here's a classifier, build be a provider.

flows/actions/classify_nlu.go
rowanseymour: Was trying to follow the pattern of action names being verb+noun.. `send_msg` vs `reply`, `enter_flow` vs `flow` etc.. `classify` feels a little vague...

nicpottier: `nlu` isn't much of a noun though. `classify_text`?

flows/engine/engine.go
rowanseymour: Leaves the door open for NLU providing more than just a `Classify` method. Likewise the `AirtimeService` has a `Transfer` method cuz it might do something else one day like a `Balance` method.

flows/actions/classify_nlu.go
nicpottier: makes sense

flows/engine/services.go
rowanseymour: The current "services" interface looks like this when used in actions...

```
s := engine.Services().Airtime(session)
s.Transfer(...)

engine.Services().NLU(session, asset)
s.Classify(...)

engine.Services().Webhooks(session)
s.Call(...)
```

So maybe instead of `FooProvider` it could be `FooService` and instead of `FooService` it would be `FooServiceFactory`? I'm fine with that.

flows/engine/engine.go
nicpottier: Fair enough

flows/actions/classify_nlu.go
rowanseymour: That works for me

flows/engine/services.go
nicpottier: I think I like that better, `provider` vs `service` doesn't really mean anything while `service` vs `service_factory` is a bit clearer I think.

rowanseymour: Reworked other services in 0f446ae0413b80735ef3c9b2adb5e967416cba90 so keep this PR on theme

flows/actions/call_classifier.go
```diff
@@ -0,0 +1,130 @@
...
+//     "type": "call_classifier",
+//     "classifier": {
+//       "uuid": "1c06c884-39dd-4ce4-ad9f-9a01cbe6c000",
+//       "name": "Booking"
+//     },
+//     "input": "@input.text",
+//     "result_name": "Intent"
+//   }
+//
+// @action call_classifier
```
rowanseymour: @nicpottier @ericnewcomer 🚨 name tweak again...

I had this called `classify_text` and the event being `text_classified`.. but this feels more accurate. This action by itself doesn't really classify text... it pulls data in from an external source. The entire node (this action plus a router) is more accurately the thing that does classification. It also generates an event that records an HTTP event rather than a classification.. so trying out a naming scheme that puts it with the webhook stuff...

`call_webook` -> `webhook_called`
`call_resthook` -> `webhook_called` x N
`call_classifier` -> `classifier_called`

nicpottier: That does feel consistent, one thing to think about is what happens if we plug in speech to text. Wit treats this as just another intent call of sorts..

IE: https://wit.ai/docs/http/20160526#post__speech_link

Maybe that's a totally different thing though, Luis doesn't seem to tie these as closely together. So I'm cool with `call_classifier`

rowanseymour: Tho was about to get rid of the `Skipped` result category and now it feels a little dishonest for and action called `call_classifier` to create a `Success` result if it didn't really call the classifier... do you think you could learn to love `Success|Skipped|Failure` 😬 ?

nicpottier: I'm fine with it, just always feel that when we get into the situation of looking at more than one status invariably we add another status and then that breaks things. But agree that may not happen. I just don't love it, but that's ok.

flows/results.go
```diff
@@ -15,14 +15,14 @@ import (
...
-	Name              string          `json:"name"`
+	Name              string          `json:"name" validate:"required"`
 	Value             string          `json:"value"`
 	Category          string          `json:"category,omitempty"`
 	CategoryLocalized string          `json:"category_localized,omitempty"`
 	NodeUUID          NodeUUID        `json:"node_uuid"`
 	Input             string          `json:"input,omitempty"`
 	Extra             json.RawMessage `json:"extra,omitempty"`
-	CreatedOn         time.Time       `json:"created_on"`
+	CreatedOn         time.Time       `json:"created_on" validate:"required"`
```
rowanseymour: these two required fields now define "what qualifies as a result for a test function that expects a result"

utils/httpx/http.go
```diff
@@ -0,0 +1,87 @@
+package httpx
```
rowanseymour: this follows the pattern we've used for time, uuids and random numbers - provide a low level utility with easy mocking

docs/md/routing.md
```diff
@@ -176,6 +176,15 @@ without any tokenization.
 @(has_beginning("The Quick Brown", "quick brown")) → false
 ```
 
+<h2 class="item_title"><a name="test:has_category" href="#test:has_category">has_category(result, categories)</a></h2>
+
+Tests whether the category of a result on of the passed in `categories`
+
+
+```objectivec
+@(has_intent(results.foo, "book_flight", 0.5)) → ERROR
```
nicpottier: wrong example here.

docs/md/routing.md
```diff
@@ -273,6 +282,15 @@ Returns whether the `contact` is part of group with the passed in UUID
 @(has_group(array(), "97fe7029-3a15-4005-b0c7-277b884fc1d5")) → false
 ```
 
+<h2 class="item_title"><a name="test:has_intent" href="#test:has_intent">has_intent(result)</a></h2>
```
nicpottier: should have `name` and `confidence`

docs/md/routing.md
```diff
@@ -176,6 +176,15 @@ without any tokenization.
 @(has_beginning("The Quick Brown", "quick brown")) → false
 ```
 
+<h2 class="item_title"><a name="test:has_category" href="#test:has_category">has_category(result, categories)</a></h2>
```
nicpottier: is categories an array or a vararg?

flows/actions/call_classifier.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (a *CallClassifierAction) saveSuccess(run flows.FlowRun, step flows.Step, input string, classification *flows.NLUClassification, logEvent flows.EventCallback) {
+	// result value is name of top ranked intent if there is one
+	value := ""
+	if len(classification.Intents) > 0 {
+		value = classification.Intents[0].Name
+	}
+	extra, _ := json.Marshal(classification)
+
+	a.saveResult(run, step, a.ResultName, value, classificationCategorySuccess, "", input, extra, logEvent)
```
nicpottier: So for the result name, I take it we aren't doing `_whatever`? Or is that expected to be done by the editor? (in which case it isn't really a special case at all from the flow engine perspective just an editor convention?)

flows/engine/services.go
```diff
@@ -7,17 +7,22 @@ import (
 // WebhookServiceFactory resolves a session to a webhook service
 type WebhookServiceFactory func(flows.Session) flows.WebhookService
 
+// NLUServiceFactory resolves a session and classifier to an NLU service
+type NLUServiceFactory func(flows.Session, *flows.Classifier) flows.NLUService
```
nicpottier: Sorry, I'm still confused on this naming.

The action is called `call_classifier` but goflow is calling the NLUService for that? Should this just be `ClassifierService`? Seems that would be more consistent.

flows/services.go
```diff
@@ -50,6 +51,29 @@ type WebhookService interface {
...
+
+// NLUClassification is the result of an NLU classification
+type NLUClassification struct {
+	Intents  []ExtractedIntent            `json:"intents,omitempty"`
+	Entities map[string][]ExtractedEntity `json:"entities,omitempty"`
+}
+
+// NLUService provides NLU functionality to the engine
+type NLUService interface {
+	Classify(session Session, input string, logEvent EventCallback) (*NLUClassification, error)
```
nicpottier: `ClassifierService` can `Classify`? Feels more obvious to me.

flows/actions/call_classifier.go
rowanseymour: So we need to talk about that and whether we want these results to be special in some way and maybe `_` indicates that.. but ultimately it's the editor that decides that. I just need to update Excellent if we want names that can start with `_`.

docs/md/routing.md
rowanseymour: ah crap forgot about making these examples be more useful..

docs/md/routing.md
rowanseymour: vararg.. will add examples

flows/actions/call_classifier.go
ericnewcomer: To be clear, this skipped scenario is if the input is empty or there is a misconfiguration and we want flow authors to route that possibility on every classify node?

rowanseymour: So we're saying there's a contract for these calley actions that they always save a result.. because then that makes it cleaner for routing. Don't have to handle missing results. But there are going to be cases where the call action doesn't actually call anything (e.g. no input to classify, or an airtime transfer of zero).. in which case it saves a result to allow routing to proceed... but it's not quite call=success or call=failure

flows/engine/services.go
rowanseymour: To me it feels consistent with say, the "airtime" service can "transfer" for now.. and might do other stuff in the future. But I guess it's not quite the same because the NLU stuff is tied to particular classifier... I'll change it...

flows/actions/call_classifier.go
ericnewcomer: Got it. Just doing my job to over-communicate that we are asking users to _always_ route what very well might _always_ be an edge case for them. I don't disagree with the reasoning.

flows/actions/classify_nlu.go
ericnewcomer: Sorry, just more over-communication on this. This is because its a runtime issue, not a flow design issue correct? Flow design issues don't deserve routing structures.

rowanseymour: Yeah I think it's useful to distinguish between something like "user wants to classify empty input" (not necessarily an error from their pov.. just empty input = empty entities) vs "their credentials for wit.ai are wrong".

flows/actions/call_classifier.go
ericnewcomer: I think we are all on the same page here. My plan is to have a single result name that the user enters "My Result" and then have a hidden implicitly prefixed result name for classification, `_classify_my_result` on the action. (or whatever prefix we agree on)

rowanseymour: I think we need to unpack what hidden means

rowanseymour: And I'm honestly not sure.. on the one hand the result.. seems like it would be best hidden because there's potential for confusing users with two similar results.. on the other hand we're saying this result contains the complete classification data - i.e. if you're the 1% user that wants all the extra gory details, use this result.. but they can't if they can't find it..

ericnewcomer: Right, well there's two different hiddens here. I was referring to the user only being asked for one name on that dialog and us constructing an extra one for them behind the scenes. How we surfaces that (ie, in completion, on the dialog itself) is the next question. I'm thinking we start with fully hidden as a starting point.

nicpottier: What does fully hidden mean? Not available in completion? We can't really have things be available in one place (used in expression for a router node) and not in others.

I would say just leave out `_` leading results from completion but if they type it explicitly in an expression it is there.

nicpottier: And not included in results / exports etc..

nicpottier: Maybe as extra credit one day we can make it appear in completions if you show your clue by typing the first `_`

rowanseymour: I like the idea of `_foo` results being completely hidden for now and becoming available if user indicates they know what they're doing by typing `_`.  I think that means from flow inspection pov.. we don't return those results so RP never knows about them and they don't appear on dashboards or in exports etc.

rowanseymour: Ok to clarify what we're agreeing to here..

- User enters `Foo` as the response name - that is the name of the router result and the action result is `_Foo Classification` which snakes to `_foo_classification`
- Editor won't display result names with `_` prefix in auto-completion for now.. (for extra credit.. typing `_` does add them to auto-completion)
- RP excludes these results from the results dashboard and exports

I got that right @ericnewcomer @nicpottier ?

PR #777, opened 2019-10-08
🍪 Classification using LUIS

rowanseymour:


services/classification/luis/client.go
```diff
@@ -0,0 +1,72 @@
...
+	httpClient *http.Client
+	appURL     string
+	key        string
+}
+
+// NewClient creates a new client
+func NewClient(httpClient *http.Client, appURL, key string) *Client {
+	return &Client{
+		httpClient: httpClient,
+		appURL:     appURL,
```
rowanseymour: looks like LUIS endpoints can differ completely between apps so need full URL to be stored with classifier

services/classification/luis/client_test.go
```diff
@@ -0,0 +1,88 @@
...
+				"entities": [
+				  {
+					"entity": "quito",
+					"type": "City",
+					"startIndex": 17,
+					"endIndex": 21,
+					"score": 0.9644149
+				  }
+				],
+				"sentimentAnalysis": {
```
rowanseymour: should we just throw this in with the entities if it exists? it's a checkbox on the LUIS side

services/classification/luis/client.go
rowanseymour: updated to take endpoint + appID + key

services/classification/luis/client_test.go
nicpottier: Sure, so entity of type `sentiment` with the entity being the `label`, `score` etc..

rowanseymour: well our common entity format is `value` + `confidence` so it would just become one of those

services/classification/luis/client.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// PredictResponse is the response from a predict request
+type PredictResponse struct {
+	Query             string            `json:"query"`
+	TopScoringIntent  *ExtractedIntent  `json:"topScoringIntent"`
+	Intents           []ExtractedIntent `json:"intents" validate:"required"`
+	Entities          []ExtractedEntity `json:"entities"`
+	SentimentAnalysis SentimentAnalysis `json:"sentimentAnalysis"`
+}
```
nicpottier: Totally a style call but I prefer to put everything in one struct if I can and there isn't other reuse. Makes it easier to read the entire request as one struct.

rowanseymour: You mean anonymous structs? Can try that but I've always run into problems then constructing them in tests etc

nicpottier: Ya, just a question of usage. If you end up using them independently then makes sense to pull out as you have, but if not having a giant one makes the relationship clearer.

PR #778, opened 2019-10-09
💸 Fix airtime transfers

rowanseymour:
- Update DTOne airtime client to use new http utilities
- Add cmd for testing airtime transfers
- Rework transfer_airtime action to *always* save a result (like `call_classifier`)


cmd/transferairtime/main.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error parsing assets")
+	}
+
+	eng := engine.NewBuilder().WithAirtimeServiceFactory(svcFactory).Build()
+	env := envs.NewEnvironmentBuilder().Build()
+	contact := flows.NewEmptyContact(sa, "", "", nil)
+	contact.AddURN(flows.NewContactURN(destination, nil))
+
+	_, sprint, err := eng.NewSession(sa, triggers.NewManual(env, assets.NewFlowReference(assets.FlowUUID("2374f60d-7412-442c-9177-585967afa972"), "Airtime"), contact, nil))
```
nicpottier: Is this building a flow / session etc, because you want to test the whole thing or because those are requirements to use the service? If the latter feels like our abstraction may be wrong.

Guessing this entire thing exists just so you could easily test against the live service while testing?

flows/actions/call_classifier.go
```diff
@@ -94,12 +88,12 @@ func (a *CallClassifierAction) classify(run flows.FlowRun, step flows.Step, inpu
 		return nil, false, errors.Errorf("missing %s", a.Classifier.String())
 	}
 
-	nluSvc := run.Session().Engine().Services().NLU(run.Session(), classifier)
-	if nluSvc == nil {
-		return nil, false, errors.Errorf("no NLU service available for %s", a.Classifier)
+	svc := run.Session().Engine().Services().NLU(run.Session(), classifier)
+	if svc == nil {
+		return nil, false, errors.Errorf("no classifier service available for %s", a.Classifier)
```
nicpottier: Not really part of this PR but really feels to me that the factory method should return an error so we have some idea as to why nil is being returned here. (IE, missing config values vs it doesn't exist)

cmd/transferairtime/main.go
rowanseymour: Yeah I just thought why not use a flow so the whole shebang gets a bit more testing. 

For my next trick I'm going to make the all the unit tests into a flow!

flows/actions/call_classifier.go
rowanseymour: Let me merge this and take a look at that

cmd/transferairtime/main.go
rowanseymour: Guess I wouldn't really call this an end-to-end test as much as debug tool I needed to get some real world data for testing. With `-v` you see all the HTTP back and forth...

```
transferairtime -v -dtone.login=XXXXXXX -dtone.token=XXXXXX 593999000001 1 USD
```

Could be useful for...  salaries.. 🤷‍♂

nicpottier: Yep yep, was commenting more the pulling in the flow stuff, not having a command line salary sender.

PR #782, opened 2019-10-15
📒 Log all HTTP activity

rowanseymour:
* Fixes not handling non-200 responses and connection errors in services correctly
* Changes service events to include multiple HTTP event logs


flows/actions/testdata/call_classifier.json
```diff
@@ -104,6 +104,14 @@
     },
     {
         "description": "Result with category success created if classification happens",
+        "http_mocks": {
```
rowanseymour: testing these with real wit client with mocked HTTP calls to better test recording of non-successful HTTP requests

Issue #784, opened 2019-10-24
Make legacy migrations part of ReadFlow


rowanseymour:
We're going to need to be able to read legacy flows for the foreseeable future because we allow importing of old flows. Currently `definition.ReadFlow` can only read a v13 flow and one must use `legacy.ReadFlow` to read a v11 or earlier flow. 

The legacy flow migration code also uses real flow objects structs (the actions etc).. so it always has to produce a flow in the current engine version. Any flow spec changes we make in future have to be implemented as both a new migration and a change to the legacy migrations.

Ideally things would work like this...

1. ReadFlow inspects the JSON to determine whether it's a legacy flow
2. A legacy package migrates that to 13.0.0
3. New migrations migrate that to 13.x.y
4. Read Flow returns a flow in the current engine version

This would simplify things greatly for callers who no longer have to know what version a flow was.

PR #785, opened 2019-10-24
👴🏼 Refactor legacy flow support

rowanseymour:
Addresses #784

* Makes handling legacy flows transparent to the caller. Everything goes thru `definition.ReadFlow`.
* Breaks dependency of legacy migration code on current flow structure objects so those can evolve.


flows/definition/legacy/definition_test.go
```diff
@@ -75,21 +74,21 @@ var legacyRuleSetHolderDef = `
 	"action_sets": [
 		{
 			"uuid": "5b977652-91e3-48be-8e86-7c8094b4aa8f",
-			"x": 0, "y": 200, 
+			"x": 0, "y": 2200,
```
rowanseymour: bumping these so the migrated node for the ruleset always ends up being the first node in the list

flows/engine/assets.go
```diff
@@ -24,7 +24,7 @@ type sessionAssets struct {
 var _ flows.SessionAssets = (*sessionAssets)(nil)
 
 // NewSessionAssets creates a new session assets instance with the provided base URLs
-func NewSessionAssets(source assets.Source) (flows.SessionAssets, error) {
+func NewSessionAssets(source assets.Source, migrationConfig *definition.MigrationConfig) (flows.SessionAssets, error) {
```
nicpottier: Just a thought, but should the migration config just be part of the asset source? Is there any scenario where the config is different given the same source? Just wondering why those would be separate, since the source has the old definition flows it seems appropriate it would also know the config needed to do the migration?

Don't feel strongly but interface just feels slightly odd, but admit I may not have the full perspective there.

rowanseymour: The way things are setup currently, asset sources don't know anything about the flow versions - at an asset level a flow is just a JSON blob with a UUID and name.. it's the `FlowAssets` object in `SessionAssets` that reads and caches flow definitions. So `AssetSource` could own the migration config.. but it would never use it.. so I guess it feels more natural as an argument to `SessionAssets`

nicpottier: AssetSources don't really use any of their things though, it's just an interface to hand stuff over isn't it? I don't know, to me it feels as much a part of data as anything else. Again if there is no need to ever have the same source using a different migration config then why separate.

Anyways, approval given! :)

PR #790, opened 2019-10-30
Escaping of expressions in contact queries

rowanseymour:


antlr/ContactQL.g4
```diff
@@ -1,34 +1,51 @@
...
+		UnicodeLetter
+		| UnicodeDigit
+		| '_'
+		| '.'
+		| '-'
+		| '+'
+		| '/'
+		| '@'
+	)+;
+STRING: '"' (~["] | '\\"')* '"';
```
rowanseymour: This line is the only thing that changed in this file besides vscode insisting it be re-formatted.

Changing contact QL to use same (golang) string escaping as Excellent2 instead of the old Excel/Excellent1 `""` syntax.. otherwise things get really weird.

Took a peek in the TextIt database and I can't see any queries involving string escaping.

flows/expressions_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/nyaruka/goflow/flows"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestContactQueryEscaping(t *testing.T) {
+	assert.Equal(t, `""`, flows.ContactQueryEscaping(``))
+	assert.Equal(t, `"bobby tables"`, flows.ContactQueryEscaping(`bobby tables`))
+	assert.Equal(t, `"\"\" OR (id = 1)"`, flows.ContactQueryEscaping(`"" OR (id = 1)`))
```
nicpottier: Can the input contain an escape?

IE, what happens with a query like `name = "\"foo"` Wouldn't that turn into `name = "\\"foo"`?

rowanseymour: that would become `"\\\"foo"`.. escaped `\` and escaped `"`, interpreted by the query parser as the value `\"foo`

PR #791, opened 2019-11-06
🧙‍♂️ Tweak criteria for deciding whether to try reading a flow as legacy

rowanseymour:


flows/definition/flow.go
```diff
@@ -348,22 +348,30 @@ type MigrationConfig struct {
 
 // ReadFlow a flow definition from the passed in byte array, migrating it to the spec version of the engine if necessary
 func ReadFlow(data json.RawMessage, migrationConfig *MigrationConfig) (flows.Flow, error) {
```
rowanseymour: 1. uuid and spec_version are required for new flows, so first of all check if we have those
2. if we have action_sets/rule_sets/flow_type keys, try reading as legacy
3. if not, then it's not a flow

PR #794, opened 2019-11-07
🗝️ Expose trigger keyword in expressions

rowanseymour:
Adds #786 

Also adds a bunch of tests.. and fixes a couple of issues found via those tests


flows/triggers/manual.go
```diff
@@ -37,23 +37,29 @@ type ManualTrigger struct {
 
 // NewManual creates a new manual trigger
 func NewManual(env envs.Environment, flow *assets.FlowReference, contact *flows.Contact, params *types.XObject) flows.Trigger {
-	if params == nil {
```
rowanseymour: this logic needs to be in `Context()` so it applies to triggers loaded from JSON too

flows/triggers/flow_action.go
```diff
@@ -77,7 +77,7 @@ var _ flows.TriggerWithRun = (*FlowActionTrigger)(nil)
 
 type flowActionTriggerEnvelope struct {
 	baseTriggerEnvelope
-	RunSummary json.RawMessage `json:"run_summary"`
```
rowanseymour: You get a panic if this field isn't included - mailroom only creates flow_action triggers with this set

PR #798, opened 2019-11-11
@ Globals

rowanseymour:


assets/interfaces.go
```diff
@@ -119,6 +119,21 @@ type Flow interface {
 	Definition() json.RawMessage
 }
 
+// Global is a named constant.
+//
+//   {
+//     "key": "organization_name",
+//     "name": "Organization Name",
```
rowanseymour: wasn't sure if name would be necessary but included it in the end because 1) it makes these very like contact fields, 2) it means user can edit a name via the UI and we generate a snaked key, rather than expecting users to enter something which is snaked

nicpottier: Oy, this one requires some thought, I am super scared of repeating the nightmare which is contact fields here. I would say that keys do not change is an anti-pattern and only that way because we are (currently) too lame to rewrite everybody's flows.

Can we get away with ONLY name? And say we are going to tackle fixing flows eventually. (and breaking them in the meantime on renames?)

rowanseymour: Well the other way to think about this is that we might as well make globals as much like fields as we can because whatever solution we come up with needs to work with fields... but sounds like we need to chat about https://github.com/rapidpro/rapidpro/issues/333 and start figuring out how we're going to crack that

nicpottier: Or just don't allow renaming. Renaming is a lot more important for fields because the value is spread across all these contacts in the system. It isn't so important when there is only one value. So we can launch with just an immutable name for the time being, which makes it clear to people that they need to migrate their flows if they really want to change the name and we completely sidestep the hell that is mismatched keys and names.

rowanseymour: nightmare, hell, lame, super scared.. I think there is a danger in overstating the problem here. I personally haven't heard many complaints from users.. tho maybe it's one of those things they just accept as not possible. Agree it's weird when users change field names to something completely different to the key.. but I also imagine many users get that, and only make cosmetic changes to the field names.

If we say globals can't be renamed, then why not have both key and name? If we need it in both RP (displayed to the user, searchable) and the engine, then might as well ensure they have the same key.

nicpottier: We have definitely had lots of drama around field renaming. Maybe you aren't involved but it happens and is super confusing for people. There's just no reason to go down a path we already know doesn't work well (for very little benefit).

I'm fine with having both key and name here, just not renaming and having these drift from each other unless we are doing the remapping for users.

rowanseymour: I'm ok with not allowing renaming (are you suggesting this for both fields and globals?) tho I do think we could also solve this by just explaining to users via a warning when they rename a field that the field is still referenced in flows by the original key.

Anyways I think we proceed with key/name/value here. We have UUID on field assets but it's optional and only for mailroom's benefit, and not needed here.

nicpottier: I think contact fields need renaming because there's no easy way to copy values across contacts. But ya for globals certainly at launch I don't think it is needed.

ericnewcomer: +1 to no renames, not unless we take care of the repercussions for them. I remember all kinds of garbage from names and keys no longer matching. We let our users down a bit on that one.

PR #803, opened 2019-11-20
📖 Add topic to send_msg actions

rowanseymour:
Can be tweaked depending on conclusions of https://github.com/rapidpro/rapidpro/issues/1103

### DIFF
--- a/flows/msg.go
+++ b/flows/msg.go
@@ -9,7 +9,7 @@ import (
 )
 
 func init() {
-	utils.Validator.RegisterAlias("msg_topic", "eq=event|eq=account|eq=purchase")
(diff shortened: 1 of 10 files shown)


flows/msg.go
```diff
@@ -8,6 +8,21 @@ import (
...
+
+// MsgTopic is the topic, as required by some channel types
+type MsgTopic string
+
+// possible msg topic values
+const (
+	NilMsgTopic      MsgTopic = ""
+	MsgTopicEvent    MsgTopic = "event"
+	MsgTopicAccount  MsgTopic = "account"
+	MsgTopicPurchase MsgTopic = "purchase"
```
nicpottier: Perhaps add `agent` as well since that is an option:
  https://developers.facebook.com/docs/messenger-platform/send-messages/message-tags/

PR #807, opened 2019-11-26
📭 Implement email as a service

rowanseymour:
`email_created` replaced with `email_sent` and engine constructed with an email service factory


flows/events/email_sent.go
```diff
@@ -0,0 +1,41 @@
...
+//     "addresses": ["foo@bar.com"],
+//     "subject": "Your activation token",
+//     "body": "Your activation token is AAFFKKEE"
+//   }
+//
+// @event email_sent
+type EmailSentEvent struct {
+	baseEvent
+
+	Addresses []string `json:"addresses" validate:"required,min=1"`
```
nicpottier: Since we are creating a new event, maybe we should look to the future and use `To` here? (which would allow us to add CC / BCC in the future?)

rowanseymour: Kinda annoying we didn't think of that when we made the action but I can change this and if we we ever add `cc`/`bcc` to the action.. we can rename `addresses` there too.. I'm hoping we can get to a point where minor flow migrations aren't a big deal.

PR #809, opened 2019-11-28
📚 Add support for localized function listings

rowanseymour:
Building on the PO stuff previously added for extracting POs from flows


locales/pt_BR/functions.po
```diff
@@ -0,0 +1,494 @@
...
+msgstr ""
+
+msgid "Converts `text` into something that can be read by IVR systems."
+msgstr ""
+
+msgid "Converts `text` to lowercase."
+msgstr "Converte `text` em minúsculas."
+
+msgid "Converts `text` to uppercase."
+msgstr "Converte `text` em maiúsculas."
```
rowanseymour, line 39: The `text` part is a reference to an arg in the display function signature. We almost certainly don't want to go the Excel route of translating function names, but you could conceivably translate args in the signature.. but then maybe that's more confusing because the function name is in English but the args are translated, e.g. `upper(texto)`. So for now I'm assuming these stay untranslated.

Issue #814, opened 2019-12-19
Error when a webhook is too large should include URL of webhook


nicpottier:
Currently if we get a webhook response that is larger than 10k we don't process it. That's fine but we don't have anything in our session that logs what URL was called.

Would be nice to have that present.

Also, not sure if this is a migration thing but when we get that kind of failure we also get knock on errors because the expected result isn't present, that's confusing for users:

<img width="768" alt="Screen Shot 2019-12-19 at 9 59 59 AM" src="https://user-images.githubusercontent.com/168067/71197424-6d001b00-2246-11ea-9923-296813c180c1.png">

rowanseymour:
Hmm me thinks both of these are the same problem - we don't save a webhook result under certain error conditions. If the response is 400+ or we timeout we still have a webhook result - a record of the URL and something to route on. But exceeding the 10,000 byte limit is just an error.

We've designed airtime transfers and classification actions to _always_ save a result and it seems we should try to do the same with webhooks.

nicpottier:
I do wonder whether we should reconsider our size approach. Realize we currently do our limits because we log the entire payload and put the entire thing in the result, but maybe we should just trim the saved payload and do something similar to the "first n keys" we used to do in the old world? Or maybe you get to access the webhook full body in the same sprint but not in future ones if over a certain size? That may be a bit subtle but this is kind of an annoying limitation, especially at 10k. 

It makes sense when you consider we save the whole thing to the session but not really to a user and they just want their thing to work. (in this particular case they also did not have access to the API (it is an external one) so asking them to reduce the size of the returned payload is a big ask)

Could trimming take place at the mailroom side of things maybe? (with a log entry that we did so?) So that in the sprint itself you get the whole shebang but we end up only logging and saving to the session some piece of it?

rowanseymour:
Something to consider is that the payload ends up being recorded in 2 places in the session:

1. Parsed as extra on the result 
2. As the response dump on the `webhook_called` event

The latter is just for logging so I think it would be fine to just trim that to X characters. We currently have a `body_ignored` flag on that event to record when the body isn't a type we want to read.. ideally that would be something like `body_state = full|trimmed|ignored`

The former is where things get tricky but I agree the current situation sucks for users who are trying to pull data from endpoints they don't control.

[...]

PR #815, opened 2019-12-20
Perform URL validation in call_webhook and skip action appropriately

rowanseymour:
Have been working on a bigger change to call_webhook that would mean it always saves a result and also need to think of a good solution to https://github.com/nyaruka/goflow/issues/814#issuecomment-567691761

But for now I think with the editor dropping URL validation it's important that goflow gracefully handles things which evaluate to invalid URLs

### DIFF
--- a/flows/actions/call_webhook.go
+++ b/flows/actions/call_webhook.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/nyaruka/goflow/flows"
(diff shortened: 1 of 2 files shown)


flows/actions/call_webhook.go
```diff
@@ -11,6 +12,8 @@ import (
 	"golang.org/x/net/http/httpguts"
 )
 
+func isValidURL(u string) bool { _, err := url.Parse(u); return err == nil }
```
rowanseymour: see https://golang.org/cl/49930.. it's just really annoying to not be able to use `url` as a variable name

PR #817, opened 2020-01-06
🐿️ Fix migration of eq/after/before ruleset tests

### LINKED ISSUE #816: Converting from Old to New Editor Loses 'Has A Date After'
When switching from old to new editor, I've seen my values go from `-7` or `-30` to `0` on more than one occasion, causing problems in my flows. I can't reproduce in TextIn because there's no more option to start with or return to the old editor, but it happened consistently with my flows everywhere I did it. Maybe something with the negative numbers. If fixed already, apologies.

image

rowanseymour:
Fixes #816

### DIFF
--- a/flows/definition/legacy/definition.go
+++ b/flows/definition/legacy/definition.go
@@ -266,7 +266,7 @@ type wardTest struct {
 	District string `json:"district"`
 }
 
-var relativeDateTest = regexp.MustCompile(`@\(date\.today\s+(\+|\-)\s+(\d+)\)`)
+var relativeDateTest = regexp.MustCompile(`@\(date\.today\s+(\+|\-)\s+(\-?\d+)\)`)
 
 //------------------------------------------------------------------------------------------
 // Migrated flow objects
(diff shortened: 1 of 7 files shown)


ericnewcomer:
Not sure this is enough. The editor doesn't actually parse functions inside operands but instead relies on config for this. The case should have a config that looks something like:

```json
"_ui": {
    "nodes": {
        "161fda79-117d-4940-be26-54a0693da457": {
        "type": "wait_for_response",
        "position": {
            "left": 0,
            "top": 0
        },
        "config": {
            "cases": {
            "5f2fb030-8b09-43a9-ae46-ce99dfc70531": {
                "arguments": [
                "-10"
                ]
            }
            }
        }
        }
    }
}

```

rowanseymour:
Ah darn that's what I first thought and then I created a new flow, looked at the JSON, somehow missed that config part, and assumed it was parsing the expression.. doh

rowanseymour:
@ericnewcomer so in the end all the grand changes to the expression aren't necessary but I think we should keep them in for the sake of migrated flows matching new flows. Have fixed the bit where we parse out the date delta to put in the config (which wasn't being tested so was always going to be broken!)

PR #821, opened 2020-01-08
🕺 Add support for jitter in webhook retries

rowanseymour:


docs/functions.json
```diff
@@ -944,11 +944,11 @@
         "examples": [
             {
                 "template": "@(rand())",
-                "output": "0.3849275689214193274523267973563633859157562255859375"
+                "output": "0.607552015674623913099594574305228888988494873046875"
```
rowanseymour: bunch of these changes are just because there's now an extra call to the mocked RNG being made during setup of the session used for testing and doc generation

utils/httpx/retries.go
```diff
@@ -1,22 +1,52 @@
...
+}
+
+// Backoff gets the backoff time for the nth retry
+func (r *RetryConfig) Backoff(n int) time.Duration {
+	if n >= len(r.Backoffs) {
+		panic(fmt.Sprintf("%d not a valid retry number for this config", n))
+	}
+
+	base := r.Backoffs[n]
+	jitter := time.Duration(r.Jitter * float64(random.IntN(int(base))-(int(base)/2)))
```
rowanseymour: this is more like the implementation in https://github.com/hashicorp/go-retryablehttp than the AWS document in that the jitter is spread around the base backoff and not always subtracted, e.g. if you have 100% jitter and a base value of 10secs, the jittered value will be in the range [5,15]

utils/httpx/http.go
```diff
@@ -23,10 +23,11 @@ func Do(client *http.Client, request *http.Request, retries *RetryConfig) (*http
...
 		response, err = currentRequestor.Do(client, request)
 
-		if retries != nil && retry < len(retries.Delays) {
-			delay := retries.Delays[retry]
-			if retries.ShouldRetry(request, response, delay) {
-				time.Sleep(delay)
+		if retries != nil && retry < retries.MaxRetries() {
+			backoff := retries.Backoff(retry)
+
+			if retries.ShouldRetry(request, response, backoff) {
```
rowanseymour: `Retry-After` headers are compared against the jittered value.. a little less predictable.. but seems less complicated than sometimes using jittered value and sometimes not

Issue #822, opened 2020-01-09
Include parent result references in flow inspection


rowanseymour:
These can be used to determine which results need to be set when adding a `enter_flow` on the parent flow. Maybe something like...

```json
{
  "templates": [
    "Hi @contact"
  ],
  "dependencies": {
    "fields": [
      {"key": "gender", "name": "Gender"}
    ]
   },
  "results": [
    {
      "key": "response_1",
      "name": "Response 1",
      "categories": ["Yes", "No"],
      "node_uuids": ["72a1f5df-49f9-45df-94c9-d86f7ea064e5"]
    }
  ],
  "result_refs": {
    "parent": [
       "age",
       "response_1"
    ]
  }
}
```

Trying to think of a reason it would be useful to capture `@child.results.*` references too...

nicpottier:
I would probably vote to just have `parent_refs` without anything under it but I suppose I could be convinced to just do both even though I don't think `child` is really useful for anything. (since we can't know for sure what child it is)

rowanseymour:
Yeah I'm fine with that.. I'm coming up blank on reasons to include child refs..

ericnewcomer:
Same. Parent is the only one that makes any sense to me. This is neat though!

PR #823, opened 2020-01-09
👩‍👧 Include parent result references in flow inspection

rowanseymour:
Adds #822

### DIFF
--- a/cmd/docgen/completion/completion.go
+++ b/cmd/docgen/completion/completion.go
@@ -3,17 +3,16 @@ package completion
 import (
 	"fmt"
 
-	"github.com/nyaruka/goflow/utils"
 	"github.com/pkg/errors"
 )
 
 // types available in root of context even without a session
-var rootNoSessionTypes = utils.StringSet(map[string]bool{
+var rootNoSessionTypes = map[string]bool{
 	"contact": true,
 	"fields":  true,
 	"globals": true,
 	"urns":    true,
-})
+}
 
 // Completion generates auto-complete paths
 type Completion struct {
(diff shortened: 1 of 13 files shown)


cmd/docgen/completion/completion.go
```diff
@@ -26,7 +27,7 @@ func NewCompletion(types []Type, root []*Property) *Completion {
 	// extract types which are available in a context without a session
 	rootNoSession := make([]*Property, 0, len(rootNoSessionTypes))
 	for _, p := range root {
-		if rootNoSessionTypes[p.Type] {
+		if rootNoSessionTypes.Contains(p.Type) {
```
rowanseymour: Feel a little like i'm cheating in go here but I find the "set of strings" such a common pattern.. and this just a little bit more readable

nicpottier: I don't care that much but I actually think there's far more value to sticking to the idiomatic way of doing it instead of creating a cutesy wrapper. Because now I have to wonder whether that wrapper does anything else different. Nobody who uses go is going to have any problem understanding the map version immediately.

rowanseymour: I bit the bullet here because sometimes you also need to end up with the strings as a slice and the regular go way doesn't do that.. getting keys from maps is one of my major gripes with go

nicpottier: Wouldn't a single function satisfy that if you are doing it often? I don't know, just feels like trying to make Go into Java. :)

rowanseymour: Maybe I dunno.. ok I'll ditch this.. and add `StringMapKeys` function

flows/definition/flow.go
```diff
@@ -100,10 +100,13 @@ func (f *flow) validate() error {
 
 // Inspect enumerates dependencies, results etc
 func (f *flow) Inspect() *flows.FlowInfo {
+	assetRefs, parentRefs := f.extractAssetAndParentRefs()
```
rowanseymour: reworked this so that we don't have to go enumerating and parsing templates twice

PR #826, opened 2020-01-13
Fix creation of non-nil interface to nil structs in context

rowanseymour:


flows/expressions.go
```diff
@@ -14,7 +14,7 @@ type Contextable interface {
 }
 
 // Context generates a lazy object for use in expressions
-func Context(env envs.Environment, contextable Contextable) *types.XObject {
+func Context(env envs.Environment, contextable Contextable) types.XValue {
```
rowanseymour: Another week, another very subtle bug caused by go's feature of automatically wrapping nil structs in non-nil interfaces. In this case nil XObjects instances returned by this function become non-nil interfaces which blow up when you try to marshal them

PR #828, opened 2020-01-15
🧰 Session Current Context Tweaks

rowanseymour:
* Add support for marshaling XObjects with their defaults
* Add tool for walking the context to find objects

So that in mailroom we can do...

```go
context := session.CurrentContext()

tools.ContextFindObjects(context, func(o *types.XObject) {
  o.SetMarshalDefault(true)
})
```

and get a version of the context where objects defaults are included in JSON as `__default__`

### DIFF
--- a/excellent/tools/context_walk.go
+++ b/excellent/tools/context_walk.go
@@ -9,8 +9,8 @@ func ContextWalk(context *types.XObject, callback func(types.XValue)) {
 	contextWalk(context, callback)
 }
 
-// ContextFindObjects traverses the given context invoking the callback for each object found
-func ContextFindObjects(context *types.XObject, callback func(*types.XObject)) {
+// ContextWalkObjects traverses the given context invoking the callback for each object found
+func ContextWalkObjects(context *types.XObject, callback func(*types.XObject)) {
 	contextWalk(context, func(v types.XValue) {
 		switch typed := v.(type) {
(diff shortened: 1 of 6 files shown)


excellent/tools/context_walk.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/nyaruka/goflow/excellent/types"
+)
+
+// ContextWalk traverses the given context invoking the callback for each value
+func ContextWalk(context *types.XObject, callback func(types.XValue)) {
+	contextWalk(context, callback)
+}
+
+// ContextFindObjects traverses the given context invoking the callback for each object found
+func ContextFindObjects(context *types.XObject, callback func(*types.XObject)) {
```
nicpottier: minor but the different between `walk` and `find` here feels a bit arbitrary.. when I saw find I was thinking, "oh, you pass in the type you want to get callbacks for".. to me `ContextWalkObjects` feels a bit more consistent / intuitive.

rowanseymour: Fine with that

rowanseymour:
@nicpottier have gone back and forth and what the least awkward way of doing this would be, since you have no context inside `MarshalJSON` methods. My first attempt was modifying how the tree is constructed.. but that got really complicated.. this is definitely simpler.

PR #831, opened 2020-01-17
🐳 Bigger webhook bodies

rowanseymour:
Adds #814 

- Adds `response_as_extra` to `call_webhook` actions
- Migration (13.2) sets this to true for existing flows/actions
- Response body only set as extra on result if this is true


flows/actions/base.go
```diff
@@ -120,15 +121,30 @@ func (a *baseAction) saveResult(run flows.FlowRun, step flows.Step, name, value,
...
+	}
 
 	a.saveResult(run, step, name, value, category, "", input, extra, logEvent)
 }
 
+func (a *baseAction) updateWebhook(run flows.FlowRun, call *flows.WebhookCall) {
+	parsed := types.JSONToXValue(utils.ExtractResponseJSON([]byte(call.Response)))
+
+	switch typed := parsed.(type) {
+	case nil, types.XError:
```
rowanseymour: `@webhook` set as empty object in cases of empty or invalid responses

flows/actions/base.go
```diff
@@ -120,15 +121,30 @@ func (a *baseAction) saveResult(run flows.FlowRun, step flows.Step, name, value,
...
 // helper to save a run result based on a webhook call and log it as an event
-func (a *baseAction) saveWebhookResult(run flows.FlowRun, step flows.Step, name string, webhook *flows.WebhookCall, status flows.CallStatus, logEvent flows.EventCallback) {
+func (a *baseAction) saveWebhookResult(run flows.FlowRun, step flows.Step, name string, webhook *flows.WebhookCall, status flows.CallStatus, bodyAsExtra bool, logEvent flows.EventCallback) {
 	input := fmt.Sprintf("%s %s", webhook.Method, webhook.URL)
 	value := strconv.Itoa(webhook.StatusCode)
 	category := webhookStatusCategories[status]
-	extra := utils.ExtractResponseJSON(webhook.Response)
+
+	var extra json.RawMessage
+	if bodyAsExtra && len(webhook.ResponseBody) < 10000 && json.Valid(webhook.ResponseBody) {
```
rowanseymour: so.. this is slightly different to what I said in #814.. but seems to me that not-setting extra if it's too big is no worse than erroring if it's too big.. both are runtime changes in behavior when the payload is big.. and at least this way.. 

- a user can use `@webhook` and not run into problems even if they've forgotten to uncheck "Save response as extra" on all the webhook actions
- that option on the action dialog can explain the 10K limit as well as advising that users use `@webhook`

flows/actions/base.go
```diff
@@ -120,15 +124,30 @@ func (a *baseAction) saveResult(run flows.FlowRun, step flows.Step, name, value,
...
 // helper to save a run result based on a webhook call and log it as an event
-func (a *baseAction) saveWebhookResult(run flows.FlowRun, step flows.Step, name string, webhook *flows.WebhookCall, status flows.CallStatus, logEvent flows.EventCallback) {
+func (a *baseAction) saveWebhookResult(run flows.FlowRun, step flows.Step, name string, webhook *flows.WebhookCall, status flows.CallStatus, bodyAsExtra bool, logEvent flows.EventCallback) {
 	input := fmt.Sprintf("%s %s", webhook.Method, webhook.URL)
 	value := strconv.Itoa(webhook.StatusCode)
 	category := webhookStatusCategories[status]
-	extra := utils.ExtractResponseJSON(webhook.Response)
+
+	var extra json.RawMessage
+	if bodyAsExtra && len(webhook.ResponseBody) < resultExtraMaxBytes && json.Valid(webhook.ResponseBody) {
```
rowanseymour: so.. this is slightly different to what I said in #814.. but seems to me that not setting extra if it's too big is no worse than erroring if it's too big.. both are runtime changes in behavior when the payload is big.. and at least this way.. 

- a user can use `@webhook` and not run into problems even if they've forgotten to uncheck "Save response as extra" on all the webhook actions
- that option on the action dialog can explain the 10K limit as well as advising that users use `@webhook`

nicpottier: So to be clear we are saying that "Save webhook as extra" is a magic opt-in that appears only on flows created before this change ya? Or are you thinking that is always available?

I don't necessarily hate it being always available but generally feel like the fewer options we expose the better, we don't want actions with a zillion options so should really fight to remove as many as we can. So would hope that we would only show this for "new-legacy" flows.

rowanseymour: Yeah I think the hypothetical end goal here is that no flows use this option and users forget it ever was an option. But it needs to be initially true and visible where we need to be backwards compatible.. so..

- Existing 11.x flows.. should be migrated with `response_as_extra=true` because they could be relying on `@legacy_extra` (need to fix that in this PR..)
- Existing 13.x flows are going to show it because they will be migrated with `response_as_extra=true` to maintain existing behavior.
- New 13.x flows don't need to show it - `response_as_extra` can be omitted as it defaults to `false`

So maybe editor can use presence of `response_as_extra` to determine if it's looking at an existing flow? @ericnewcomer?

Is there a point for an existing flow where the user has set it to false on all webhook actions that we can drop it from the UI?

nicpottier: Ugh, man these all sound kinda terrible honestly. I know you suggested this earlier but maybe it is worth revisiting the pros/cons of just only setting it when less than 10k and removing the `response_as_extra` option entirely? IE, we always set to extra when less than 10k but make `@webhook` be transient. All our docs talk about using `@webhook` and that uses up to 1M responses but only for the last thing?

Not sure all the above is worth the effort of the pure approach, even if `extra` becomes somewhat unpredictable. (but we can tell them to use `webhook` then instead)

rowanseymour: I think I'd prefer that... a somewhat related thought at https://github.com/rapidpro/rapidpro/issues/923#issuecomment-576785859

rowanseymour: The bit I'm not sure about is making `@webhook` transient.. there are likely to be users relying on that not being the case.. so I think we have to continue reconstructing that from `.extra` if those exist.

nicpottier: Sure, I guess that makes sense.

rowanseymour:
- In the `webhook_event`, the request and response traces are both trimmed at a rune-level, and I just made it so they end with elipses when they have been trimmed `"..."`. Those events don't care what the type is because they're never parsed as JSON - they're just for logging.
- `.extra` on results has to be valid JSON since it's part of the session blob.. so we only set those if the response body is less than 10K AND valid JSON.

PR #833, opened 2020-01-21
🕵️‍♀️ Allow getting current context even for ended sessions

rowanseymour:


flows/engine/session.go
```diff
@@ -101,14 +101,24 @@ func (s *session) Status() flows.SessionStatus { return s.status }
...
 	}
-
 	return types.NewXObject(run.RootContext(s.env))
 }
 
+// looks through this session's run for the one that was last modified
+func (s *session) currentRun() flows.FlowRun {
+	var lastRun flows.FlowRun
+	for _, run := range s.runs {
+		if lastRun == nil || run.ModifiedOn().After(lastRun.ModifiedOn()) {
```
nicpottier: Just thinking aloud here. Last modified will be parent in the case of a child exiting and returning to parent?

rowanseymour: yeah we always go back thru the stack - added a subflow test to verify this

PR #835, opened 2020-01-22
⚖️ Template size limit

rowanseymour:
Limit size of evaluated templates to a configurable number of characters


flows/actions/call_resthook.go
```diff
@@ -88,8 +88,8 @@ func (a *CallResthookAction) Execute(run flows.FlowRun, step flows.Step, logModi
 		return nil
 	}
 
-	// build our payload
-	payload, err := run.EvaluateTemplate(ResthookPayload)
+	// build our payload (not truncated)
+	payload, err := run.EvaluateTemplateText(ResthookPayload, nil, false)
```
rowanseymour: so body for `call_webhook` and `call_resthook` are the two exceptions to the template truncation.. go as big as you want.

We control the resthook payload, and we truncate the webhook request/responce traces before writing them to the event.

PR #836, opened 2020-01-22
Stringify contactql like the queries they came from

nicpottier:
So we don't really know types of the values here, they are strings, which means we aren't simplifying quotations like we do on the RapidPro side.

A few options:
 1. Say we want to train users to use quotes everywhere and leave it as is
 2. I can look at values and try to parse them as decimals, if successful, then use unquoted, this would make numeric queries stay as `age = 14` instead of `age = "14"`
 3. Look at the value and figure out if it can be unquoted, this would basically be checking whether quoted and unquoted are the same save surrounding quotes (ie, does it have escaped chars) and whether it has whitespace. This would make `gender = M` be `gender = "M"`

Opinions @rowanseymour ?


rowanseymour:
Number 2. sounds most correct to me and that's what we've been doing in Python https://github.com/nyaruka/rapidpro/blob/ddbc43592635482111bc788a1c15ad3279b22cd2/temba/contacts/search/parser.py#L155. `age = 38 AND gender = "M"`

nicpottier:
updated @rowanseymour

PR #838, opened 2020-01-23
🐒 Ignore content-type headers and try to parse all webhook responses as JSON

### LINKED ISSUE #837: Can't fetch webhook responses of type application/geo+json
We only fetch bodies which have an explicit content type from an accepted list, or if there's no content-type header, we use a library to try and sniff it. Can't currently make requests to https://api.weather.gov/ because `application/geo+json` isn't in our list.

We could..

1. Keeping adding types to the list as we find them
2. Sniff any response even we don't recognize the content-type
3. Download all bodies and try to parse as JSON

rowanseymour:
Fixes #837


services/webhooks/service.go
```diff
@@ -105,9 +105,10 @@ func (s *service) newCallFromResponse(requestTrace []byte, response *http.Respon
 		return nil, err
 	}
 
-	if body != nil {
-		w.Response = append(w.Response, body...)
-		w.ResponseBody = body
+	w.Response = append(w.Response, body...)
```
rowanseymour: even if we ignore a non-JSON body, it should still appear in the response trace of event

PR #841, opened 2020-01-24
🔥 Fix problems with contact searching and add support for URN as attribute

rowanseymour:


flows/contact.go
```diff
@@ -369,6 +370,11 @@ func (c *Contact) ReevaluateDynamicGroups(env envs.Environment, allGroups *Group
 func (c *Contact) QueryProperty(env envs.Environment, key string, propType contactql.PropertyType) []interface{} {
 	if propType == contactql.PropertyTypeAttribute {
 		switch key {
+		case contactql.AttributeID:
```
rowanseymour: this was missing.. I guess cuz we previously only used this code for dynamic groups which don't support id

flows/contact.go
```diff
@@ -388,7 +401,7 @@ func (c *Contact) QueryProperty(env envs.Environment, key string, propType conta
 		urnsWithScheme := c.urns.WithScheme(key)
 		vals := make([]interface{}, len(urnsWithScheme))
 		for i := range urnsWithScheme {
-			vals[i] = string(urnsWithScheme[i].URN())
+			vals[i] = urnsWithScheme[i].URN().Path()
```
rowanseymour: this was wrong.. not sure how we hadn't noticed until now.. other than people don't tend to make groups like `tel = X`

flows/contact.go
nicpottier: is that the problems?

rowanseymour: the problems?

nicpottier: referring to the PR name, but see the other one too.

flows/contact.go
nicpottier: Very edge case, but for the per-contact case, do we handle the minimum of length 3 for searching? IE, we build trigrams in our index for path (I think name too) so nothing shorter than 3 characters would ever match for those two.

rowanseymour: you've lost me.. where are we building trigrams?

nicpottier: Elastic can't do arbitrary substring searches of any length efficiently. IE, when you search "nic" to match "nicolas" it has to look in the inverted index for candidate hits which it then double checks. (otherwise it would have to scan every document)

The way it determines which are candidates is that for things that support substring searching you tell it the length of the n-gram to build. So for something like "nicolas" it builds the ngrams "nic", "ico", "col".. etc..

It does the same with the incoming query, finds the matches, then filters etc..

The consequence of that is for name and path searches we need at least three letters, otherwise we can't match in elastic. Perhaps this just needs to be a contactql restriction / search error for the `~` operator then evaluation is simpler.

rowanseymour: Ok sounds like parsing needs to be updated to disallow `~` with anything shorter than 3 characters: https://github.com/nyaruka/goflow/issues/842

PR #843, opened 2020-01-27
🔎 Match evaluation of contact queries in ES

rowanseymour:


contactql/parser_test.go
```diff
@@ -175,16 +227,11 @@ func TestEvaluateQuery(t *testing.T) {
 		// location field condition
 		{`state = kigali`, true},
 		{`state = "kigali"`, true},
-		{`state = "NY"`, false},
-		{`state ~ KIG`, true},
```
rowanseymour: we don't support `~` for fields.. but is there maybe some ES indexing for them? Am confused by https://github.com/nyaruka/rp-indexer/blob/master/indexer.go#L424

contactql/visitor.go
```diff
@@ -75,31 +75,43 @@ func (v *visitor) VisitImplicitCondition(ctx *gen.ImplicitConditionContext) inte
...
-		return newCondition(PropertyTypeScheme, urns.TelScheme, "~", value, assets.FieldTypeText)
+		return newCondition(PropertyTypeScheme, urns.TelScheme, ComparatorContains, value, assets.FieldTypeText)
 	}
 
-	return newCondition(PropertyTypeAttribute, AttributeName, "~", value, attributes[AttributeName])
+	// convert to contains condition only if we have the right tokens, otherwise make equals check
+	comparator := ComparatorContains
+	if len(tokenizeNameValue(value)) == 0 {
+		comparator = ComparatorEqual
+	}
```
rowanseymour: So effectively if user types `X`, instead of erroring that `X` is too short to make `name ~ "X"`, we rewrite query as `name = "X"`

utils/text_test.go
```diff
@@ -72,12 +72,57 @@ func TestTokenizeStringByChars(t *testing.T) {
 	}
 }
 
+func TestTokenizeStringByUnicodeSeg(t *testing.T) {
+	tokenizerTests := []struct {
+		text   string
+		result []string
+	}{
+		// test cases taken from Apache Lucene (TestStandardAnalyzer.java) to assert equivalency
```
rowanseymour: Our ElasticSearch implementation of name searching uses the `StandardTokenizer` from Apache Lucene (https://github.com/apache/lucene-solr/blob/1d85cd783863f75cea133fb9c452302214165a4d/lucene/core/src/java/org/apache/lucene/analysis/standard/StandardTokenizerImpl.java) so added some of it's unit tests here to see how close we are.

contactql/parser_test.go
```diff
@@ -240,13 +287,10 @@ func TestEvaluationErrors(t *testing.T) {
 		query  string
 		errMsg string
 	}{
-		{`gender > Male`, "can't query text fields with >"},
```
rowanseymour: these are parser errors now

contactql/parser_test.go
nicpottier: Ooh, good catch. So this is basically indexing only the last part of something that looks like `Rwanda > Kigali > Kimihurura` .. so that you can query just `stat = Kimihurura`.

Since fields always contain the full path `Rwanda > Kigali > Kimihurura` then that's needed to allow that kind of querying. To tokenizer does not tokenize the remaining words though, so "Dakota" will not match "South Dakota" and "North Dakota".

So this is still a `=` but only on the leaf of the location.

rowanseymour: From rp-indexer.. 

```
"tokenizer": {
  "location_tokenizer": {
    "type": "pattern",
    "pattern": "(.* > )?([^>]+)",
    "group": 2
  },
  "trigram": {
    "type" : "ngram",
    "min_gram" : 3,
    "max_gram" : 3
  }
}
```

Ok to clarify..

1. The trigram stuff isn't being used? We can drop that in rp-indexer?
2. If a contact has `state` value of `Rwanda > Kigali > South Kimihurura` it would tokenize to `South Kimihurura` as a _single_ token, which would match if the user searched for exactly `state = "South Kimihurura"`

Shouldn't the tokenization go both ways? Looks like you can't search for something like `state = "Rwanda > Kigali > South Kimihurura"`.. or maybe my location data isn't right..

nicpottier: 1. Trigram tokenizer is being used for URNs, tabs are messed up in that file so it is hard to tell but that tokenizer is used for the trigram analyzer.
2. Correct.

So I was actually wrong, but wrong twice, so correct! I love it when that happens.

Turns out when we query, we query against the `_keyword` version of the field (both in Python and Golang):
    https://github.com/nyaruka/mailroom/blob/master/search/search.go#L224

This is not tokenized in any way, just like text fields it is just the full thing. So you would expect that to contain `Rwanda > Kigali > Kimihurura`, except that the indexer itself takes care of the trimming to just the appropriate level:
  https://github.com/nyaruka/rp-indexer/blob/master/indexer.go#L363

So what is indexed is indeed just `Kimihurura` and it is an exact term query there.

So no there's no way to type in the whole thing. I think we decided that if you had multiple wards with the same name it was more intuitive to do "state = kigali and district = kimihurura" then have people try to figure out the whole `Rwanda > Kigali > Kimihurura` nonesense.

rowanseymour: Ah ok and it seems `Contact.QueryProperty` already returns `Kimihurura` for `Rwanda > Kigali > Kimihurura` so looks like this all works fine already

Issue #845, opened 2020-01-29
Webhook time taken values don't include reading body


rowanseymour:
It's just the time taken to get the response headers which is misleading

PR #846, opened 2020-01-29
🧺 Webhook fixes and cleanup

### LINKED ISSUE #814: Error when a webhook is too large should include URL of webhook

rowanseymour:
This was not meant to end up looking like a refactor but in the course of trying to fix #814 I figured it was worth merging the two different code paths we have for making traced HTTP calls (webhooks vs everything else) into one, and then also noticed #845, and had to make some changes to the testing layer to allow easier testing of big payloads.


flows/actions/testdata/call_webhook.json
```diff
@@ -469,5 +469,72 @@
...
+                "step_uuid": "59d74b86-3e2f-4a93-aece-b05d2fdcde0c",
+                "text": "webhook response body exceeds 100000 bytes limit",
+                "type": "error"
+            },
+            {
+                "created_on": "2018-10-18T14:20:30.000123456Z",
+                "elapsed_ms": 0,
+                "request": "GET / HTTP/1.1\r\nHost: temba.io\r\nUser-Agent: goflow-testing\r\nAccept-Encoding: gzip\r\n\r\n",
+                "response": "HTTP/1.0 200 OK\r\nContent-Length: 100010\r\n\r\n",
+                "status": "connection_error",
```
rowanseymour: should this be a different status.. like `download_error`? These statuses are only used for event logging

services/airtime/dtone/client_test.go
```diff
@@ -19,15 +19,15 @@ func TestClient(t *testing.T) {
...
-			httpx.NewMockResponse(200, "info_txt=pong\r\n", nil),                  // successful ping
-			httpx.NewMockResponse(400, "error_code=1\r\nerror_txt=Oops\r\n", nil), // unsuccessful ping
-			httpx.NewMockResponse(200, withCRLF(msisdnResponse), nil),             // successful msdninfo query
-			httpx.NewMockResponse(200, "xxx=yyy\r\n", nil),                        // unexpected response to msdninfo query
-			httpx.NewMockResponse(200, withCRLF(reserveResponse), nil),            // successful reserve ID request
-			httpx.NewMockResponse(200, "xxx=yyy\r\n", nil),                        // unexpected response to reserve ID request
-			httpx.NewMockResponse(200, withCRLF(topupResponse), nil),              // successful topup request
-			httpx.NewMockResponse(200, "xxx=yyy\r\n", nil),                        // unexpected response to topup request
-			httpx.MockConnectionError,                                             // timeout
+			httpx.NewMockResponse(200, nil, "info_txt=pong\r\n", 1),                  // successful ping
```
rowanseymour: half of this diff is just adding a new param to mock responses to repeat the body - needed a JSON serializable way to make big payloads

services/webhooks/service.go
```diff
@@ -2,31 +2,13 @@ package webhooks
...
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/flows/engine"
-	"github.com/nyaruka/goflow/utils/dates"
 	"github.com/nyaruka/goflow/utils/httpx"
-
-	"github.com/pkg/errors"
 )
 
-// response content-types that we'll fetch
-var fetchResponseContentTypes = map[string]bool{
```
rowanseymour: no longer used

services/webhooks/mock.go
```diff
@@ -1,55 +0,0 @@
-package webhooks
```
rowanseymour: no longer used - mocking happens at the `httpx` level below

flows/actions/call_webhook.go
```diff
@@ -157,15 +157,15 @@ func (a *CallWebhookAction) Results(node flows.Node, include func(*flows.ResultI
...
 
 // determines the webhook status from the HTTP status code
-func callStatus(call *flows.WebhookCall, isResthook bool) flows.CallStatus {
-	if call.StatusCode == 0 {
+func callStatus(call *flows.WebhookCall, err error, isResthook bool) flows.CallStatus {
+	if call.Response == nil || err != nil {
 		return flows.CallStatusConnectionError
 	}
-	if isResthook && call.StatusCode == 410 {
+	if isResthook && call.Response.StatusCode == 410 {
```
nicpottier: might as well use the constant here while you are at it.

services/airtime/dtone/client_test.go
nicpottier: Seems having a different constructor may have made that easier no?

rowanseymour: yup!

Issue #840, opened 2020-01-24
ContactQL implicit parsing of US phone numbers goofy


nicpottier:
Typical formatting is `(206) 555-1212` that turns into `(tel ~ 206 AND tel ~ 555) AND tel ~ 1212`, which does match but is... weird.

PR #847, opened 2020-01-29
☎️ Phone number parsing fixes

rowanseymour:
* Fix `has_phone` so that it finds phone numbers with text after them
 * Rewrite queries which are formatted telephone numbers as `tel = ` (#840)


contactql/visitor.go
```diff
@@ -14,8 +14,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-var telRegex = regexp.MustCompile(`^[+ \d\-\(\)]{4,}$`)
-var cleanSpecialCharsRegex = regexp.MustCompile(`[+ \-\(\)]+`)
+// a query like (800) 345-3536 or (800)4567890 will be interpreted as a single tel ~ condition
+var queryIsPhoneNumberRegex = regexp.MustCompile(`^\(\d{3}\)\s*\d{3}\-?\d{4}$`)
```
rowanseymour: not sure how loose this should be.. can tweak (see https://regex101.com/r/sVXW3U/1)

nicpottier: There are a lot of different formats used across countries, should we just be trying to parse as a phone number using the phone numbers library and if it thinks it is a possible number then use a `tel` query? Gotcha there would be that we need country code to make educated guesses there.

rowanseymour: So we're trying to identify phone numbers in two places now:

1. The whole query is a formatted phone number `(800) 123-4567` This PR is basically just trying to strip out the `(` `)` and spaces before the get interpreted as query syntax
2. An implicit condition is part of a phone number `bob 9799` => `name = "Bob" AND tel ~ 9799`

if 1. is always a full phone number, then maybe we should the phone number library to convert it to an exact match, i.e. `tel = +18001234567` on the normalized form? I assume we always have access to country code as we already need access to whether URNs are redacted.

nicpottier: Good point, if we can parse it as a full number then no reason not to do equality there.

rowanseymour: @nicpottier having difficulty getting phonenumbers to parse in a way that means the _entire input is a phone number_, i.e. for queries like `tel = +250788383383` it just swallows the `tel = ` part.

Have also noticed that parsing isn't quite right for the `has_phone` router case.. it swallows input before the number but not after it, i.e. `+250788383383 is my number` doesn't parse. I think maybe we need to implement the `findNumbers` method from the Java version of the library for that.

nicpottier: Ya, the library only eats extra characters before the number. Should we gate the parsing attempt to just things that are numeric + `.-()+`?

rowanseymour: Can play with that.

What's the story with `findNumbers`? Seems like `has_phone` should match `+250788383383 is my number`

nicpottier: As you said, not implemented. Would have to implement a go version of: https://github.com/google/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberMatcher.java

rowanseymour: Hmm well that doesn't sound like fun... maybe we can approximate that by trimming non-tel chars from either end of the string

rowanseymour: @nicpottier have gone with a regex to identify "candidates" that could-maybe-possibly be phone numbers which it then hands to the library for real parsing. Please take another look.

PR #850, opened 2020-02-04
🕵️‍♀️ Missing dependencies part 2

rowanseymour:
Adds `node_uuids` to each dependency same as we have on results


flows/inspect.go
```diff
@@ -77,16 +112,6 @@ func checkDependency(sa SessionAssets, ref assets.Reference) bool {
 	}
 }
 
-// derives a dependency type name (e.g. group) from a reference
-func referenceTypeName(ref assets.Reference) string {
```
rowanseymour: forgot that references already have type names.. don't need this

PR #851, opened 2020-02-05
🚧 Missing dependencies part 3

rowanseymour:
Update actions to log error events when dependencies are missing


nicpottier:
So am I reading that right that we don't even require a contact now? That feels like slightly too big a swing towards dynamic to me, what's your reasoning there?

rowanseymour:
I didn't change anything related to that in this PR... there are a couple of actions which work without contact being set...and some that error... and in theory you can have a session without a contact but we've never really tested that. I would happily remove that in the name of simplicity.

nicpottier:
Oh I see, that test case just moved in the file, thought it was new. I suppose if we've gotten this far no reason to change it now!

rowanseymour:
Smarter versions of ourselves thought it was good to build something more generic 🤷‍♂

PR #853, opened 2020-02-07
🌎 Only send using tel channels of different country if explicitly allowed

rowanseymour:


flows/channel.go
```diff
@@ -116,7 +116,9 @@ func (s *ChannelAssets) GetForURN(urn *ContactURN, role assets.ChannelRole) *Cha
 		candidates := make([]*Channel, 0)
 
 		for _, ch := range s.all {
-			if ch.HasRole(role) && ch.SupportsScheme(urns.TelScheme) && (countryCode == "" || countryCode == ch.Country()) && !ch.HasParent() {
+			countryAllowed := countryCode == "" || countryCode == ch.Country() || ch.AllowInternational()
+
+			if ch.HasRole(role) && ch.SupportsScheme(urns.TelScheme) && countryAllowed && !ch.HasParent() {
```
nicpottier: minor / bikesheddy, kinda feels like this would be slightly more readable with `ifs` that disqualified a channel with a continue.. but totally whatever.

```golang
// not tel and not sendable, continue
if !ch.SupportsScheme(urns.TelScheme) || !ch.hasRole(role){
  continue
}
// international and we don't send there, continue
if (countryCode != "" && countryCode != ch.Country() && !ch.AllowInternational()){
  continue
}
...
```

PR #855, opened 2020-02-11
👓 List dependencies with action UUID and presence on router

rowanseymour:


flows/actions/testdata/add_contact_groups.json
```diff
@@ -48,9 +48,11 @@
                     "name": "Climbers",
                     "type": "group",
                     "missing": true,
-                    "node_uuids": [
-                        "72a1f5df-49f9-45df-94c9-d86f7ea064e5"
-                    ]
+                    "nodes": {
```
rowanseymour: for consistency I think I want to switch inspected results to use this instead of a list of `node_uuids` as well

PR #858, opened 2020-02-13
⚠️ Find problems during inspection

rowanseymour:


flows/actions/testdata/add_contact_groups.json
```diff
@@ -47,17 +47,23 @@
...
-                        ]
-                    }
+                    "missing": true
                 }
             ],
             "results": [],
             "waiting_exits": [],
-            "parent_refs": []
+            "parent_refs": [],
+            "problems": [
```
rowanseymour: I prefer `problem` to `warning`.. lots of warnings about things in the app already.. `problem` can have specific meaning and possibly db tables one day

flows/actions/testdata/add_contact_groups.json
```diff
@@ -47,17 +47,23 @@
...
+                {
+                    "type": "missing_dependency",
+                    "node_uuid": "72a1f5df-49f9-45df-94c9-d86f7ea064e5",
+                    "action_uuid": "ad154980-7bf7-4ab8-8728-545fd6378912",
```
rowanseymour: we're recording one problem per problem.. so describing location in the flow gets a bit easier

flows/actions/testdata/add_contact_groups.json
```diff
@@ -47,17 +47,23 @@
...
+                    "missing": true
                 }
             ],
             "results": [],
             "waiting_exits": [],
-            "parent_refs": []
+            "parent_refs": [],
+            "problems": [
+                {
+                    "type": "missing_dependency",
```
rowanseymour: I'm really not convinced we should be having the engine to render localizable descriptions here - seems like the same problem as rendering events in the simulator. Editor is going to want to render these richly with links to things.. emphasized words etc.

flows/actions/testdata/call_classifier.json
```diff
@@ -60,7 +55,18 @@
...
             "waiting_exits": [],
-            "parent_refs": []
+            "parent_refs": [],
+            "problems": [
+                {
+                    "type": "missing_dependency",
+                    "node_uuid": "72a1f5df-49f9-45df-94c9-d86f7ea064e5",
+                    "action_uuid": "ad154980-7bf7-4ab8-8728-545fd6378912",
+                    "reference": {
+                        "uuid": "63998ee7-a7a5-4cc5-be67-c773e1b6b9b1",
```
nicpottier: Don't we need the type of the reference somewhere?

rowanseymour: Added now

rowanseymour:
@nicpottier I'm not averse to a default description in English which I guess the editor can use if it ever sees a problem it doesn't recognize, or we might log somewhere else. There was a talk on the call of a templatized thing that the editor could localize which I talked myself out of.

Another reason to not call these things _warnings_ is we might give them severity levels at some point, some might be warnings (deprecated context use), some might errors (missing deps)

flows/actions/testdata/add_contact_groups.json
rowanseymour: Have added description

nicpottier:
I think my take on problems is just are we going to use that language in the UI? I tend to hate naming things in ways that are different than how we refer to them to users because then we constantly have to be translating in our head.

Telling a user there are "problems" on their flow sounds weird to me. Telling them there are "warnings" on their flow seems more right. Perhaps there could be severity, but even then I just find the language around problem a bit weird.

@ericnewcomer any feelings on what the UI is going to call these?

nicpottier:
(also just so I understand your reservation, where else do we use warnings?)

rowanseymour:
My take is that this is like when you wanted to create a Django model called `Model` but instead we came up with `Classifier` and since that's a new word in our glossary, it takes a little getting used to, but now you know what it means when you see/hear it without context. I also don't understand the issue with surfacing these as `Problems` in the UI. It leaves the door open to having problems of different severities etc. But anyways it's easy to change if I'm outvoted here.

nicpottier:
Ya, I see where you are coming from. Problem just sounds super weird to me
from a users perspective. It feels overly negative. A system telling you
your flow has problems just sounds a lot stronger / negative to me than it
saying your flow has some warnings.

I feel warnings can have severity no differently than problems.

Eric can choose because users!

nicpottier:
What about Issues?

rowanseymour:
I like it!

ericnewcomer:
In my head these were alerts, warnings or notifications, but didn't love any of them. Issues works just fine.

Issue #857, opened 2020-02-13
Query parsing should allow implicit parts to be quoted


rowanseymour:
E.g. https://sentry.io/organizations/nyaruka/issues/1511290329/?referrer=alert_email

PR #859, opened 2020-02-13
📇 Contact query parsing tweaks 

rowanseymour:
* Allow implicit conditions to be quoted, e.g. whole query is `"bobby"` (https://github.com/nyaruka/goflow/issues/857)
* Don't strip `+` when cleaning up phone number looking values, i.e. so you can search by intl code


contactql/parser_test.go
```diff
@@ -86,7 +88,7 @@ func TestParseQuery(t *testing.T) {
 
 		// implicit combinations
 		{`will felix`, `name ~ "will" AND name ~ "felix"`, "", envs.RedactionPolicyNone},
-		{`will +123456566`, `name ~ "will" AND tel ~ 123456566`, "", envs.RedactionPolicyNone},
+		{`will +123456566`, `name ~ "will" AND tel ~ +123456566`, "", envs.RedactionPolicyNone},
```
nicpottier: Feels slightly weird to not have the number quoted when it leads with a `+` but I guess I'm fine with it.

rowanseymour: Was thinking same thing - would be easy to tweak that

rowanseymour: Updated

PR #864, opened 2020-02-18
👮 Add issue to report on usage of @legacy_extra

rowanseymour:
Originally thought about a more generic issue type, like "deprecated_context_reference".. but I think it's better to be specific here. We'd want to render this in the editor along with an explanation of how the expression should be changed (i.e. use `@trigger.params` or `@webhook`)


nicpottier:
Ya, how we explain the "fix" is the tough one here.

rowanseymour:
@ericnewcomer not gonna merge this just yet - but figured it was good to try implementing another issue type and we can have a think about how this would look in the editor

rowanseymour:
@nicpottier @ericnewcomer issues could have associated articles that we link to.. tho obviously that would have to be configurable/optional because not every RP deployment has support pages.

ericnewcomer:
I've wanted to add the ability for the editor to link into help (if configured to have it), this would be a cool opportunity to try that out.

nicpottier:
Wait, so what are we telling people in the UI about this? I really don't want a flurry of these appearing without really nailing how they should fix it.

rowanseymour:
Not sure what the current editor handling is but seems reasonable for us to say that every issue has a corresponding article. 

It does make me wonder tho if the current tomato red for everything styling is a bit strong for some of these things which are.. "hey it'd be great if you weren't doing this, but your flow will work fine".

nicpottier:
Ya, I just feel like this particular issue is one we shouldn't be exposing just yet until we have a better story on how we are guiding people through these and have some experience how people respond to them. An awful lot of flows are going to have this issue and we really don't want people to wig out about it.

rowanseymour:
Can we just ignore them on the editor side for now? I'm assuming at some point we want to either start saving these issues in the DB or at least saving aggregate statistics about them so it's possible to start calculating things "how many flows are doing X"

ericnewcomer:
This issue is why I added the intercom stuff yesterday.

Issue #865, opened 2020-02-18
Group modifier should disallow adding to dynamic groups


nicpottier:

rowanseymour:
So as we've used modifiers to date, they have no concept of errors because we assume that the action validated all the inputs and turned errors into error events.

We have a few options here.. suppose the engine gets a `groups` modifier which has 2 groups and one of them is dynamic, it could..

1. Return an actual `error` making no changes to the contact
2. Ignore the dynamic group, log `contact_groups_changed event` for the other group
3. Log error event for the dynamic group, log `contact_groups_changed event` for the other group

Note that in the case of a dynamic group with a query error, modifiers can log error events

nicpottier:
3) feels the most right to me.

rowanseymour:
Yeah I think so too after I noticed they already do log error events in some cases

PR #866, opened 2020-02-18
👪 Groups modifier should log error events for dynamic groups

rowanseymour:
* Move the check for dynamic groups in add/remove actions to the modifier (#865)
 * Make modifier tests snapshot based


flows/actions/base.go
```diff
@@ -297,11 +297,7 @@ func resolveGroups(run flows.FlowRun, references []*assets.GroupReference, stati
 		}
 
 		if group != nil {
-			if staticOnly && group.IsDynamic() {
```
rowanseymour: nicely there's no duplication required.. we can just move this down to the modifier

flows/actions/modifiers/testdata/channel.json
```diff
@@ -58,11 +58,11 @@
         "contact_after": {
             "uuid": "5d76d86b-3bb9-4d5a-b822-c9d86f5d8e4f",
             "name": "Bob",
+            "created_on": "2018-06-20T11:40:30.123456789Z",
             "urns": [
-                "tel:+12065551212?channel=3a05eaf5-cb1b-4246-bef1-f277419c83a7&id=123",
+                "tel:+12065551212?channel=3a05eaf5-cb1b-4246-bef1-f277419c83a7\u0026id=123",
```
nicpottier: I was seeing this too.. for our generic tests should we go the extra mile and do: https://github.com/golang/go/commit/ab52ad894f453a02153fb2bc106d08c47ba643b6

rowanseymour: I'm confused.. all of this JSON is marshaled by goflow's own encoder which disables at https://github.com/nyaruka/goflow/blob/master/utils/jsonx/json.go#L38 .. in theory

nicpottier: Oh hah (and hrmm)

rowanseymour: Fixed now - had to call the custom `jsonx.Marshal` all down the chain of marshals involved in marshalling a contact

PR #867, opened 2020-02-18
📇 Allow urns modifier to remove a URN

rowanseymour:
Contains #866 which should be merged first


flows/actions/modifiers/urn.go
```diff
@@ -45,8 +46,16 @@ func NewURN(urn urns.URN, modification URNModification) *URNModifier {
 
 // Apply applies this modification to the given contact
 func (m *URNModifier) Apply(env envs.Environment, assets flows.SessionAssets, contact *flows.Contact, log flows.EventCallback) {
-	contactURN := flows.NewContactURN(m.URN.Normalize(string(env.DefaultCountry())), nil)
-	if contact.AddURN(contactURN) {
+	urn := m.URN.Normalize(string(env.DefaultCountry()))
```
nicpottier: ooh!

nicpottier: Very edge case-y, but should we be looking at the country for the channel the URN is associated with if it has one?

rowanseymour: That's a weird one.. cuz potentially a local number could then match against more than one tel URN.. I'm inclined to say this modifier expects something that's already normalized or in the org country.

nicpottier: It's weird enough that I'm fine taking a wait and see approach.

PR #870, opened 2020-02-20
🔍 Group conditions and fixes in contact queries

rowanseymour:
* Add `group` as an attribute meaning "any group"
* Fix != with multiple values (e.g. multiple URNs or groups)
* Fix `id` conditions being allowed with `>` `>=` `<` `<=`

### DIFF
--- a/contactql/parser.go
+++ b/contactql/parser.go
@@ -147,18 +147,28 @@ func (c *Condition) Evaluate(env envs.Environment, queryable Queryable) (bool, e
 		return false, nil
 	}
 
-	// check each resolved value
+	// evaluate condition against each resolved value
+	anyTrue := false
+	allTrue := true
 	for _, val := range vals {
 		res, err := c.evaluateValue(env, val)
 		if err != nil {
 			return false, err
 		}
 		if res {
-			return true, nil
+			anyTrue = true
+		} else {
+			allTrue = false
 		}
 	}
 
-	return false, nil
+	// foo != x is only true if all values of foo are not x
+	if c.comparator == ComparatorNotEqual {
+		return allTrue, nil
+	}
+
+	// foo = x is true if any value of foo is x
+	return anyTrue, nil
 }
 
 func (c *Condition) evaluateValue(env envs.Environment, val interface{}) (bool, error) {
--- a/contactql/parser_test.go
+++ b/contactql/parser_test.go
@@ -51,6 +51,11 @@ func TestParseQuery(t *testing.T) {
 		{`name != "felix"`, `name != "felix"`, "", envs.RedactionPolicyNone}, // is not equal to value
 		{`Name ~ ""`, ``, "value must contain a word of at least 2 characters long for a contains condition on name", envs.RedactionPolicyNone},
 
+		// explicit attribute conditions
+		{`language = spa`, `language = "spa"`, "", envs.RedactionPolicyNone},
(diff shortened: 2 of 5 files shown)


contactql/parser.go
```diff
@@ -147,18 +147,28 @@ func (c *Condition) Evaluate(env envs.Environment, queryable Queryable) (bool, e
...
+			anyTrue = true
+		} else {
+			allTrue = false
 		}
 	}
 
-	return false, nil
+	// foo != x is only true if all values of foo are not x
+	if c.comparator == ComparatorNotEqual {
+		return allTrue, nil
```
rowanseymour: need to check the ES implementation handles this the same..

contactql/parser_test.go
```diff
@@ -111,6 +116,58 @@ func TestParseQuery(t *testing.T) {
...
+		{`name > felix`, ``, "comparisons with > can only be used with date and number fields", envs.RedactionPolicyNone},
+		{`language > eng`, ``, "comparisons with > can only be used with date and number fields", envs.RedactionPolicyNone},
+		{`group > reporters`, ``, "comparisons with > can only be used with date and number fields", envs.RedactionPolicyNone},
+		{`created_on > 20-02-2020`, `created_on > "20-02-2020"`, "", envs.RedactionPolicyNone},
+		{`tel > 02352`, ``, "comparisons with > can only be used with date and number fields", envs.RedactionPolicyNone},
+		{`urn > 02352`, ``, "comparisons with > can only be used with date and number fields", envs.RedactionPolicyNone},
+		{`age > 18`, `age > 18`, "", envs.RedactionPolicyNone},
+		{`gender > male`, ``, "comparisons with > can only be used with date and number fields", envs.RedactionPolicyNone},
+		{`dob > 20-02-2020`, `dob > "20-02-2020"`, "", envs.RedactionPolicyNone},
+		{`state > Pichincha`, ``, "comparisons with > can only be used with date and number fields", envs.RedactionPolicyNone},
```
rowanseymour: @nicpottier these new tests above look consistent with recent tweaks on the ES side?

contactql/visitor.go
```diff
@@ -31,14 +31,16 @@ const (
 	AttributeName      = "name"
 	AttributeLanguage  = "language"
 	AttributeURN       = "urn"
+	AttributeGroup     = "group"
 	AttributeCreatedOn = "created_on"
 )
 
 var attributes = map[string]assets.FieldType{
-	AttributeID:        assets.FieldTypeNumber,
+	AttributeID:        assets.FieldTypeText,
```
rowanseymour: shouldn't support `>` etc

contactql/parser.go
nicpottier: Oh interesting.. pretty sure it should but not positive.

contactql/parser_test.go
nicpottier: I think so, but as I see this I wonder if we should just move the `search` package into goflow so these things are closer together. It doesn't really depend on anything within mailroom and having them both in goflow would allow for more natural testing etc.. King of like the providers for email etc, having a standard implementation of contactql to elastic in goflow seems totally reasonable.

rowanseymour: Yeah that doesn't sound terrible. I've been wracking my brain thinking of ways to enforce consistency between these two evaluator implementations.. and if we moved the ES stuff into goflow.. I think we could have some tests that assert consistency

Issue #861, opened 2020-02-17
Missing assets when creating contacts


rowanseymour:
This appears to be just a contact who has an inactive group in their group list in the db.. https://sentry.io/organizations/nyaruka/issues/1518121919/activity/?project=227973&referrer=alert_email

I think we convert that to report as a missing asset instead of being an error

PR #872, opened 2020-02-20
💥 Handle missing groups on contact creation

rowanseymour:
Addresses #861

### DIFF
--- a/flows/contact.go
+++ b/flows/contact.go
@@ -46,20 +46,18 @@ func NewContact(
 	timezone *time.Location,
 	createdOn time.Time,
 	urns []urns.URN,
-	groups []assets.Group,
-	fields map[string]*Value) (*Contact, error) {
+	groups []*assets.GroupReference,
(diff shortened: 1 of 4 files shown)


flows/contact.go
```diff
@@ -46,20 +46,18 @@ func NewContact(
...
+	urnList, err := ReadURNList(sa, urns, missing)
 	if err != nil {
 		return nil, err
 	}
 
-	groupList, err := NewGroupListFromAssets(sa, groups)
-	if err != nil {
-		return nil, err
-	}
+	groupList := NewGroupList(sa, groups, missing)
```
rowanseymour: I think we've run into problems because mailroom can create a contact on its side that has groups which, because they are inactive, aren't in the session assets... so for goflow to be sure that the contact only has fields and groups etc that exist in the session assets, it takes references and tries to resolve them, and potentially reports them as missing.

PR #873, opened 2020-02-20
🐌 Remove contacts from broken groups

### LINKED ISSUE #856: ReevaluateDynamicGroups doesn't remove contacts from groups with errors
Opening an issue instead of just fixing it because maybe it requires discussion, but:
  https://github.com/nyaruka/goflow/blob/master/flows/contact.go#L342

Doesn't remove the contact from groups when there are errors in the query for a dynamic group.

Feels like it should, but perhaps this is on purpose?

rowanseymour:
Fixes #856

### DIFF
--- a/flows/contact_test.go
+++ b/flows/contact_test.go
@@ -1,6 +1,8 @@
 package flows_test
 
 import (
+	"encoding/json"
+	"io/ioutil"
 	"testing"
 	"time"
 
@@ -12,6 +14,7 @@ import (
 	"github.com/nyaruka/goflow/excellent/types"
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/flows/engine"
+	"github.com/nyaruka/goflow/flows/triggers"
 	"github.com/nyaruka/goflow/test"
 	"github.com/nyaruka/goflow/utils/jsonx"
 	"github.com/nyaruka/goflow/utils/uuids"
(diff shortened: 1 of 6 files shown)


flows/contact_test.go
```diff
@@ -186,97 +189,51 @@ func TestContactSetPreferredChannel(t *testing.T) {
...
-
-	males := test.NewGroup("Males", `gender="M"`)
-	tel1800 := test.NewGroup("Tel with 1800", `tel ~ 1800`)
-	twitterCrazies := test.NewGroup("Twitter Crazies", `twitter ~ crazy`)
-	groups := []*flows.Group{males, tel1800, twitterCrazies}
-
-	contact := flows.NewEmptyContact(session.Assets(), "Joe", "eng", nil)
-	contact.AddURN(urns.URN("tel:+12345678999"), nil)
+	for _, tc := range tests {
+		sa, err := engine.NewSessionAssets(source, nil)
```
rowanseymour: Rewriting these tests led me to realize something odd - session assets aren't tied to particular environment - I tried to use the same session assets for each test case.. but.. the parsed contact query of a group is cached on the group. and it is environment specific

That led me to try reworking the whole assets creation thing so that session assets are created with an environment, query parsing happens at group creation (no caching needed)... but then it wasn't obvious how that parse error gets back to the user... so reverted for now.

nicpottier: Is there ever a time in the real world where an environment is different for the same session assets? Not sure I understand. It is very very important to mailroom that Session assets are both thread safe and able to be reused on the same org so we should definitely discuss if that isn't the case.

rowanseymour: No.. this isn't something we do. But it does kinda annoy me that it's possible, but subtly doesn't work correctly because of the cached contact query. Would be nice to get rid of that little bit of caching.

PR #875, opened 2020-02-24
👮 Stricter contact query parsing

rowanseymour:
* Validate that groups exist 
 * Validate that language codes are valid
 * Disallow `= ""` except for name, language, fields and urns

### DIFF
--- a/contactql/parser_test.go
+++ b/contactql/parser_test.go
@@ -113,7 +113,8 @@ func TestParseQuery(t *testing.T) {
 			envs.RedactionPolicyNone,
 		},
 
-		{`xyz != ""`, "", "can't resolve 'xyz' to attribute, scheme or field", envs.RedactionPolicyNone},
+		{`xyz != ""`, ``, "can't resolve 'xyz' to attribute, scheme or field", envs.RedactionPolicyNone},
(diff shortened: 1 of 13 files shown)


contactql/parser_test.go
```diff
@@ -143,6 +145,19 @@ func TestParseQuery(t *testing.T) {
 		{`dob != 20-02-2020`, `dob != "20-02-2020"`, "", envs.RedactionPolicyNone},
 		{`state != Pichincha`, `state != "Pichincha"`, "", envs.RedactionPolicyNone},
 
+		// = "" supported for name, language, fields and urns
+		{`id = ""`, ``, "can't check whether 'id' is set or not set", envs.RedactionPolicyNone},
+		{`name = ""`, `name = ""`, "", envs.RedactionPolicyNone},
+		{`language = ""`, `language = ""`, "", envs.RedactionPolicyNone},
+		{`group = ""`, ``, "can't check whether 'group' is set or not set", envs.RedactionPolicyNone},
```
rowanseymour: there's definitely an argument for simplicity that we should just say things like `id != ""` always be true, but since these aren't currently allowed in mailroom, and the group one would be tricky given how we index user and system groups together, I'm making them errors here for now

PR #877, opened 2020-02-27
💇 Text trimming functions

rowanseymour:
Adds https://github.com/rapidpro/rapidpro/issues/1162

### DIFF
--- a/docs/functions.json
+++ b/docs/functions.json
@@ -1179,12 +1179,12 @@
         ]
     },
     {
-        "signature": "split(text, delimiters)",
-        "summary": "Splits `text` based on the given characters in `delimiters`.",
-        "detail": "Empty values are removed from the returned list.",
+        "signature": "split(text, [,delimiters])",
+        "summary": "Splits `text` into an array of separated words.",
+        "detail": "Empty values are removed from the returned list. There is an optional final parameter `delimiters` which\nis string of characters used to split the text into words.",
         "examples": [
             {
-                "template": "@(split(\"a b c\", \" \"))",
+                "template": "@(split(\"a b c\"))",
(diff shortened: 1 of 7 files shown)


docs/functions.json
```diff
@@ -1179,12 +1179,12 @@
         ]
     },
     {
-        "signature": "split(text, delimiters)",
-        "summary": "Splits `text` based on the given characters in `delimiters`.",
-        "detail": "Empty values are removed from the returned list.",
+        "signature": "split(text, [,delimiters])",
```
rowanseymour: There is a family of functions which tokenize text into words and can take a string of delimiter chars, i.e. `split`, `word`, `word_count`, `word_slice`.. figured for consistency `split` could let the delimiters be optional like the rest of them.

docs/functions.json
```diff
@@ -1357,6 +1357,51 @@
             }
         ]
     },
+    {
+        "signature": "trim(text, [,chars])",
```
rowanseymour: Was on the fence about whether we need `trim` but `trim_left(trim_right(" abc ", "+"), "+")` is a bit of a mouthful and there's little downside to one extra function

docs/functions.json
```diff
@@ -1357,6 +1357,51 @@
...
+                "output": "hello world"
+            },
+            {
+                "template": "@(trim(\"+123157568\", \"+\"))",
+                "output": "123157568"
+            }
+        ]
+    },
+    {
+        "signature": "trim_left(text, [,chars])",
```
rowanseymour: was also wondering if `trim_start` and `trim_end` might be nicer for non-Go programmers.. but no strong feelings

PR #880, opened 2020-03-03
🗺️ Override env country if contact has preferred channel with country

rowanseymour:
Addresses https://github.com/rapidpro/rapidpro/issues/1166

Things like excellent functions (`has_phone`) only have access to an `Environment`.. and know nothing about runs and sessions. Inside a run, when we have templates to evaluate, we create a special `RunEnvironment` which overrides things like timezone and language based on values on the current session contact.

### DIFF
--- a/cmd/flowrunner/main.go
+++ b/cmd/flowrunner/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
@@ -25,6 +26,7 @@ import (
 	"github.com/nyaruka/goflow/utils/jsonx"
 	"github.com/nyaruka/goflow/utils/uuids"
 
+	"github.com/buger/jsonparser"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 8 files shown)


cmd/flowrunner/main.go
```diff
@@ -43,7 +45,7 @@ const contactJSON = `{
 }
 `
 
-const usage = `usage: flowrunner [flags] <assets.json> <flow_uuid>`
+const usage = `usage: flowrunner [flags] <assets.json> [flow_uuid]`
```
rowanseymour: this is an unrelated change to make `flowrunner` easier to use by defaulting to first flow it finds in an assets file

test/testdata/runner/phone_number.json
```diff
@@ -0,0 +1,176 @@
...
+      "name": "Android Channel",
+      "address": "+17036975131",
+      "schemes": [
+        "tel"
+      ],
+      "roles": [
+        "send",
+        "receive"
+      ],
+      "country": "US"
```
rowanseymour: country will come from this channel and let user enter a local US number

PR #881, opened 2020-03-07
🛑 Allow HTTP services to be configured with a list of disallowed hosts

rowanseymour:
Includes webhooks and LUIS classifiers

### DIFF
--- a/utils/httpx/access.go
+++ b/utils/httpx/access.go
@@ -4,23 +4,32 @@ import (
 	"net"
 	"net/http"
 	"strings"
+	"time"
+
+	"golang.org/x/net/context"
 )
 
 // AccessConfig configures what can be accessed
 type AccessConfig struct {
-	DisallowedIPs []net.IP
+	ResolveTimeout time.Duration
+	DisallowedIPs  []net.IP
 }
 
 // NewAccessConfig creates a new access config
-func NewAccessConfig(disallowedIPs []net.IP) *AccessConfig {
-	return &AccessConfig{DisallowedIPs: disallowedIPs}
+func NewAccessConfig(resolveTimeout time.Duration, disallowedIPs []net.IP) *AccessConfig {
+	return &AccessConfig{
+		ResolveTimeout: resolveTimeout,
+		DisallowedIPs:  disallowedIPs,
+	}
 }
 
 // Allow determines whether the given request should be allowed
 func (c *AccessConfig) Allow(request *http.Request) (bool, error) {
 	host := strings.ToLower(request.URL.Hostname())
 
-	addrs, err := net.LookupIP(host)
+	ctx, _ := context.WithTimeout(context.Background(), c.ResolveTimeout)
+
+	addrs, err := net.DefaultResolver.LookupIPAddr(ctx, host)
 	if err != nil {
 		return false, err
 	}
@@ -28,7 +37,7 @@ func (c *AccessConfig) Allow(request *http.Request) (bool, error) {
 	// if any of the host's addresses appear in the disallowed list, deny the request
 	for _, addr := range addrs {
 		for _, disallowed := range c.DisallowedIPs {
-			if addr.Equal(disallowed) {
+			if addr.IP.Equal(disallowed) {
 				return false, nil
 			}
 		}
--- a/utils/httpx/access.go
+++ b/utils/httpx/access.go
@@ -27,7 +27,8 @@ func NewAccessConfig(resolveTimeout time.Duration, disallowedIPs []net.IP) *Acce
 func (c *AccessConfig) Allow(request *http.Request) (bool, error) {
 	host := strings.ToLower(request.URL.Hostname())
 
-	ctx, _ := context.WithTimeout(context.Background(), c.ResolveTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), c.ResolveTimeout)
+	defer cancel()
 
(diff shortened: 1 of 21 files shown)


utils/httpx/http_test.go
```diff
@@ -27,6 +28,38 @@ func TestNewTrace(t *testing.T) {
...
+	_, err = call("https://localhost/path")
+	assert.EqualError(t, err, "requests to host localhost are disallowed")
+
+	_, err = call("https://LOCALHOST:80")
+	assert.EqualError(t, err, "requests to host localhost are disallowed")
+
+	_, err = call("https://127.0.0.1")
+	assert.EqualError(t, err, "requests to host 127.0.0.1 are disallowed")
+
+	_, err = call("https://[::1]:80")
```
rowanseymour: I need to read up on IPv6 addresses but I assume that we need to do some normalization as same addresses can have slightly different representations

rowanseymour:
Huh didn't know about `localhost.me` - I should use that for casepro testing! 

On reflection.. I think even lists of IPs isn't going to cut it.. or at least it puts a bit of burden on deployments to know every possible routing to boxes they want to protect. I think the truly correct solution here is proxying. All user controlled requests need to made from outside of any "trusted" network zones.

And it seems that doesn't even require a code change in goflow. On the mailroom side you can configure the `http.Transport` with a proxy, or even just set the `HTTP_PROXY` env variable. Of course it does require maintaining a new box somewhere running a proxy but that doesn't sound terrible.

nicpottier:
I mean there aren't many network interfaces on boxes are there? Basically localhost and a real private IP in most cases. At least that's the case for ours. And restricting those does fully restrict everything that isn't explicitly open via either our load balancers tunneling in or our security groups.

It would be easy for say Mailroom to start and figure out those IPs then restrict them appropriately no?

rowanseymour:
I mean we could probably compile a list of IPs to black list.. and keep it up to date.. but letting users execute arbitrary HTTP requests from inside your network.. seems like something we should be avoiding, no?

nicpottier:
I don't think there's a list to keep up to date as it can be derived on startup right?

Don't really disagree that a proxy would be the ultimate solution but having gone that route in the past it isn't a piece of infrastructure I really want to introduce to our stack if we can help it, it adds an extra level of things that can go wrong and things to keep "up". Just feel that cure may be worse than the disease.

rowanseymour:
I think we're talking about different goals here... I don't think the goal here is blocking access just to localhost (which obviously has a known set of interfaces). I think the goal here preventing access to anything inside a deployment's private network. We don't know what else other deployments are hosting and where.

I can easily change this PR to be a list of IPs instead of hostnames... and if we feel that's an acceptable trade off to avoid having a separate proxy instance to maintain.. then ok.

I do think it would be worth us posting on the developers mailing list so that other deployments can have  a think about what makes sense for them. It sounds like they wouldn't even need a code change as go's http library uses the `HTTP_PROXY` env variable by default.

nicpottier:
Fair point that we would want to prevent originating traffic regardless of destination from the origin VPC.

A proxy would require n+1 instances running squid in a separate VPC which is only accepting connections from our source VPCs that have mailroom etc.. Have to have those monitored etc, so yes would be great but that isn't a small lift. 

Note that those would essentially have the same problem of localhost, etc, so would probably need to do some filtering there as well, but them being contained in a separate VPC would reduce what you could get at so would definitely be an improvement.

Guess another method would be to have separate mailroom instances that are further locked down in some way, so essentially they are the "proxies" in that case and are running in a VPC with less trust. But they would be on a network that has access to PG / Redis so not really reducing attack vector much there.

Seems to me that in either case we do want to restrict connections by IP but we may additionally want to force connections via a proxy. So I think this is necessary but perhaps not sufficient.

rowanseymour:
Ok I'll switch this over to be IP based, and we can continue to ponder on this!

utils/httpx/access.go
```diff
@@ -0,0 +1,37 @@
...
+// NewAccessConfig creates a new access config
+func NewAccessConfig(disallowedIPs []net.IP) *AccessConfig {
+	return &AccessConfig{DisallowedIPs: disallowedIPs}
+}
+
+// Allow determines whether the given request should be allowed
+func (c *AccessConfig) Allow(request *http.Request) (bool, error) {
+	host := strings.ToLower(request.URL.Hostname())
+
+	addrs, err := net.LookupIP(host)
```
nicpottier: I think we probably want to use https://golang.org/pkg/net/#Resolver.LookupIPAddr because otherwise there's no timeout on this and we can just hang for any numbers of reasons.

utils/httpx/access.go
```diff
@@ -0,0 +1,46 @@
...
+		ResolveTimeout: resolveTimeout,
+		DisallowedIPs:  disallowedIPs,
+	}
+}
+
+// Allow determines whether the given request should be allowed
+func (c *AccessConfig) Allow(request *http.Request) (bool, error) {
+	host := strings.ToLower(request.URL.Hostname())
+
+	ctx, _ := context.WithTimeout(context.Background(), c.ResolveTimeout)
```
nicpottier: Does this work like this? I thought you were meant to save `cancel` then immediately did a `defer cancel()` for contexts?

See: https://developer.squareup.com/blog/always-be-closing/

rowanseymour: oops.. never used a Context  before. Guess it works til the memory leak gets you

PR #884, opened 2020-03-17
⚙️ Rework group asset loading so that parsing is not deferred

rowanseymour:
Pre-requisite for #880. Currently we lazy parse and cache contact queries, but parsing a contact query requires an environment, so we end up using the run-environment of the first contact this gets called for... which might be a different country to subsequent contacts which also use the same session assets.

This PR reworks things so when we're loading assets, we parse group queries, and determine which groups are "broken".

### DIFF
--- a/flows/engine/assets.go
+++ b/flows/engine/assets.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"github.com/nyaruka/goflow/assets"
+	"github.com/nyaruka/goflow/envs"
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/flows/definition"
 	"github.com/nyaruka/goflow/flows/definition/migrations"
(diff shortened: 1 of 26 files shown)


flows/engine/assets.go
```diff
@@ -64,14 +65,17 @@ func NewSessionAssets(source assets.Source, migrationConfig *migrations.Config)
 		return nil, err
 	}
 
+	fieldAssets := flows.NewFieldAssets(fields)
+	groupAssets, _ := flows.NewGroupAssets(env, fieldAssets, groups)
```
rowanseymour: Not entirely sure what to do here.. this method can return the list of "broken" groups.. but what should we do with that? My inclination is that we ignore broken groups at this point - we should be a better job of indicating in the RP UI that a dynamic group is broken.

PR #885, opened 2020-03-20
Allow queries on URNs to check if they are set or not

norkans7:
Addresses https://github.com/rapidpro/rapidpro/issues/1170


rowanseymour:
Added some more contact query evaluation testing in https://github.com/nyaruka/goflow/commit/0db0a7ad5611a61b00395b1741f0c3d2c646d50e to make sure this is correct at that level

PR #889, opened 2020-03-25
🌐 PO File Support

rowanseymour:
This part is adding support for generating them, next would come updating flows using them


i18n/testdata/two_questions.en.po
```diff
@@ -0,0 +1,62 @@
...
+
+#: 615b8a0f-588c-4d20-a05f-363b0b4ce6f4/5ce6c69a-fdfe-4594-ab71-26be534d31c3/name:0,615b8a0f-588c-4d20-a05f-363b0b4ce6f4/78ae8f05-f92e-43b2-a886-406eaea1b8e0/name:0
+msgid "Other"
+msgstr ""
+
+#: 615b8a0f-588c-4d20-a05f-363b0b4ce6f4/2ab9b033-77a8-4e56-a558-b568c00c9492/name:0
+msgid "Pepsi"
+msgstr ""
+
+#: 615b8a0f-588c-4d20-a05f-363b0b4ce6f4/598ae7a5-2f81-48f1-afac-595262514aa1/name:0,615b8a0f-588c-4d20-a05f-363b0b4ce6f4/e97cd6d5-3354-4dbd-85bc-6c1f87849eec/quick_replies:0
```
rowanseymour: string appears in two places - a category name and a quick reply

i18n/testdata/two_questions.en.po
```diff
@@ -0,0 +1,62 @@
...
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"POT-Creation-Date: 2020-03-25 13:57+0000\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+
+#: 615b8a0f-588c-4d20-a05f-363b0b4ce6f4/d2a4052a-3fa9-4608-ab3e-5b9631440447/text:0
```
rowanseymour: flow-uuid/item-uuid/property-name:array-index

i18n/testdata/two_questions.en.po
nicpottier: Any reason not to use two lines here instead of a comma? (notice Django uses separate lines). Also wonder if it may be worth annotating with the flow name just for anybody who is peeping the file? (we would still use UUID but just for reference sake)

```
#: temba/api/v2/views.py:278 temba/orgs/models.py:2213
#: templates/orgs/org_manage_accounts.haml:220
msgid "Administrator"
msgstr ""
```

nicpottier: (just noticed the django version is a bit of a mix of newlines vs same line actually)

nicpottier: And just to be clear on the direction we decided, if "Red" appeared twice but was translated separately in the two places, then.. we add a `msgctxt`? Or we use the latest?

i18n/po.go
```diff
@@ -0,0 +1,186 @@
+package i18n
```
nicpottier: Very go of you to write this from scratch. :)

rowanseymour: I actually did shop around but didn't love any of the options.. and figured the subset we need isn't that complex 🤷‍♂

i18n/testdata/two_questions.en.po
rowanseymour: Didn't know you could put them on separate lines.. guessing that's different versions of `xgettext` doing that..  certainly easier to read on separate lines so I can update that if poedit can handle it

rowanseymour: And re second question, we did decide that we include context where it's needed to separate differing translations of the same base - and that in theory the translator could fix that themselves if they decided it was unintentional. Making a test file for that scenario currently.

i18n/po.go
nicpottier: To be clear I did the same and agree with your decision. :)

i18n/testdata/two_questions.en.po
rowanseymour: Switched to using flow name as part of the reference - but URL encoding it to avoid problems with spaces. I think it's fine to use name over UUID here - these aren't real links.. you can't click on them to jump into the source code.. so they're just tips for the translator.

nicpottier: Just tried and POEdit does preserve dual comments in generated translation files.

i18n/testdata/translation_mismatches.noargs.es.po
```diff
@@ -0,0 +1,41 @@
...
+
+#: Translated/e42deebf-90fa-4636-81cb-d247a3d3ba75/quick_replies:0
+msgctxt "e42deebf-90fa-4636-81cb-d247a3d3ba75/quick_replies:0"
+msgid "Red"
+msgstr "Rojo"
+
+#: Translated/d1ce3c92-7025-4607-a910-444361a6b9b3/name:0
+msgctxt "d1ce3c92-7025-4607-a910-444361a6b9b3/name:0"
+msgid "Red"
+msgstr "Roja"
```
rowanseymour: @nicpottier this test is showing what happens when we find two different translations of a base text

i18n/testdata/translation_mismatches.noargs.es.po
```diff
@@ -0,0 +1,41 @@
...
+#, fuzzy
+msgid ""
+msgstr ""
+"POT-Creation-Date: 2020-03-25 13:57+0000\n"
+"Language: es\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+
+#: Translated/43f7e69e-727d-4cfe-81b8-564e7833052b/name:0
+#: Translated/e42deebf-90fa-4636-81cb-d247a3d3ba75/quick_replies:1
```
nicpottier: I like this for readability but you don't want the UUID in there still? Guess we can say don't rename your flows after generating PO files though.

i18n/po.go
rowanseymour: At least one of the ones I looked at didn't even support hacking together your own PO file

i18n/testdata/translation_mismatches.noargs.es.po
rowanseymour: To be clear.. we're not going to use this programmatically for anything.. it's just a hint to a translator... and I'm leaning to toward the simplicity of not having two UUIDs in there. I'm thinking we'd use `msgctxt` to match a specific node to update with a new translation..

nicpottier: Ok, ya, that makes sense and is consistent with the use of those comments in other tools. 👍

rowanseymour: We can experiment in future - maybe Transifex would work with URLs, and the editor could accept URLs with `#` links to specific parts of flows 🤔

PR #892, opened 2020-03-26
🗜️ Merge PO entries with the majority translation

rowanseymour:
When translations differ, still merge the extractions with the majority translation. For example in the covid bot flow there are two places _MENU_ is translated to _Menu_, one where it's translated to _Other_. _Menu_ wins and we create a context-less entry for that, and a context-specific one for _Other_. I.e. this..

Captura de Pantalla 2020-03-26 a la(s) 17 42 35

becomes..

Captura de Pantalla 2020-03-26 a la(s) 17 43 17

### DIFF
--- a/i18n/testdata/translation_mismatches.noargs.es.po
+++ b/i18n/testdata/translation_mismatches.noargs.es.po
@@ -17,15 +17,14 @@ msgstr "Azul"
 msgid "Other"
 msgstr "Otro"
 
-#: Translated/e42deebf-90fa-4636-81cb-d247a3d3ba75/quick_replies:0
-msgctxt "e42deebf-90fa-4636-81cb-d247a3d3ba75/quick_replies:0"
-msgid "Red"
-msgstr "Rojo"
-
 #: Translated/d1ce3c92-7025-4607-a910-444361a6b9b3/name:0
 msgctxt "d1ce3c92-7025-4607-a910-444361a6b9b3/name:0"
 msgid "Red"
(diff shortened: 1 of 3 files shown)


i18n/testdata/translation_mismatches.noargs.es.po
```diff
@@ -17,16 +17,15 @@ msgstr "Azul"
...
-msgstr "Rojo"
-
 #: Translated/d1ce3c92-7025-4607-a910-444361a6b9b3/name:0
 msgctxt "d1ce3c92-7025-4607-a910-444361a6b9b3/name:0"
 msgid "Red"
 msgstr "Roja"
 
+#: Translated/e42deebf-90fa-4636-81cb-d247a3d3ba75/quick_replies:0
+msgid "Red"
+msgstr "Rojo"
```
rowanseymour: this became the majority translation because it occurs first

nicpottier:
Do opposing translations get grouped as well? IE, two groups, one doing `yes` -> `si` and one doing `yes` -> `oui`?

rowanseymour:
@nicpottier the problem there is there isn't an obvious way to describe the context of more than one group... currently the majority group gets the context = `""`.. everyone else get's a context describing a unique location in the flow

nicpottier:
Ah right, that makes sense.

nicpottier:
PS: political junkies would call out that you are looking for a plurality not a majority. ;)

PR #896, opened 2020-03-27
🚚 Support importing translations from a PO

rowanseymour:
* Simplifies `Localization` on flows
 * Adds `Source-Flows` header to exported PO files
 * Adds `ImportIntoFlows` to update flow translations from a PO file


flows/definition/localization.go
```diff
@@ -26,8 +26,8 @@ type itemTranslation map[string][]string
 // }
 type languageTranslation map[uuids.UUID]itemTranslation
 
-// GetTextArray returns the requested item translation
-func (t languageTranslation) GetTextArray(uuid uuids.UUID, property string) []string {
+// returns the requested item translation
+func (t languageTranslation) getTextArray(uuid uuids.UUID, property string) []string {
```
rowanseymour: made some simplifications to `Flow.Localization` stuff so that the API is just `GetItemTranslation` and `SetItemTranslation `

i18n/flows.go
```diff
@@ -211,3 +217,111 @@ func poFromExtracted(initialComment string, lang envs.Language, extracted []*ext
...
+	return nil
+}
+
+// TranslationUpdate describs a change to be made to a flow translation
+type TranslationUpdate struct {
+	textLocation
+	Base string
+	Old  string
+	New  string
+}
```
rowanseymour: Figuring out the list of things that should change before making the changes made it a bit easier to test this stuff.. and also gives us an easy way forward for providing a preview view to users.

PR #897, opened 2020-04-02
🥟 Include 3-char language code as extra header in PO files

rowanseymour:


i18n/testdata/exports/two_questions.en.po
```diff
@@ -7,6 +7,7 @@ msgstr ""
 "Language: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
+"Language-3: \n"
```
nicpottier, line 10: Guessing there's no base language but then why is the file called `.en`?

i18n/testdata/exports/two_questions.po
```diff
@@ -7,6 +7,7 @@ msgstr ""
 "Language: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
+"Language-3: \n"
```
nicpottier, line 10: Should we include empty headers?  Don't feel strongly either way but feels slightly weird.

rowanseymour: I noticed that POEdit always includes all official headers with blank values.. so 🤷‍♂

i18n/testdata/exports/two_questions.en.po
rowanseymour: Oh seems this is the case where you specify the flow base language as the language to extract.. which we treat the same as not specifying a language - creating a "language independent" POT... which a person might use as starting point for creating a new translation in a PO editor.

Which would result in a PO file that has the `Language` header set but not our custom `Language-3` header

filename here is just what the testcase is setting it to, can change for clarity.

PR #898, opened 2020-04-07
🖋️ Non-UTF8 HTTP responses 

rowanseymour:


flows/events/base_test.go
```diff
@@ -487,6 +490,6 @@ func TestWebhookCalledEventBadUTF8(t *testing.T) {
 	event := events.NewWebhookCalled(call, flows.CallStatusSuccess, "")
 
 	assert.Equal(t, "http://temba.io/", event.URL)
-	assert.Equal(t, "HTTP/1.0 200 OK\r\nContent-Length: 2\r\n\r\n\xa0\xa1", event.Response)
-	assert.False(t, utf8.Valid([]byte(event.Response)))
+	assert.Equal(t, "HTTP/1.0 200 OK\r\nContent-Length: 2\r\n\r\n...", event.Response)
+	assert.True(t, utf8.ValidString(event.Response))
```
rowanseymour, line 494: anything written into an event should be serializable as JSON, and should be valid utf-8.. so if we get a big binary payload, we replace it with "..." which we use elsewhere when things like traces need to be truncated to 10K chars

services/webhooks/service.go
```diff
@@ -43,22 +43,24 @@ func (s *service) Call(session flows.Session, request *http.Request) (*flows.Web
 		}
 	}
 
+	// If user has explicitly set Accept-Encoding: gzip, remove it as Transport will add this itself,
+	// and it only does automatic decompression if its the one to set it.
+	if request.Header.Get("Accept-Encoding") == "gzip" {
```
rowanseymour, line 48: A user was running into this problem yesterday - they were setting `Accept-Encoding` explicitly - usually `Transport` does this automatically and then transparently decodes the response.. if you set it explicitly.. there's no automatic decompression.. and we get a binary response which eventually blows up when mailroom tries to save it as UTF8 in the database.

nicpottier: Oh man, subtle!

PR #902, opened 2020-04-16
Contact blocked stopped fields modifiers

norkans7:


flows/contact.go
```diff
@@ -360,6 +382,10 @@ func (c *Contact) ReevaluateDynamicGroups(env envs.Environment) ([]*Group, []*Gr
 	removed := make([]*Group, 0)
 	errs := make([]error, 0)
 
+	if c.Blocked() || c.Stopped() {
+		return added, removed, errs
```
rowanseymour, line 386: isn't this supposed to remove the contact from groups?

rowanseymour: i.e. return all groups as in removed?

norkans7: I think this is a protection to make sure that we can not recalculate the groups for blocked contacts, 
and the groups are removed on https://github.com/nyaruka/goflow/pull/902/files#diff-1f0f8102df6183de84f3f896458164e3R43 and their events are produced there

rowanseymour: This seems more complicated than it needs to be... if `Contact.ReevaluateDynamicGroups` knows that blocked/stopped means remove from any group, then all the modifiers have to do is

1. Change the state of `stopped` / `blocked`
2. Call `m.reevaluateDynamicGroups(env, assets, contact, log)`

Wouldn't that work?

norkans7: Yep, that will work

norkans7:
@rowanseymour it is now updates

flows/contact.go
```diff
@@ -360,23 +382,33 @@ func (c *Contact) ReevaluateDynamicGroups(env envs.Environment) ([]*Group, []*Gr
...
-				added = append(added, group)
+		for _, group := range c.assets.Groups().All() {
+			if !group.IsDynamic() {
+				continue
 			}
-		} else {
-			if c.groups.Remove(group) {
-				removed = append(removed, group)
+
+			qualifies, err := group.CheckDynamicMembership(env, c)
```
rowanseymour, line 399: Ok sorry one more tweak!  Just push the check for stopped/blocked into `Group.CheckDynamicMembership`.. then you don't need any extra logic here. All the groups will return false. 

Plus it's good to have that method be correct for stopped/blocked contacts.

We're essentially saying every dynamic group has an implicit `AND stopped = false AND blocked = false in its query`

norkans7: What about the static groups? I think the logic we want is to remove them as well

rowanseymour: Ah good point. I still think we want `Group.CheckDynamicMembership` and `Contact.ReevaluateDynamicGroups` to be correct for stopped/blocked.

How about rename `BaseModifier.reevaluateDynamicGroups` to `BaseModifier.reevaluateGroups` and give it a new bool arg like `removeStatic` that you set to true if `stopped||blocked`

Then it can include the static groups in the event that it generates?

norkans7: I updated that

PR #904, opened 2020-04-20
🈹 Fix anywhere we truncate strings to do it by rune

rowanseymour:
Otherwise you can end up with invalid UTF-8 byte sequences

This might fix https://sentry.io/organizations/nyaruka/issues/1621490559/activity/?project=1285856&referrer=alert_email .. not sure.. but it's definitely a bug


flows/actions/modifiers/field.go
```diff
@@ -40,13 +41,17 @@ func (m *FieldModifier) Apply(env envs.Environment, assets flows.SessionAssets,
...
 	if !m.value.Equals(oldValue) {
-		// truncate text value if necessary
-		if m.value != nil && m.value.Text.Length() > env.MaxValueLength() {
-			m.value.Text = m.value.Text.Slice(0, env.MaxValueLength())
+		var value *flows.Value
+
+		// copy and truncate text value if necessary
+		if m.value != nil {
+			v := *m.value
+			value = &v
```
rowanseymour, line 49: need to make a copy otherwise the modifier itself is modified

flows/actions/modifiers/testdata/field.json
```diff
@@ -118,5 +118,51 @@
...
+            "created_on": "2018-06-20T11:40:30.123456789Z"
+        },
+        "modifier": {
+            "type": "field",
+            "field": {
+                "key": "gender",
+                "name": "Gender"
+            },
+            "value": {
+                "text": "創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息創建流程以發送消息"
```
rowanseymour, line 141: using mandarin characters to ensure we're counting characters and not bytes

flows/actions/modifiers/field.go
nicpottier: Ooh, this seems kinda subtle. Almost makes me wonder whether `Get` should be returning a value vs a pointer then this wouldn't be a mistake that could be made.

rowanseymour: We're using `value: null` for clearing fields right now

The good news is that it's fairly easy to spot in our tests because the modifier get's re-serialized

nicpottier: Just noting it. Empty value would work just as well for the empty case. Just feels a little too subtle but we can track it.

PR #906, opened 2020-04-23
Rename blocked and stopped modifiers to contact status modifier

norkans7:


flows/actions/modifiers/status.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Apply applies this modification to the given contact
+func (m *StatusModifier) Apply(env envs.Environment, assets flows.SessionAssets, contact *flows.Contact, log flows.EventCallback) {
+
+	clearStatic := false
+	changed := false
+
+	if m.Status == flows.ContactStatusBlocked && !contact.Blocked() {
+		contact.SetBlocked(true)
```
rowanseymour, line 42: you should update contact to have `Status()` and `SetStatus()`.. that's how we want to expose it in the engine, and it'll make this method much simpler

flows/actions/modifiers/status.go
```diff
@@ -0,0 +1,55 @@
...
+	return &StatusModifier{
+		baseModifier: newBaseModifier(TypeStatus),
+		Status:       status,
+	}
+}
+
+// Apply applies this modification to the given contact
+func (m *StatusModifier) Apply(env envs.Environment, assets flows.SessionAssets, contact *flows.Contact, log flows.EventCallback) {
+
+	changed, blockedOrStopped := contact.SetStatus(m.Status)
```
rowanseymour, line 38: This can be implemented just like the other modifiers that operate on one field - just compare `contact.Status()` with `m.Status` here

You also don't need `blockedOrStopped`.. you're passing contact to `m. reevaluateGroups` so it can look at the status there

flows/contact.go
```diff
@@ -143,14 +157,34 @@ func (c *Contact) Language() envs.Language { return c.language }
...
-func (c *Contact) SetBlocked(blocked bool) { c.blocked = blocked }
-
 // Stopped returns whether the contact is stopped or not
 func (c *Contact) Stopped() bool { return c.stopped }
 
-// SetStopped sets the stopped stte of this contact
-func (c *Contact) SetStopped(stopped bool) { c.stopped = stopped }
+// Status returns the contact status
+func (c *Contact) Status() ContactStatus {
+	if c.stopped {
```
rowanseymour, line 165: replace `c.stopped` and `c.blocked` with `c.status`.. `Contact`  here doesn't need to reflect what we currently have in the RP database.. which is almost certainly going to change

flows/actions/modifiers/status.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func init() {
+	registerType(TypeStatus, readStatusModifier)
+}
+
+// TypeStatus is the type of our status modifier
+const TypeStatus string = "status"
+
+const modifierActiveContactStatus flows.ContactStatus = "active"
```
norkans7, line 20: modifier will be using `"active"` but the `contact` will be using ""

We map that on

https://github.com/nyaruka/goflow/pull/906/files?file-filters%5B%5D=.go&file-filters%5B%5D=.json&hide-deleted-files=true#diff-248bca69f6b915f47e0f442a92a70902R40-R43

rowanseymour: Not following.. we've said we want `contact` to have a `status` field the way `run` and `session` have..

```
@run.status = active|completed|...
@contact.status = active|blocked|...
```

norkans7: I get that now

flows/contact.go
```diff
@@ -526,8 +530,7 @@ func ReadContact(sa SessionAssets, data json.RawMessage, missing assets.MissingC
 		id:        envelope.ID,
 		name:      envelope.Name,
 		language:  envelope.Language,
-		blocked:   envelope.Blocked,
-		stopped:   envelope.Stopped,
+		status:    envelope.Status,
```
rowanseymour, line 533: one last tiny tweak - need to default `status` to `active` if it's not set in the JSON

PR #907, opened 2020-04-23
🎟️ Tickets

rowanseymour:


assets/interfaces.go
```diff
@@ -267,6 +268,24 @@ type TemplateTranslation interface {
...
+// Ticketer is a system which can open or close tickets
+//
+//   {
+//     "uuid": "37657cf7-5eab-4286-9cb0-bbf270587bad",
+//     "name": "Support Tickets",
+//     "type": "mailgun"
+//   }
+//
+// @asset ticketer
+type Ticketer interface {
```
rowanseymour, line 283: Captura de Pantalla 2020-04-23 a la(s) 16 23 25

nicpottier: I'm not going to fight this super hard but I think this is less clear than TicketingService, especially because "Ticketer" evokes a person, which this most definitely is not.

nicpottier:
I think tickets need a body as well as a subject.

assets/interfaces.go
rowanseymour: Service has a meaning in goflow already.. if we need a service for Tickets.. then that's the TicketService.. so it can't also be the name of the asset

rowanseymour: Open to suggestions here but we can't pass TicketService assets to the TicketService

rowanseymour:
That's part of the reason I separated the service and client for these things - so at least then the client is reusable.. guess we'll see how things look when it all comes together in mailroom.

rowanseymour:
Something that's still not quite right here is that there's an event `ticket_opened` that includes HTTP logs.. but nowhere for HTTP logs if ticket opening failed. 

Maybe it should be more like `classifier_called` which exists for both success and failure cases?

rowanseymour:
My inclination here is separating things out into a..
- generic `service_called` event that bundles the HTTP logs with a reference to the service that was called
- `ticket_opened` / `airtime_transferred` / etc events not trying to model failures as well as successes.. they should be simple descriptions of a thing happening.

@nicpottier ?

nicpottier:
I think I'm fine with that. Is there any argument that the service called stuff should just be the same as calling a webhook with some extra attributes? Guess the question there is whether your @webhook gets overwritten if you do both in a sprint. But conceptually all these things are just formalized webhook calls that also expose these events so maybe that makes sense?

rowanseymour:
I feel like those things are sufficiently different to justify separate events..
 * `@webhook` I think should remain only for user initiated HTTP calls from `call_webhook` actions. I don't think users need/want access to the raw HTTP logs for other service calls at a flow level, and we'd have to be careful about redacting things if we went that route.
 * `webhook_called` creates `WebHookResults` and `WebHookEvents` in the case of resthook calls.
* `service_called` would group a bunch of HTTP logs together - i.e. it takes 3 HTTP logs to make a DTOne airtime transfer.

nicpottier:
K, I'm fine with all that. Also down for having a `ticket_opened` event for symmetry.

rowanseymour:
@nicpottier going to merge this - all the service/client stuff has moved to mailroom

PR #914, opened 2020-05-14
update our format_datetime docs to properly show range of 01-24

nicpottier:
Good old off by 1 error.


docs/expressions.html
```diff
@@ -408,7 +408,7 @@ <h2 class="item_title">
 <li><code>DD</code> - day of month, zero padded 0-31</li>
 <li><code>h</code> - hour of the day 1-12</li>
 <li><code>hh</code> - hour of the day 01-12</li>
-<li><code>tt</code> - twenty four hour of the day 01-23</li>
+<li><code>tt</code> - twenty four hour of the day 01-24</li>
```
rowanseymour, line 411: shouldn't this be 00-23 ?

nicpottier: Ack, I guess so, though both are accepted, so different off by 1. Will update to 0-23:
  https://en.wikipedia.org/wiki/24-hour_clock#Midnight_00:00_and_24:00

docs/md/expressions.md
```diff
@@ -678,7 +678,7 @@ and '_' are ignored. Any other character is an error.
 
 * `h`         - hour of the day 1-12
 * `hh`        - hour of the day 01-12
-* `tt`        - twenty four hour of the day 01-23
+* `tt`        - twenty four hour of the day 01-24
```
rowanseymour, line 681: this one needs fixed to `00-23`

PR #916, opened 2020-05-25
🌐 Pass language to bothub API calls

rowanseymour:


rowanseymour:
@johncordeiro how does this look?

services/classification/bothub/service.go
```diff
@@ -25,7 +26,10 @@ func NewService(httpClient *http.Client, httpRetries *httpx.RetryConfig, classif
 }
 
 func (s *service) Classify(session flows.Session, input string, logHTTP flows.HTTPLogCallback) (*flows.Classification, error) {
-	response, trace, err := s.client.Parse(input)
+	locale := session.Environment().DefaultLocale()
+	localeStr := strings.ReplaceAll(strings.ToLower(locale.ToISO639_2()), "-", "_") // en-US -> en_us
```
johncordeiro, line 30: @rowanseymour is there the possibility of returning non-existing languages such as `pt_us` in this line?

rowanseymour: As far as I understand there is no exhaustive list of what is a valid a language/country pairing.. or at least that's what I got from scanning https://tools.ietf.org/html/bcp47

So in this case the onus is on you to handle `pt_us` the same way you might handle `pt_mz` if you haven't configured Mozambique Portuguese - i.e. back down to version of `pt` that you have configured.

koallann: Hey guys, this implementation worked well, but we need a country defined in the channel to pass more specific languages to Bothub... Should we get the country from org if the channel doesn't have one? Once we should set a country to all channels from an org to have more specific languages in Bothub calls.

rowanseymour: Mmm I should be able to fix that fairly easily... the issue is that the environment on the session is the "default" environment that you trigger the session with.. but the environment on the runs has values like country overridden by the contact.

rowanseymour: Updated.. are you testing with this branch?

koallann: Yes I'm. Just to be clear, is the following order correct to define the country?

Contact's preferred channel country -> Country of the channel that flow is running -> Country of the first channel with a country set (alphabetical order for more than one) -> None

rowanseymour: The logic is...

1. Country of contact preferred channel (this just means the channel that is being used or will be used to send messages)
2. Country of org calculated from looking at the org channels for first channel with a country

koallann: Okay, thanks a lot!

Issue #915, opened 2020-05-15
Sessions should know if they're being started in a batch


rowanseymour:
Then we can have the engine do footgun protection for start_session, open_ticket etc

PR #919, opened 2020-06-02
🦶 Builtin footgun prevention

rowanseymour:
Adds #915

### DIFF
--- a/flows/actions/testdata/send_broadcast.json
+++ b/flows/actions/testdata/send_broadcast.json
@@ -73,7 +73,30 @@
         }
     },
     {
-        "description": "Error event if executed in batch start",
+        "description": "Error event if executed in batch start and uses a group",
+        "as_batch": true,
+        "action": {
+            "type": "send_broadcast",
+            "uuid": "ad154980-7bf7-4ab8-8728-545fd6378912",
+            "groups": [
+                {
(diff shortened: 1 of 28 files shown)


flows/actions/testdata/send_broadcast.json
```diff
@@ -72,6 +72,26 @@
...
+                "tel:+1234567890"
+            ],
+            "text": "Hi there!"
+        },
+        "events": [
+            {
+                "type": "error",
+                "created_on": "2018-10-18T14:20:30.000123456Z",
+                "step_uuid": "59d74b86-3e2f-4a93-aece-b05d2fdcde0c",
+                "text": "can't send broadcasts during batch starts"
```
rowanseymour, line 91: is this maybe too strict?

nicpottier: sending broadcasts? It is the original footgun so no, I don't think so.

rowanseymour: I thought maybe there's a valid case where your sending to one person.. like you start a bunch of people in a flow that messages their supervisor... so I was wondering if this should allow a broadcast to 1 URN or contact

nicpottier: Hah, was literally just looking at that myself. But yes, should be consistent.

PR #921, opened 2020-06-05
🌐 Update change-language functionality to allow missing translations

rowanseymour:


flows/definition/testdata/TestChangeLanguage_change_language_to_spa.snap
```diff
@@ -63,43 +63,6 @@
                     "Which pill?"
                 ]
             }
-        },
```
rowanseymour, line 66: this was a mistake - it was leaving behind the translation that becomes the flow text

flows/definition/flow.go
```diff
@@ -189,42 +187,34 @@ func (f *flow) ChangeLanguage(lang envs.Language) (flows.Flow, error) {
...
 
-		newValues := newTranslation.getTextArray(uuid, property)
+		newValues := inTranslation.getTextArray(uuid, property)
 
-		if len(oldValues) > 0 && len(newValues) == 0 {
-			missing = append(missing, fmt.Sprintf("%s/%s", uuid, property))
-		} else {
-			w(newValues) // update flow text in the definition
+		// if we have a translation, update flow text in the definition, if not then leave it as is
+		if len(newValues) > 0 {
```
rowanseymour, line 205: seems fine to let engine be dumb here and just default any missing translations to base text.. editor is the one tracking rules/categories translated and can make the decision about whether a flow is allowed to be translated

PR #922, opened 2020-06-08
🚏 Localize completion help strings

rowanseymour:
Adds the help strings from completion to the locale files


cmd/docgen/docs/base.go
```diff
@@ -21,11 +21,8 @@ const (
 type Generator interface {
 	Name() string
 
-	// ExtractText returns strings that can be translated
-	ExtractText(items map[string][]*TaggedItem) []string
-
 	// Generate does the actual generation in the given language
-	Generate(baseDir, outputDir string, items map[string][]*TaggedItem, po *i18n.PO) error
+	Generate(baseDir, outputDir string, items map[string][]*TaggedItem, gettext func(string) string) error
```
rowanseymour, line 25: does gathering of unique msg ids now at the same time as doc generation instead of an explicit step.. any string we request from gettext will end up in the POT file at the end of doc generation

PR #924, opened 2020-06-11
🐝 Fix clearing of fields

rowanseymour:
Fixes https://github.com/rapidpro/rapidpro/issues/1239. Problem here is clearing a field removes it from the field->value map so if you try to reference it again in the same sprint.. boom.

The need for a set action prior is just to make sure there's a value to be cleared otherwise this clear-field action is a noop.

Captura de Pantalla 2020-06-11 a la(s) 10 22 02

### DIFF
--- a/flows/field.go
+++ b/flows/field.go
@@ -202,17 +202,11 @@ func (f FieldValues) Get(field *Field) *Value {
 
 // Set sets the value for the given field (can be null to clear it)
 func (f FieldValues) Set(field *Field, value *Value) {
-	if value == nil {
-		f.Clear(field)
-	} else {
-		fieldValue := NewFieldValue(field, value)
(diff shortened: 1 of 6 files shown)


flows/field.go
```diff
@@ -202,17 +193,11 @@ func (f FieldValues) Get(field *Field) *Value {
...
-		f[field.Key()] = fieldValue
+	var fv *FieldValue
+	if value != nil && !value.Text.Empty() {
+		fv = NewFieldValue(field, value)
 	}
-}
-
-// Clear clears the value set for the given field
-func (f FieldValues) Clear(field *Field) {
-	delete(f, field.Key())
```
rowanseymour, line 215: this is the offender!

PR #926, opened 2020-06-24
📇 URNs modifier to replace/set all URNs

rowanseymour:
From the RP UI you can completely change/reorder the contact URNs, which currently we'd have to describe using a sequence of single URN add/remove modifiers. This adds a new "URNs" modifier which can eventually replace the "URN" modifier.

### DIFF
--- a/flows/actions/modifiers/urn.go
+++ b/flows/actions/modifiers/urn.go
@@ -27,7 +27,8 @@ const (
 	URNRemove URNModification = "remove"
 )
 
-// URNModifier modifies a URN on a contact
+// URNModifier modifies a URN on a contact. This has been replaced by URNsModifier but is kept here for now
+// to support processing of old Surveyor submissions.
 type URNModifier struct {
 	baseModifier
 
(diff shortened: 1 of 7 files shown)


flows/actions/modifiers/urn.go
```diff
@@ -27,7 +27,8 @@ const (
 	URNRemove URNModification = "remove"
 )
 
-// URNModifier modifies a URN on a contact
+// URNModifier modifies a URN on a contact. This has been replaced by URNsModifier but is kept here for now
```
rowanseymour, line 30: we could eventually drop this once Surveyor has been updated enough time has passed that we're not worried about old submissions out there using this modifier type

nicpottier: But wouldn't that get us in trouble? I thought the idea of modifiers is just that, to modify, not set, otherwise a surveyor submission might override URNs that have been set from the website instead of only adding a new URN.

rowanseymour: So this new modifier can `append` / `remove` / `set` 
- flow actions are still going to use `append`
- requests from the RP UI are going to use `set`

nicpottier: Oh I see, ok great.

PR #928, opened 2020-06-25
🌾 Make field modifiers take unparsed values

rowanseymour:
This so RP doesn't have to replicate all the field parsing logic

Which sounds simple but required refactoring how we create environment instances which can location parsing so that they don't require sessions/runs. So now mailroom should be able to create an environment suitable for modifier applying like:

```go
env := flows.NewEnvironment(org.Env(), oa.Locations())
```


flows/routers/cases/tests.go
```diff
@@ -599,12 +599,12 @@ func HasTopIntent(env envs.Environment, result *types.XObject, name types.XText,
 //
 // @test has_state(text)
 func HasState(env envs.Environment, text types.XText) types.XValue {
-	runEnv, _ := env.(flows.RunEnvironment)
-
-	states, err := runEnv.FindLocationsFuzzy(text.Native(), flows.LocationLevelState, nil)
-	if err != nil {
-		return types.NewXError(err)
+	locations := env.LocationResolver()
```
rowanseymour, line 602: this feels a lot cleaner than type cracking

flows/routers/cases/tests.go
```diff
@@ -636,26 +639,17 @@ func HasDistrict(env envs.Environment, args ...types.XValue) types.XValue {
 		}
 	}
 
-	states, err := runEnv.FindLocationsFuzzy(stateText.Native(), flows.LocationLevelState, nil)
-	if err != nil {
```
rowanseymour, line 640: all the error cases here were just the case that the org doesn't have locations.. which is now handled above

assets/static/source.go
```diff
@@ -15,17 +16,17 @@ import (
...
-		Templates   []*types.Template          `json:"templates" validate:"omitempty,dive"`
-		Ticketers   []*types.Ticketer          `json:"ticketers" validate:"omitempty,dive"`
+		Channels    []*types.Channel          `json:"channels" validate:"omitempty,dive"`
+		Classifiers []*types.Classifier       `json:"classifiers" validate:"omitempty,dive"`
+		Fields      []*types.Field            `json:"fields" validate:"omitempty,dive"`
+		Flows       []*types.Flow             `json:"flows" validate:"omitempty,dive"`
+		Globals     []*types.Global           `json:"globals" validate:"omitempty,dive"`
+		Groups      []*types.Group            `json:"groups" validate:"omitempty,dive"`
+		Labels      []*types.Label            `json:"labels" validate:"omitempty,dive"`
+		Locations   []*envs.LocationHierarchy `json:"locations"`
```
rowanseymour, line 26: locations have never really been assets.. and this is making more want to ditch that idea.. but for now it still works

Issue #929, opened 2020-07-02
Manual triggers should contain user information


rowanseymour:
Which can be exposed in the context like..

```
@trigger.origin = ui | api | zapier 
@trigger.user = bob@nyaruka.com | ..
```

rowanseymour:
Added in https://github.com/nyaruka/goflow/pull/932

PR #930, opened 2020-07-03
🏗️ Add builder for triggers

rowanseymour:
Trying to think of a way to not make our convoluted trigger constructors more convoluted by adding even more arguments, e.g. https://github.com/nyaruka/goflow/issues/929 and whilst still keeping triggers immutable.

The complication here is that there are different types of triggers with different params..

```go
t := triggers.NewBuilder(env, flow.Reference(), contact).  <-- required opts for all triggers here
    Msg(msg).                                              <-- type and required opts for that type
    WithMatch(keyword).                                    <-- optional opts
    Build()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ env := envs.NewBuilder().Build()
 source, _ := static.LoadSource("myassets.json")
 assets, _ := engine.NewSessionAssets(env, source, nil)
 contact := flows.NewContact(assets, ...)
-trigger := triggers.NewManual(env, contact, flow.Reference(), nil, false, nil, time.Now())
+trigger := triggers.NewBuilder(env, contact, flow.Reference()).Manual().Build()
 eng := engine.NewBuilder().Build()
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ env := envs.NewBuilder().Build()
 source, _ := static.LoadSource("myassets.json")
 assets, _ := engine.NewSessionAssets(env, source, nil)
 contact := flows.NewContact(assets, ...)
-trigger := triggers.NewManual(env, contact, flow.Reference(), nil, false, nil, time.Now())
```
rowanseymour, line 21: `nil, false, nil` is nice example of why we wouldn't want to add more args here

rowanseymour:
I read the same article after going looking for what might be a more go way of doing this. I don't hate that functional approach but I wasn't convinced it's that much better than builders and this at least follows the pattern we've already used for environments and engines. I like that with the builder approach it's more obvious what the possible options are tho that might be also a function of how good your auto-complete is.

PR #932, opened 2020-07-09
👤 Add user and origin to manual triggers

rowanseymour:


flows/triggers/base.go
```diff
@@ -133,6 +118,49 @@ func EnsureDynamicGroups(session flows.Session, logEvent flows.EventCallback) {
 	}
 }
 
+//------------------------------------------------------------------------------------------
+// Expressions context
+//------------------------------------------------------------------------------------------
+
+// Context is the schema of trigger objects in the context, across all types
```
rowanseymour, line 125: needed a better way to keep the schema of all triggers consistent in expressions

flows/triggers/testdata/campaign.json
```diff
@@ -57,9 +57,11 @@
         },
         "events": [],
         "context": {
-            "keyword": null,
-            "params": null,
-            "type": "campaign"
+            "keyword": "",
```
rowanseymour, line 60: I've convinced myself that distinguishing between `null` and `""` isn't helpful.. and is a safe change to make since they both render to empty strings. The type of the trigger is in the context so if you need to split on that you can.. you don't need to compare `null` vs `""`

PR #938, opened 2020-07-21
🚱 Prevent looping across sessions

rowanseymour:


flows/interfaces.go
```diff
@@ -411,6 +413,7 @@ type FlowRun interface {
 	LogEvent(Step, Event)
 	LogError(Step, error)
 	Events() []Event
+	ReceivedInput() bool
```
rowanseymour, line 416: can use this in mailroom where we calculate this to set `responded` in the database

docs/en_US/md/sessions.md
```diff
@@ -74,6 +74,11 @@ Is used when another session triggered this run using a trigger_flow action.
         "uuid": "b7cf0d83-f1c9-411c-96fd-c511a4cfa86d",
         "name": "Collect Age"
     },
+    "history": {
+        "parent_uuid": "a5b25fb0-75fd-4898-a34f-5ff14fc19078",
+        "ancestors": 3,
+        "ancestors_since_input": 1
```
rowanseymour, line 80: `ancestors_since_input` is the value we actually use but the other two could be handy when debugging

flows/actions/base_test.go
```diff
@@ -852,3 +855,217 @@ func TestResthookPayload(t *testing.T) {
...
+
+	assert.Equal(t, 5, len(sessions))
+
+	// final session should have an error event
+	finalEvent := sprint.Events()[len(sprint.Events())-1]
+	assert.Equal(t, events.TypeError, finalEvent.Type())
+	assert.Equal(t, "too many sessions have been spawned since the last time input was received", finalEvent.(*events.ErrorEvent).Text)
+}
+
+func TestStartSessionLoopProtectionWithInput(t *testing.T) {
```
rowanseymour, line 943: tests the case where sessions keep spawning sessions.. but they wait and receive input so it's ok

flows/events/session_triggered.go
```diff
@@ -40,6 +40,11 @@ const TypeSessionTriggered string = "session_triggered"
 //           "created_on": "2000-01-01T00:00:00.000000000-00:00"
 //         }
 //       }
+//     },
+//     "history": {
+//       "parent_uuid": "55105da5-abb5-4690-b1f6-ec2e5762a561",
+//       "ancestors": 3,
+//       "ancestors_since_input": 1
```
rowanseymour, line 47: @nicpottier so I don't think just having this on the event is going to be enough - for the IVR case this would need to be saved to `flows_flowstart`.. which could be achieved a few different ways..

 1. Each of these things becomes a field on `flows_flowstart` (great for querying, maybe useful..)
 2. Add new JSON column to hold the history blob
 3. Save the entire `session_triggered` event to a new blob field - eventually it would replace use of `parent_summary`

docs/en_US/md/sessions.md
nicpottier: Since this is just our session blob and we are going through this, why not just have it be complete instead of this summary? IE something more like:

```
"history": [{
  "session_uuid": "arst",
  "created_on": "2020---"
},{
  "session_uuid": "arst",
  "had_input": true,
  "created_on": "2020---"
}
]
```

Just brainstorming. I guess I feel like maintaining size is not a super important consideration here because these are sessions and we may as well annotate with all the information that is useful right in the session. Other thought is that maybe having a bit more information would help us more easily adapt our rules as to when we shortcircuit but perhaps not an important consideration.

flows/events/session_triggered.go
nicpottier: I *think* I prefer 2, but I don't feel super strongly on these.

docs/en_US/md/sessions.md
rowanseymour: I talked myself out of this approach because 1) it would have to be trimmed because it's valid usecase to eternally spawn new sessions that wait and receive input and 2) from a debugging POV (where performance is not an issue) that data would be available by just giving each session a parent UUID and walking those.

But I'm kinda on the fence.. it works as long as the trimming length >= the limit.. hmm..

nicpottier: I guess the other approach would be do we need to "keep" history that involves input? Then we can remove that concept entirely then no? Basically if there has been input in the current sprint then history is not passed through to any new sessions created. But then maybe the concept of "history" is too vague.

rowanseymour: But then you lose potentially interesting information if there is input.. or ability to re-use this information with slightly different rules for loop detection.. we'd probably call this something like `loop_detection_history`

I think I've talked myself into keeping this as is and if in the future we want more debugging info.. it's a dict so could add an array field like `last_sessions` or whatever.

Also if we store this as a blob called `history` on the flow start then we can add whatever there without DB changes

PR #946, opened 2020-07-27
USSD Channel - initial changes to flows

dsmagicug:
These changes are requested to allow the flows to indicate to the USSD channel when a conversation should end.


rowanseymour:
As we said in the email thread, please open an issue on https://github.com/rapidpro/rapidpro to discuss USSD changes. I'm pretty sure we don't need any changes to goflow.

PR #953, opened 2020-08-06
👁️ Add last_seen_on to contacts and expose in expressions and queries

rowanseymour:


flows/engine/session.go
```diff
@@ -159,6 +166,9 @@ func (s *session) start(trigger flows.Trigger) (flows.Sprint, error) {
 		return sprint, err
 	}
 
+	// ensure dynamic groups are correct
+	s.ensureDynamicGroups(sprint.LogEvent)
```
rowanseymour, line 170: this was previously happening in resumes and triggers but easier to do it in the session, to ensure it happens after things that can affect the groups.. like setting input which changes last_seen_on

test/testdata/runner/two_questions.test.json
```diff
@@ -231,10 +217,24 @@
...
+                            "name": "Nexmo",
+                            "uuid": "57f1078f-88aa-46f4-a59a-948a5739c03d"
+                        },
+                        "text": "I like blue!",
+                        "urn": "tel:+12065551212",
+                        "uuid": "9bf91c2b-ce58-4cef-aacc-281e03f69ab5"
+                    },
+                    "step_uuid": "8720f157-ca1c-432f-9c0b-2014ddc77094",
+                    "type": "msg_received"
+                },
```
rowanseymour, line 237: this is just an event ordering change - the `contact_refreshed` coming before the `msg_received`  so changes to last_seen_on aren't immediately lost by replacing the contact

flows/resumes/base.go
```diff
@@ -52,7 +51,7 @@ func (r *baseResume) Contact() *flows.Contact       { return r.contact }
 func (r *baseResume) ResumedOn() time.Time          { return r.resumedOn }
 
 // Apply applies our state changes and saves any events to the run
-func (r *baseResume) Apply(run flows.FlowRun, logEvent flows.EventCallback) error {
+func (r *baseResume) Apply(run flows.FlowRun, logEvent flows.EventCallback) {
```
rowanseymour, line 54: kinda unrelated but I noticed that there's no code paths now where Resume apply methods can error

contactql/es/query.go
```diff
@@ -256,6 +256,35 @@ func attributeConditionToElastic(env envs.Environment, c *contactql.Condition) e
...
+		}
+
+		value := c.ValueAsDate()
+		start, end := dates.DayToUTCRange(value, value.Location())
+
+		switch c.Operator() {
+		case contactql.OpEqual:
+			return elastic.NewRangeQuery("last_seen_on").Gte(start).Lt(end)
+		case contactql.OpNotEqual:
+			return not(elastic.NewRangeQuery("last_seen_on").Gte(start).Lt(end))
```
nicpottier, line 276: So `last_seen` can be null, is this going to behave the way we expect (or the user's expect) for these range queries? IE, does Elastic interpret `null` as a value? Notice that for the field case we check that that field is set as part of the query but that may be due to their structure. Could be worth trying out manually if you haven't already. (to be clear I *think* user expectation would be that last_seen comparisons always excluded null values)

rowanseymour: I checked that when I was updating rp-indexer - NULLs just aren't returned by range queries https://github.com/nyaruka/rp-indexer/blob/d85bf42559aa2afe71f91d39036b2e694fe2cd16/indexer_test.go#L102-L107

So this should behave the same as custom date field with no value

PR #962, opened 2020-09-03
🧏 Fix contact language not overriding language on environment

### LINKED ISSUE #960: Change request: language param in Bothub classifier requests
Hey @rowanseymour, some time ago you [implemented](https://github.com/nyaruka/goflow/pull/916/files) the passing of the language param in Bothub classifier requests, where this param is defined from the org's language. Currently, we are noticing that the ideal is for this parameter to be defined from the contact's language if it's set, otherwise from the org's language.

We would be very grateful with this adjustment.

rowanseymour:
Fixes #960


flows/runs/environment.go
```diff
@@ -25,22 +25,45 @@ func newRunEnvironment(base envs.Environment, run *flowRun) *runEnvironment {
...
 		cc := contact.Country()
 		if cc != envs.NilCountry {
 			return cc
 		}
 	}
-	return e.run.Session().Environment().DefaultCountry()
+	return e.Environment.DefaultCountry()
+}
+
+func (e *runEnvironment) DefaultLocale() envs.Locale {
```
rowanseymour, line 58: the other fix needed was redeclaring `DefaultLocale` otherwise with the lack of polymorphism it's the method on the base env using the base env's values

PR #967, opened 2020-10-12
Include country in msg templating on msg_created events

rowanseymour:


flows/actions/testdata/send_msg.json
```diff
@@ -485,6 +486,7 @@
                             "name": "affirmation"
                         },
                         "language": "spa",
+                        "country": "",
```
rowanseymour, line 489: I assume not all templates have country set

nicpottier: In the whatsapp world they always will but ya, generally no.

nicpottier (approved):

PR #968, opened 2020-10-12
⏳ Unwind accumulated run expirations as child runs complete

rowanseymour:
Fixes https://github.com/nyaruka/mailroom/issues/369

In b1f2f151805b91357c144631f68b316ff213a522 and e9696d9dcdec6b499b25655c3610aa866a2401ec I added new runner tests where a flow A calls flow B calls flow C - flow C waits for a question, resumes B which waits for another question, resumes A which waits for another question. The expiration time for A is 60 minutes, for B is 120 minutes and for C is 180 minutes.

So when you're waiting in C for the first question, the expiration time for..

- C is NOW+180 minutes
- B is NOW+300 (120 + 180) minutes
- A is NOW+360 (60 + 120 + 180) minutes

Previously we'd only recalculate expirations in a run when that run got input so for example if C and B expired, A's expiration time would remain at NOW+360. This PR changes it so that we now clear and recalculate expiration times on parents as runs exit. So now if C expires..

- C is null
- B is NOW+120 minutes
- A is NOW+180 (60 + 120) minutes

and then B expires...

- C is null
- B is null
- A is NOW+60 minutes

and then A expires...

- C is null
- B is null
- A is null

Clearing the expiration when a run exits seems correct.. but it's a change that affects a lot of test files so apologies for the diff below but I can promise I've checked it's behaving as described above.


rowanseymour:
cc @tybritten

nicpottier (approved):
Yep.

I do wonder if our approach of current flow always determines expiration is quite right but I guess we need some more concrete thoughts on why that is wrong before changing it.

But it does strike me that our behavior is different than say Golang context timeouts, where a child context / method can shorten the context, but never lengthen it.

IE, it is kinda weird that a parent flow can have say a five minute expiration but call a child flow with a two week expiration and get "caught" in that child flow if the contact never responds. Almost feels like the expiration should never be later than the parent's. But again, can't make a stronger case for that apart from "feels weird"

rowanseymour:
Hmm yeah... arguably setting a 5 minute expiration on a parent flow should guarantee we complete that parent flow in 5 minutes, regardless of what child runs we trigger. When the parent expires, all the children expire even if they have a 2 week expiration. 

But then we've got the complication of sometimes being able to resume the parent, sometimes not because it has already expired. And I'm not sure users would expect a flow with a two week expiration to expire in 5 minutes if it happens to be started from a certain parent... 

I dunno.. worth thinking about but we might be committed to this understanding of expirations for now.

rowanseymour:
Looks like mailroom always assumes `expires_on` has a value so that would need tweaking before this goes live

rowanseymour:
@nicpottier thinking about this a bit more... with this bug fixed.. we don't actually add expirations in any meaningful sense. When flow A enters B and its expiration time becomes 60 + 120... we never actually expire A like that. We basically add to its expiration to ensure it's not expired before B, but we could punt its expiration to any arbitrary value in the future.. or just clear it.

Effectively when you enter a subflow, expiration is disabled on the parent.. so maybe we should model that explicitly and say its `expires_on` becomes `null`, and only when you re-enter that parent does it get reset.

nicpottier:
Right right.. I maybe have some concerns that expirations should always be set for active flows in case something goes sideways, so perhaps the current method isn't so bad.

tybritten:
Also, can we configure the default expiration to something other than `10080` minutes?

rowanseymour:
@nicpottier ok let's try this for now

@tybritten I guess that could be made configurable - it's just a constant in RapidPro (30 days)

PR #975, opened 2020-10-28
SetPreferredChannel only when the channel has the send role

norkans7:


flows/contact.go
```diff
@@ -400,13 +400,13 @@ func (c *Contact) UpdatePreferredChannel(channel *Channel) bool {
 
 		for _, urn := range c.urns {
 			// tel URNs can be re-assigned, other URN schemes are considered channel specific
-			if urn.URN().Scheme() == urns.TelScheme && channel.SupportsScheme(urns.TelScheme) {
+			if urn.URN().Scheme() == urns.TelScheme && channel.SupportsScheme(urns.TelScheme) && channel.HasRole(assets.ChannelRoleSend) {
```
rowanseymour, line 403: I think this fix has it backward.. if we allow channel roles to be changed then we're accepting the URNs may be associated with a channel that can't send. 

I think you need to tweak `ChannelAssets.GetForURN` so it only uses the URN channel if it has the correct role.

rowanseymour (changes requested):

flows/contact.go
nicpottier: While channels being mutable does bring up the possibility that a previously preferred channel is no more, it does seem like we would want to avoid overwriting previous affinity when we know a channel isn't eligible to send no?

IE, I think we need both no?

norkans7: Right, I think we still want to prevent the channel to be set on the URL if we know we cannot send with that.
And the flow editor will also allow to use channel that can sent to be used to update a contact

I am adding the changes for `GetForURN`

rowanseymour: If that's the case (which I think does make sense) then the action should create an error event rather than just having this method silently do nothing. 

And possibly the editor shouldn't allow it in the first place but we can worry about that another time.

rowanseymour:
@norkans7 I think you're still missing having the action error if the channel doesn't have the send role. Otherwise the action just silently does nothing.

flows/contact.go
```diff
@@ -401,13 +402,21 @@ func (c *Contact) UpdatePreferredChannel(channel *Channel) bool {
 		for _, urn := range c.urns {
 			// tel URNs can be re-assigned, other URN schemes are considered channel specific
 			if urn.URN().Scheme() == urns.TelScheme && channel.SupportsScheme(urns.TelScheme) {
-				urn.SetChannel(channel)
+				if channel.HasRole(assets.ChannelRoleSend) {
+					urn.SetChannel(channel)
+				} else {
+					errs = append(errs, errors.Errorf("unable to set channel on urn '%s', missing the send role on channel '%s'", urn.URN().String(), channel.UUID()))
```
rowanseymour, line 408: This seems more complicated than it needs to be. returning a separate error for every URN

Just have the action generate an error event instead of a modifier if the channel doesn't have a send role.

This method could start by checking if the channel has the send role once.. but that's not so important since it'll never be called with such a channel.

norkans7: so I can make that one error instead of one error per URN,
but I don't understand how to generate the error event instead of a modifier?

rowanseymour: You just need something like this in the action:

```
	if a.Channel != nil {
		channel = run.Session().Assets().Channels().Get(a.Channel.UUID)
		if channel == nil {
			logEvent(events.NewDependencyError(a.Channel))
			return nil
		}

		if !channel.HasRole(assets.ChannelRoleSend) {
			logEvent(events.NewErrorf("can't set channel that can't send as the preferred channel"))
			return nil
		}
	}
```

norkans7: never mind I get that now

rowanseymour (approved):

rowanseymour:
Added a test for good measure https://github.com/nyaruka/goflow/commit/20a4162292c5039c4f852c23df38ffa07c2bd349

Issue #980, opened 2020-12-08
Flow inspection should include information about waits and sends


rowanseymour:
In RapidPro we want to give users options to start flows that don't interrupt other flows. This definitely means that the flow can't lead to a wait, tho could possibly allow sending of messages. Probably best not to allow subflows so we don't have to recursively check each subflow destination.

Maybe the most flexible approach at the engine level would be just list out features that the flow uses. That way the caller (in this case RapidPro) can determine for itself what it wants to allow with a given flow without having understand the flow spec, e.g. include this in flow inspection:

```
"features": ["waits", "sends", "webhooks", "subflows", "email"],
"issues": [...],
"dependencies": [...]
```

And that information may be useful elsewhere, e.g. determining whether a flow can be started without an org having a channel.

nicpottier:
I like this. Any reason for this not to be exactly the union if actions + waits?

rowanseymour:
I went back and forth on that.. and figured maybe that was asking the caller to know too much about the flow spec - i.e. RapidPro has to know that sending is both `send_broadcast` and `send_msg`, subflows are both `enter_flow` and `start_session`.. but I could be persuaded that is fine.

rowanseymour:
Have persuaded myself it's most flexible to keep this as high-level flags so it can include anything.. maybe `looping` is a feature

rowanseymour:
See PR for why we created a new flow type instead..

PR #981, opened 2020-12-11
🔬 Include high-level features in flow inspection

rowanseymour:
Does #980

### DIFF
--- a/flows/actions/testdata/add_contact_groups.json
+++ b/flows/actions/testdata/add_contact_groups.json
@@ -42,6 +42,7 @@
             }
         ],
         "inspection": {
+            "features": [],
             "dependencies": [
                 {
                     "uuid": "33382939-babf-4982-9395-8793feb4e7c6",
@@ -112,6 +113,7 @@
             "Customers@(1 / 0)"
         ],
         "inspection": {
+            "features": [],
             "dependencies": [],
             "issues": [],
             "results": [],
@@ -142,6 +144,7 @@
             "Climbers"
         ],
         "inspection": {
+            "features": [],
             "dependencies": [],
             "issues": [],
             "results": [],
--- a/flows/actions/testdata/add_contact_urn.json
+++ b/flows/actions/testdata/add_contact_urn.json
@@ -20,6 +20,7 @@
             "+1234567890"
         ],
         "inspection": {
+            "features": [],
             "dependencies": [],
             "issues": [],
             "results": [],
--- a/flows/actions/testdata/add_input_labels.json
+++ b/flows/actions/testdata/add_input_labels.json
@@ -44,6 +44,7 @@
             "Spam@(1 / 0)"
         ],
         "inspection": {
+            "features": [],
             "dependencies": [],
             "issues": [],
             "results": [],
@@ -119,6 +120,7 @@
             }
         ],
         "inspection": {
+            "features": [],
             "dependencies": [
                 {
                     "uuid": "33382939-babf-4982-9395-8793feb4e7c6",
--- a/flows/actions/testdata/call_classifier.json
+++ b/flows/actions/testdata/call_classifier.json
@@ -32,6 +32,7 @@
             "@input.text"
         ],
         "inspection": {
+            "features": [],
             "dependencies": [
                 {
                     "uuid": "63998ee7-a7a5-4cc5-be67-c773e1b6b9b1",
@@ -103,6 +104,7 @@
             "@(\"\")"
         ],
         "inspection": {
+            "features": [],
             "dependencies": [
                 {
                     "uuid": "1c06c884-39dd-4ce4-ad9f-9a01cbe6c000",
@@ -200,6 +202,7 @@
             "@input.text"
         ],
         "inspection": {
+            "features": [],
             "dependencies": [
                 {
                     "uuid": "1c06c884-39dd-4ce4-ad9f-9a01cbe6c000",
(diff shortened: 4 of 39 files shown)


rowanseymour:
Been going back and forth on this wondering if it shouldn't be _even_ higher-level - maybe call it `flags` and have things like `passive`, `channel_io` be flags. The danger of having RP look for the absence of features is that after you've added new feature checkers, you don't know if an existing inspection blob represents the actual absence of that feature, or the fact you haven't run inspection since adding a few feature checker...

nicpottier:
Ya, as this stands this smells a bit of over/pre engineering to me, in that we are trying to create some classification/categorization of flows without knowing what will be useful in the future, so it feels like it is bound to be wrong.

If all we care about is passive/active maybe we should just ONLY do that?

rowanseymour:
Tho what's the actual definition of passive.. do we want to support non-interrupting flows that send messages? 

You could imagine surfacing this to users in RP where non-messaging passive flow is given the option to be run without interrupting. A passive flow that sends messages is given that option and maybe a warning?

nicpottier:
Yes, I do think we want to allow sending messages while other flows are active. (one concrete use case there might be sending a last "make sure to reply!" message 23 hours after a last response to a contact while keeping them in the flow so that when they do reply their response is processed)

But you raise a fair point of do we need to differentiate a flow that is passive and sends no messages from one that is passive and does send messages. 🤷 `flags` could maybe capture that.. (would be absence of `wait` then whether `messaging` exists or something?)

rowanseymour:
OK then seems there are two things goflow needs to figure out on behalf of RP:

1. whether or not a flow always finishes in a single sprint (no waits, no subflows)
2. whether or not a flow sends messages

Tho 2. on its own isn't particularly useful. If we want to let users start flows without having a channel, the only way to be sure that it's not going to try sending messages is if it doesn't have subflows.

So maybe 3 flags like `waits`, `subflows`, `messaging` is the way forward

nicpottier:
We would still consider a passive flow that calls a subflow that is also passive, passive right?

Are we thinking about this backwards? Should this instead be a restriction based on the type of flow instead of figuring out if a flow is passive after the fact. I can imagine for example you having some kind of trigger that requires a passive flow and we'd want to make sure that the flow the trigger points to REMAINS passive.

So almost wonder if this is more a type of flow (passive) that can call other passive flows only and it is an attribute of the editor that it disallows waits.

rowanseymour:
I think if we weren't talking about including non-interrupting messaging flows then that would be the obvious solution. But thus far flow type has kinda meant the medium it's used over, so I'm kinda struggling with the idea of non-waiting as a type.

You could imagine having a non-messaging type called `utility` (it doesn't run over messaging/voice/surveyor thus it's not one of those types). If we wanted to get fancy then `utility` flows can only call `utility` flows, but any other type can call a `utility` flow. You can run a `utility` flow without having a channel etc.

But a user who wants to send a message is going to reach for a messaging flow... feels odd maybe that you can make the same flow to send message over a channel in two ways, and the only difference is the type which the user can't change.

nicpottier:
Ya, sorry, shouldn't have used `type`. My main point there is that it is a mode of a flow (or a flag or whatever) that drives what you can do in it rather than vice versa.

rowanseymour:
Well having slept on this now I think I might have convinced myself _it is a type_. Putting aside UI concerns for now, the existing types in the flow spec are:

* `messaging`
* `messaging_offline`
* `voice`

And they are essentially subsets of actions. It's not the Surveyor type, it's the subset of messaging actions that can be run offline. Flow type is the thing that determines whether flow A can enter flow B, and it's how we filter lists of flows, so it would feel weird to complicate that with another field that does the same thing.

`messaging_passive` or `messaging_noninteractive` would seem natural as a type. It just requires some thought about how we present that in the UI - doesn't seem like sticking `Passive` in the type dropdown would be enough.

nicpottier:
👍 

`messaging_broadcast`
`messaging_onestep`
`messaging_nowait`
`messaging_instant`
`messaging_outbound`?

I don't like any of those FWIW

Agree that how we represent this in the UI could be tricky.

rowanseymour:
Or `messaging_background`... your preference is `messaging_passive` ?

nicpottier:
`_passive` is maybe the best worst? Could imagine that being used in the UI and that becoming "a thing".

rowanseymour:
I could see `Messaging` vs `Messaging (Non-interactive)` working in the UI which is why I'm slightly leaning toward that tho it's annoyingly long

nicpottier:
Ya, part of me recoils against something being `non-` as opposed to using just a clearer word. But even doing antonym searches couldn't find any great opposite of `interactive`. `passive` seemed best in that regard but agree it isn't necessarily as clear as `non-interactive`. Maybe for a type that's a ok and we can tweak the UI if we want?

ericnewcomer:
Why is naming always the hardest part. Caught up on this thread and I am happy with the conclusion of this being a type, as @rowanseymour mentioned, that is indeed what our current types are doing.

I find passive pretty odd for users, though I'm not super concerned with those names lining up with how we put it in code.

ericnewcomer:
My gut says `background` is closer to the right thing for users, but even that is pretty dang confusing for non-programmer types.

rowanseymour:
Ok well `messaging_offline` doesn't match `Surveyor` so think we can move forward with `messaging_passive` and hash out UI stuff later.

ericnewcomer:
Jen votes `utility` fwiw.

rowanseymour:
Background might work well in the UI because for users the most important thing to grasp is that these flows don't interrupt your other flows, and background would seem to convey that.

ericnewcomer:
I also do hate just calling them "quick" flows. Anyways, no reason to choose that word now. I'm fine with `passive` in code and we hash the rest out later.

PR #982, opened 2020-12-15
🛏️ Add new flow type for passive messaging flows

rowanseymour:
Currently defined as a messaging flow without waits

### DIFF
--- a/flows/definition/node.go
+++ b/flows/definition/node.go
@@ -43,14 +43,7 @@ func (n *node) Validate(flow flows.Flow, seenUUIDs map[uuids.UUID]bool) error {
 	for _, action := range n.Actions() {
 
 		// check that this action is valid for this flow type
-		isValidInType := false
-		for _, allowedType := range action.AllowedFlowTypes() {
-			if flow.Type() == allowedType {
-				isValidInType = true
-				break
-			}
-		}
-		if !isValidInType {
+		if !flow.Type().Allows(action) {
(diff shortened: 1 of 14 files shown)


flows/interfaces.go
```diff
@@ -224,6 +210,7 @@ type Timeout interface {
 // Wait tells the engine that the session requires input from the user
 type Wait interface {
 	utils.Typed
+	FlowTypeRestricted
```
rowanseymour, line 213: A theoretical wait type is the "nothing" wait type - used to get the session committed before moving on to something else. In theory it could be part of a passive flow. Anyways cross that bridge if we ever get to it!

flows/definition/node.go
```diff
@@ -43,14 +43,7 @@ func (n *node) Validate(flow flows.Flow, seenUUIDs map[uuids.UUID]bool) error {
 	for _, action := range n.Actions() {
 
 		// check that this action is valid for this flow type
-		isValidInType := false
-		for _, allowedType := range action.AllowedFlowTypes() {
-			if flow.Type() == allowedType {
-				isValidInType = true
```
rowanseymour, line 49: so ready for golang generics...

rowanseymour:
Seems like these should maybe also disallow `start_session` actions now that I think about it...

rowanseymour:
I guess allowing `start_session` actions brings us one step closer to "triggers as flows" ™ and finally "everything is a flow" ™

nicpottier (approved):

Issue #984, opened 2020-12-16
Exponent operator gives wrong results for non-integer exponents


rowanseymour:
Bug in decimal library (https://github.com/shopspring/decimal/issues/55) means only integer part of exponent is considered, thus `4 ^ 2.5 = 16`

PR #985, opened 2020-12-16
🧮 Add workaround for exponent expressions with non-integer exponents

rowanseymour:
`float64` workaround for #984


nicpottier (approved):
This means some loss of accuracy ya? (not that I think it matters)

rowanseymour:
Yeah tho in theory not at a level that makes any difference to our users

PR #987, opened 2021-01-07
start of ivr forward action/event

nicpottier:


flows/actions/forward_ivr.go
```diff
@@ -0,0 +1,60 @@
...
+//     "type": "forward_ivr",
+//     "phone": "+12065551212"
+//   }
+//
+// @action forward_ivr
+type ForwardIVRAction struct {
+	baseAction
+	voiceAction
+
+	Phone string `json:"phone" validate:"required" engine:"evaluated"`
```
nicpottier, line 31: @rowanseymour before I go to much farther on this, any strong feelings on this being represented as a URN vs a text named `phone`? I *think* phone numbers are the only valid value here and it should be an expression so `phone` felt more right here than say `urn`, but on the other hand for the evaluated event, `urn` definitely feels more right. Opinions?

rowanseymour: Just checking you know https://github.com/nyaruka/goflow/pull/972 exists?

rowanseymour: Agree event should be URN, and despite what I have in that PR, agree in shouldn't be a validated URN field on the action. Phone.. maybe tying our hands a little.. whatsapp voice forwards? Could be scheme/path like add_contact_urn action

```
Scheme string `json:"scheme" validate:"urnscheme"`
Path   string `json:"path" validate:"required" engine:"evaluated"`
```

nicpottier: Oh no, didn't know about #972.. doh. Will take that over.

I suppose I'm fine with Scheme/Path though wouldn't want to expose that in the UI.

rowanseymour: Yeah just phone in the editor for now

nicpottier: Although, thinking through this further having this be settable doesn't seem valid. I mean, you can't forward a physical phone call to whatsapp or vice versa (and it is hard to imagine us ever somehow supporting that), so why even open up the possibility of that being out of sync? Maybe we should just derive the scheme from the current contact scheme?

rowanseymour: True, then just `path`?

nicpottier: I'm ok with `path` but would still vote for `phone` just because it feels clearer. Your choice though.

rowanseymour:
I'm fine with phone. Don't want to overthink something that would be an
easy migration in the future anyway if anyone starts doing non-phone IVR.

PR #990, opened 2021-01-11
🐜 Add @node to context

rowanseymour:
Kinda frustrating that now anything which changes the generated docs creates so many file changes but here we are.

### DIFF
--- a/docs/en-us/completion.json
+++ b/docs/en-us/completion.json
@@ -301,6 +301,21 @@
             ]
         },
         {
+            "name": "node",
+            "properties": [
+                {
+                    "key": "uuid",
+                    "help": "the UUID of the node",
+                    "type": "text"
+                },
+                {
+                    "key": "visit_count",
+                    "help": "the count of visits to the node by this contact",
+                    "type": "number"
+                }
+            ]
+        },
+        {
             "name": "related_run",
             "properties": [
                 {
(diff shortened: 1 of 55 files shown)


docs/en-us/completion.json
```diff
@@ -300,6 +300,21 @@
...
+            "name": "node",
+            "properties": [
+                {
+                    "key": "uuid",
+                    "help": "the UUID of the node",
+                    "type": "text"
+                },
+                {
+                    "key": "visit_count",
+                    "help": "the count of visits to the node by this contact",
```
nicpottier, line 313: Worth clarifying it is in this run?

flows/runs/run.go
```diff
@@ -296,6 +298,27 @@ func (r *flowRun) Context(env envs.Environment) map[string]types.XValue {
...
+//
+// @context node
+func (r *flowRun) nodeContext(env envs.Environment) map[string]types.XValue {
+	_, node, _ := r.PathLocation()
+	visitCount := 0
+	for _, s := range r.path {
+		if s.NodeUUID() == node.UUID() {
+			visitCount++
+		}
+	}
```
nicpottier, line 314: Kinda feels to me like this should only be calculated when evaluated, iterating the entire path comparing UUIDs isn't the end of the world but then again it isn't nothing either.

nicpottier (approved):

flows/runs/run.go
rowanseymour: Most of the context is lazy loaded already using `flows.Context` or `flows.ContextFunc` - in this case it's loaded at the point of `@node` being accessed in the root.

nicpottier: Ah, ok, perfect.

PR #998, opened 2021-01-20
☎️ IVR redirects take III

rowanseymour:


flows/resumes/base.go
```diff
@@ -73,15 +73,34 @@ func (r *baseResume) Apply(run flows.FlowRun, logEvent flows.EventCallback) {
...
+// Expressions context
+//------------------------------------------------------------------------------------------
+
+// Context is the schema of trigger objects in the context, across all types
+type Context struct {
+	type_ string
+	dial  types.XValue
+}
+
+func (c *Context) asMap() map[string]types.XValue {
```
rowanseymour, line 86: copies same approach we use in triggers to enforce same schema across different types

test/testdata/runner/ivr_dial.json
```diff
@@ -0,0 +1,111 @@
...
+                                "exit_uuid": "52597931-6f0c-4fa3-9bdc-3c619664cb61"
+                            },
+                            {
+                                "uuid": "9f3cd249-c1be-4fe7-9503-e2bf2b988536",
+                                "name": "Failed",
+                                "exit_uuid": "fda29737-867c-45d7-89ab-f9c97a3f08cb"
+                            }
+                        ],
+                        "default_category_uuid": "9f3cd249-c1be-4fe7-9503-e2bf2b988536",
+                        "operand": "@(default(resume.dial.status, \"\"))",
```
rowanseymour, line 42: In the case that the wait can't make a valid tel URN, it skips itself so we fall through this router without a `@resume`, landing in the `Failed` default category. Wrapping this expression in `default()` ensures we don't generate a confusing expression error event.

nicpottier (approved):
+1 to `dial`. Almost wonder if `result` of the dial (`completed`, etc..) would make more sense than `status`, though we use `status` elsewhere for things very similar to this so probably better to stay consistent.

rowanseymour:
and `result` has a meaning already..

PR #999, opened 2021-01-20
🧮 Add @resume to context

rowanseymour:
Exposing just type for now. RapidPro will require https://github.com/nyaruka/temba-components/pull/49 as well.


flows/expressions.go
```diff
@@ -38,8 +38,10 @@ var RunContextTopLevels = []string{
 	"globals",
 	"input",
 	"legacy_extra",
+	"node",
```
rowanseymour, line 41: this is only needed for parsing of expressions without a real context - i.e. possible future flow migrations

nicpottier (approved):
What is `@resume` on a new run? null?

Does feel like long term a msg resume should maybe contain the message?

rowanseymour:
@nicpottier yes and yes was thinking the same

PR #1003, opened 2021-02-25
🌐 Add contact language to resthook payload

rowanseymour:


nicpottier (approved):
Looked at Zapier just now and I don't think we do any sort of rewriting of this side of things so I think this should "just work" once shipped.

rowanseymour:
My favorite kind of fix

nicpottier:
We should test after fixing to make sure though because there's certainly room for error there.

Issue #956, opened 2020-08-13
Retry email sending?


nicpottier:
Servers occasionally have connectivity issues, maybe we should retry emails a few times if they fail with a timeout between them. Essentially something similar to what we do with webhooks.

rowanseymour:
I've noticed the `go-mail` package actually does 1 retry with no timeout.. I think we'd disable that if we added our own retry stuff

rowanseymour:
Looked into this a bit - the challenge is knowing whether an err we get back from `DialAndSend` is retry-able. Our current SMTP library doesn't easily support that, and isn't the most maintained project right now so did start looking at whether it would be worth switching to https://github.com/jordan-wright/email and https://github.com/knadh/smtppool. We don't really need pooling but the latter library adds support for retrying.

nicpottier:
Just a quick note here that we saw this again, this time gmail SMTP server timing out and a customer was annoyed about it. So having some kind of retry on email sends may be worth looking into again.

PR #1007, opened 2021-03-29
📡 Add support for retrying SMTP sends

rowanseymour:
Adds #956


utils/smtpx/retries.go
```diff
@@ -0,0 +1,58 @@
...
+	return r.Backoffs[n]
+}
+
+// DefaultShouldRetry is the default function for determining if a send should be retried
+func DefaultShouldRetry(err error) bool {
+	code := extractCode(err)
+
+	// if error is a transient negative completion reply, it can be retried
+	// see https://en.wikipedia.org/wiki/List_of_SMTP_server_return_codes
+	if code >= 400 && code < 500 {
```
rowanseymour, line 39: @nicpottier take back I said about it not being clear which errors you can retry - apparently responses always start with a code and a 4xx code is fine to retry

rowanseymour: checked this with gmail SMTP responses

PR #1021, opened 2021-06-21
👥 Ticket assignees and trigger users as user assets

rowanseymour:
Adds a new user asset type which for now is just email+name and uses this for a new assignee field on tickets and for the user field on manual triggers

### DIFF
--- a/assets/interfaces.go
+++ b/assets/interfaces.go
@@ -287,6 +287,19 @@ type Ticketer interface {
 	Type() string
 }
 
+// User is an person who can trigger flows or be assigned tickets etc.
+//
+//   {
+//     "email": "bob@nyaruka.com",
+//     "name": "Bob"
+//   }
+//
+// @asset user
+type User interface {
+	Email() string
+	Name() string
+}
+
 // Source is a source of assets
 type Source interface {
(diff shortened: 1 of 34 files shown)


flows/tickets.go
```diff
@@ -14,27 +14,29 @@ type TicketUUID uuids.UUID
 
 // Ticket is a ticket in a ticketing system
 type Ticket struct {
-	uuid       TicketUUID `json:"uuid"`
```
rowanseymour, line 17: this is just crud from when this struct was unmarshaled directly

assets/interfaces.go
```diff
@@ -287,6 +287,19 @@ type Ticketer interface {
...
+//   {
+//     "email": "bob@nyaruka.com",
+//     "name": "Bob"
+//   }
+//
+// @asset user
+type User interface {
+	Email() string
+	Name() string
+}
```
rowanseymour, line 301: I've made users be assets even tho we don't really need them to be just yet because it seems inevitable that flows will need to know about users at some point - users will want to do ticket assignment in the actual flow action etc

norkans7 (approved):

PR #1026, opened 2021-07-02
👮 Strip invalid UTF-8 from webhook responses

rowanseymour:
`json.Valid` seems to only check structure even tho JSON "must" be valid UTF8 according to the [RFC](https://datatracker.ietf.org/doc/html/rfc8259#section-8.1)

https://play.golang.org/p/pXRiYg5sEPZ 

Had my suspicions that webhook responses were one place that invalid UTF-8 can get into the session state and sure enough it looks like https://sentry.io/organizations/nyaruka/issues/2483695525/events/5374e6c59d1d44f49fcc337469302de0/?project=1281372 is caused by a webhook result that includes `"id_number":"01\u00006\u00002"`.. NULL chars.

Initially was thinking we should reject invalid UTF-8 but seem like we can afford to be a bit more forgiving and strip out the invalid UTF-8.

### DIFF
--- a/services/webhooks/service_test.go
+++ b/services/webhooks/service_test.go
@@ -26,10 +26,10 @@ type call struct {
 func (c *call) String() string { return c.method + " " + c.url }
 
 type webhook struct {
-	request   string
-	response  string
-	body      string
-	validJSON bool
+	request  string
+	response string
+	body     string
+	bodyJSON string
 }
 
 func TestWebhookParsing(t *testing.T) {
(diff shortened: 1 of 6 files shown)


services/webhooks/service_test.go
```diff
@@ -48,64 +48,73 @@ func TestWebhookParsing(t *testing.T) {
...
+				bodyJSON: `{ "ok": "true" }`,
+			},
+		}, {
+			// successful POST with non-UTF8 response body
+			call: call{"POST", "http://127.0.0.1:49994/?cmd=badutf8", ""},
+			webhook: webhook{
+				request:  "POST /?cmd=badutf8 HTTP/1.1\r\nHost: 127.0.0.1:49994\r\nUser-Agent: goflow-testing\r\nContent-Length: 0\r\nAccept-Encoding: gzip\r\n\r\n",
+				response: "HTTP/1.1 200 OK\r\nContent-Length: 15\r\nContent-Type: text/plain\r\nDate: Wed, 11 Apr 2018 18:24:30 GMT\r\n\r\n",
+				body:     "{ \"bad\": \"\x80\x81\" }",
+				bodyJSON: `{ "bad": "" }`,
```
rowanseymour, line 90: so the invalid utf8 is stripped and if what remains is valid JSON.. we use that

PR #1027, opened 2021-07-08
Replace escaped nulls in webhook payloads before converting to JSON

rowanseymour:
So the situation with Postgres and NULL characters is weirder than I had assumed. A string with an actual null-char, an actual zero codepoint fails `json.Valid` so we never expose that to expressions. Since https://github.com/nyaruka/goflow/pull/1026 a string with invalid unicode sequences is turned into valid UTF8 before trying to turn it into JSON so that is also fixed...

But Postgres has its own unique limitation on JSON - it can't contain `\u0000` because it's going to try to unescape that when converting to JSONB. 

So with this PR we deal with null chars, escaped nulls and invalid unicode sequences:
  * In webhook events and HTTP logs these are replaced with `�`
  * For `@webhook` and `@results.foo.extra` those are stripped out before trying to parse as JSON


services/webhooks/service.go
```diff
@@ -61,10 +62,16 @@ func (s *service) Call(session flows.Session, request *http.Request) (*flows.Web
...
-			bodyUTF8 := bytes.ToValidUTF8(call.ResponseBody, nil)
-			if json.Valid(bodyUTF8) {
-				call.ResponseJSON = bodyUTF8
+			// we make a best effort to turn the body into JSON, so we strip out:
+			//  1. any invalid UTF-8 sequences
+			//  2. null chars
+			//  3. escaped null chars (\u0000)
+			cleaned := bytes.ToValidUTF8(call.ResponseBody, nil)
+			cleaned = bytes.ReplaceAll(cleaned, []byte{0}, nil)
+			cleaned = utils.ReplaceEscapedNulls(cleaned, nil)
```
rowanseymour, line 71: note that we're trying to extracting JSON from `call.ResponseBody` rather than `call.SanitizedResponse("...")` so we don't have to worry about things that have already been replaced with `�` - those are only being used for events and logs.

utils/text_test.go
```diff
@@ -186,3 +186,12 @@ func TestRedactor(t *testing.T) {
...
 	assert.Equal(t, "**** def **** jkl", utils.NewRedactor("****", "abc", "ghi")("abc def ghi jkl")) // all values redacted
 }
+
+func TestReplaceEscapedNulls(t *testing.T) {
+	assert.Equal(t, []byte(nil), utils.ReplaceEscapedNulls(nil, []byte(`?`)))
+	assert.Equal(t, []byte(`abcdef`), utils.ReplaceEscapedNulls([]byte(`abc\u0000def`), nil))
+	assert.Equal(t, []byte(`abc?def`), utils.ReplaceEscapedNulls([]byte(`abc\u0000def`), []byte(`?`)))
+	assert.Equal(t, []byte(`�ɇ�ɇ`), utils.ReplaceEscapedNulls([]byte(`\u0000\u0000`), []byte(`�ɇ`)))
+	assert.Equal(t, []byte(`abc  \\u0000 \\ \\\\u0000 def`), utils.ReplaceEscapedNulls([]byte(`abc \u0000 \\u0000 \\\u0000 \\\\u0000 def`), nil))
+	assert.Equal(t, []byte(`0000`), utils.ReplaceEscapedNulls([]byte(`\u00000000`), nil))
```
rowanseymour, line 196: JSON escapes are only 4 hex digits

nicpottier (approved):

PR #1028, opened 2021-07-12
🔌 Allow `msg` type triggers to have connections

rowanseymour:
Messages can trigger IVR flows but IVR flows can be run without a connection. 

Usually this isn't a problem because if a message triggers an IVR flow, that becomes a flow start which eventually is started in the engine with a manual trigger. We forget that it came from a message.

But in the simulator, if a message matches a keyword trigger and we want to simulate switching to that flow, we use a Msg type trigger... but we don't set connection.. so if that flow is IVR, we blow up, e.g. https://sentry.io/organizations/nyaruka/issues/2506463195/?project=1281372


nicpottier (approved):

rowanseymour:
@nicpottier tho on reflection just making simulation use a msg trigger + connection is not the short term fix we need otherwise simulation is going to differ from running an IVR flow in real life. We'll need to add `msg` to flow starts and use the correct trigger type for real IVR starts.

For now I think mailroom just triggers the IVR flow with a manual trigger + connection to match real life...

PR #1029, opened 2021-07-13
Words with hyphens should be considered as one word

norkans7:


norkans7:
I also verified we do not support hyphens as message form delimiter

rowanseymour:
Whilst I'm inclined to agree that `foo-bar` / `Nord-Kivu` should be considered one word everywhere.. how sure are we that we're not going to break something with this change? Changing the behavior of `utils.TokenizeString` also affects triggers.

norkans7:
I had checked 
We do not support triggers such as `foo-bar` already.
Should we allow messages such as `foo-bar` to match the `foo` keyword trigger?
If so I can add a different method `utils.TokenizeTriggerString` that use the previous behavior

The other probably incompatible effect is "has any of these words" with `foo-bar` rule, should `foo` be matched?

rowanseymour:
And related I do think `ex-wife` should be an allowed trigger keyword which would become possible if we made this switch.

norkans7:
Okay, so we keep these changes and make rapidpro as well support the same regex for keyword triggers

rowanseymour:
Not quite: we stop and think hard if the benefits of this change (I agree this change is good) is worth the risk of breaking existing stuff.

My inclination is that there's probably no-one relying on `@(word("a-b c", 1)) == b`.. but hard to be sure.

nicpottier:
The question I have is my "matches only phrase" does not match an operand of "north-kivu" and text  of "north-kivu". The splitting should be consistent so there should be a match no?

nicpottier:
So some tests show that they do match, but now I'm remembering that the problem is that "has only the phrase" with an operand of "kivu" matches "north-kivu", which I'm struggling with.

nicpottier:
Our docs for "has only the phrase" say it shouldn't. https://help.nyaruka.com/en/articles/3140560-response-rules

nicpottier:
No, I take that back, it is behaving correctly. "north-kivu" matches appropriately for "has only phrase" with two tests, one with "north-kivu" and one with "north". So what's the problem?

norkans7:
I think that is properly handled https://github.com/nyaruka/goflow/blob/main/flows/routers/cases/tests_test.go#L129

nicpottier:
I thought you said Caroline was doing this and it wasn't working?

norkans7:
The problem is that the text `foo-bar` is not matching "has only the phrase" `foo-bar`

nicpottier:
Screen Shot 2021-07-14 at 7 18 07 AM

It does match for me...

norkans7:
What is the value matched? Not the category only

norkans7:
@nicpottier I think you are right the issue probably happened using "has the phrase" and not "has only the phrase"
Experiment with `Kasai`, `Kasai-Central` and `Kasai-Oriental` all could match rule "has the phrase" "kasai" will match all 
I just added more tests for that

nicpottier:
That's expected though. `has phrase` means the words are present in that order. `has only the phrase` means all the words are present in that order but no others. I think everything is working right. The problem is they were using `has the phrase` so were matching the `north` case with a value of `north-kivu` which wouldn't occur if they were using `has only the phrase`

norkans7:
I believe `Toronto, we the north` text should not match "has the phrase" `north-kivu` similarly as 
`computerized equipment` should not match "has the phrase"  `computer`

nicpottier:
@norkans7 Neither of those match those as currently implemented.

norkans7:
So it is fine to have the text `north-kivu`  be matched for "has the phrase" `north`?

nicpottier:
Yes, that's correct. `has only the phrase` should be used otherwise.

norkans7:
That is clear now, I think this PR is not needed

PR #1030, opened 2021-07-16
🎏 Ticket Topics

rowanseymour:
This PR leaves subjects as required and adds topic as optional.. but I kinda like the idea that we migrate from subjects to topics and you need one of those.

Maybe every org gets a topic called `General` - and at least if we always have a topic, we always have something we can turn into a subject.


assets/interfaces.go
```diff
@@ -287,6 +287,22 @@ type Ticketer interface {
...
+//
+//   {
+//     "uuid": "cd48bd11-08b9-44e3-9778-8e26adf08a7a",
+//     "name": "Weather"
+//   }
+//
+// @asset topic
+type Topic interface {
+	UUID() TopicUUID
+	Name() string
```
rowanseymour, line 303: we already have `MsgTopic` as a thing in the engine.. but can imagine one day it's just another type of topic

PR #1035, opened 2021-07-30
💈 Simplify converting queries to elastic

rowanseymour:
Use resolver that query was parsed with no there is no possibility of erroring


contactql/es/testdata/to_query.json
```diff
@@ -112,26 +112,6 @@
             }
         }
     },
-    {
```
rowanseymour, line 115: these are parser errors which we test elsewhere

PR #1081, opened 2022-01-18
Contact querying by flow

rowanseymour:


flows/contact.go
```diff
@@ -467,16 +467,14 @@ func (c *Contact) ReevaluateQueryBasedGroups(env envs.Environment) ([]*Group, []
...
 func (c *Contact) QueryProperty(env envs.Environment, key string, propType contactql.PropertyType) []interface{} {
 	if propType == contactql.PropertyTypeAttribute {
 		switch key {
 		case contactql.AttributeUUID:
 			return []interface{}{string(c.uuid)}
-		case contactql.AttributeID:
-			if c.id != 0 {
-				return []interface{}{fmt.Sprintf("%d", c.id)}
-			}
-			return nil
```
rowanseymour, line 479: id and group aren't allowed in query based groups so this is never used in the real world

PR #1085, opened 2022-02-04
Fix group removal

rowanseymour:


flows/modifiers/base.go
```diff
@@ -46,10 +46,10 @@ func ReevaluateGroups(env envs.Environment, assets flows.SessionAssets, contact
 	if contact.Status() != flows.ContactStatusActive {
 		for _, g := range contact.Groups().All() {
 			if !g.UsesQuery() {
-				contact.Groups().Remove(g)
```
rowanseymour, line 49: modifying the thing we're iterating over..

PR #1087, opened 2022-03-03
`start_session` actions should generate error event if flow asset missing

rowanseymour:


test/testdata/runner/all_actions.test.json
```diff
@@ -109,7 +109,7 @@
                     ],
                     "created_on": "2018-07-06T12:30:17.123456789Z",
                     "flow": {
-                        "name": "Collect Language",
+                        "name": "Registration Flow",
```
rowanseymour, line 112: uses correct name from asset rather than just the reference on the flow action

PR #1088, opened 2022-03-14
Add `concat` excellent function

rowanseymour:
```
@(concat(array("a", "b"), array("c", "d"))) -> [a, b, c, d]
@(unique(concat(array(1, 2, 3), array(3, 4)))) -> [1, 2, 3, 4]
```

### DIFF
--- a/excellent/functions/builtin.go
+++ b/excellent/functions/builtin.go
@@ -107,7 +107,7 @@ func init() {
 		"sort":    OneArrayFunction(Sort),
 		"sum":     OneArrayFunction(Sum),
 		"unique":  OneArrayFunction(Unique),
-		"append":  TwoArrayFunction(Append),
+		"concat":  TwoArrayFunction(Concat),
 
 		// encoded text functions
 		"urn_parts":        OneTextFunction(URNParts),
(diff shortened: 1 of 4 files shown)


excellent/functions/builtin.go
```diff
@@ -1628,6 +1629,25 @@ func Unique(env envs.Environment, array *types.XArray) types.XValue {
 	return types.NewXArray(unique...)
 }
 
+// Append returns the result of appending one array to another as a new array.
+//
+//   @(append(array("a", "b"), array("c", "d"))) -> [a, b, c, d]
```
norkans7, line 1634: I think the second argument could be an element to append and not array 
or probably we rename the function to concat instead of append

`@(append(array("a", "b"), "c")) -> [a, b, c]`

norkans7: `@(concat(array("a", "b"), array("c", "d"))) -> [a, b, c, d]`

rowanseymour: I was thinking that `concat` sounded too much like a string operation but now I see Javascript uses `concat` for arrays so that's perfect https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

Re adding single elements.. since you can just the single element with `array()`.. not feeling that's required.

PR #1096, opened 2022-04-25
`send_msg` action should fallback to template translation in env default language

rowanseymour:
If no translation found for contact language. See https://groups.google.com/g/rapidpro-dev/c/-kR1MuvSdng/m/KNaLIlhzAwAJ

### DIFF
--- a/flows/actions/send_msg.go
+++ b/flows/actions/send_msg.go
@@ -97,10 +97,10 @@ func (a *SendMsgAction) Execute(run flows.Run, step flows.Step, logModifier flow
 
 		// do we have a template defined?
 		if a.Templating != nil {
-			// looks for a translation in these locales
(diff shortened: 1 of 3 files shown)


flows/actions/send_msg.go
```diff
@@ -97,10 +97,10 @@ func (a *SendMsgAction) Execute(run flows.Run, step flows.Step, logModifier flow
 
 		// do we have a template defined?
 		if a.Templating != nil {
-			// looks for a translation in these locales
+			// looks for a translation in the contact locale or environment default
 			locales := []envs.Locale{
-				run.Contact().Locale(run.Environment()),
 				run.Environment().DefaultLocale(),
```
rowanseymour, line 102: alas `run.Environment()` is the environment which is populated with values from the contact.. so this case both of these locales are the same

PR #1101, opened 2022-06-08
Add option to exclude contacts in a flow on start session action

rowanseymour:
Part 1 of https://github.com/rapidpro/rapidpro/issues/1662


flows/actions/start_session.go
```diff
@@ -28,6 +28,7 @@ const TypeStartSession string = "start_session"
 //     "groups": [
 //       {"uuid": "1e1ce1e1-9288-4504-869e-022d1003c72a", "name": "Customers"}
 //     ]
+//     "exclusions": {"in_a_flow": true}
```
rowanseymour, line 31: making this match the format we use to handle flow starts so easier to add more exclusions in future

PR #1119, opened 2022-07-26
Add contact status to variable context

norkans7:
This would allow to send the status in a webhook request for example


rowanseymour:
Did someone ask for this? Just curious, we kept it out of the context originally just in case we changed our minds about `status=blocked` vs `is_blocked=True` but I think we're now committed to `status`..

rowanseymour (approved):

PR #1125, opened 2022-09-06
Add keys function

norkans7:


rowanseymour:
@norkans7 the other way we could approach this is making `foreach` work with objects... 

```
foreach(obj, (k) => k & ": " & obj[k] & "\n")

foreach({"a": 1, "b": 2}) ->

a: 1
b: 2
```

norkans7:
I will try that approach

norkans7:
Replaced by #1126

PR #1126, opened 2022-09-07
Extend foreach to support objects and return object keys

norkans7:


excellent/functions/builtin.go
```diff
@@ -2104,17 +2104,32 @@ func ExtractObject(env envs.Environment, args ...types.XValue) types.XValue {
 // ForEach creates a new array by applying `func` to each value in `values`.
 //
 // If the given function takes more than one argument, you can pass additional arguments after the function.
+// If the first argument is an object, creates a new array by applying `func` to each key in the object.
 //
 //	@(foreach(array("a", "b", "c"), upper)) -> [A, B, C]
+//	@(foreach(object("a", 123, "b", "hello", "c", "world"), upper)) -> [A, B, C]
```
rowanseymour, line 2110: this returning an array makes me wonder.. if maybe your first idea of a `keys` function wasn't more intuitive - only I'd call it `properties` as that's how we refer to object props elsewhere.

norkans7: Why is an array an issue? 

`keys` seems more intuitive for users as it close to JSON keys

rowanseymour: Just seeing it, and I'm not convinced foreach(object) = array is intuitive... 

I guess Javascript has `Object.keys` and `Object.getOwnPropertyNames` and therefore `keys` wouldn't be too odd. And adding `keys` doesn't mean we can't also in future expand `foreach`. I like the brevity of `keys`.

norkans7: Should I rename `properties` to `keys` then https://github.com/nyaruka/goflow/pull/1127/files ?

norkans7: I guess users could do `foreach(keys(object))` if they really needed that.

norkans7:
replaced by https://github.com/nyaruka/goflow/pull/1127

PR #1127, opened 2022-09-09
Add support to get object properties

norkans7:


excellent/functions/builtin.go
```diff
@@ -136,6 +136,8 @@ func init() {
 		"extract_object": MinArgsCheck(2, ExtractObject),
 		"foreach":        MinArgsCheck(2, ForEach),
 		"foreach_value":  MinArgsCheck(2, ForEachValue),
+
+		"properties": OneArgFunction(Properties),
```
rowanseymour, line 140: I think make a wrapper `OneObjectFunction` like we have `OneArrayFunction`

excellent/functions/builtin.go
```diff
@@ -2101,6 +2103,32 @@ func ExtractObject(env envs.Environment, args ...types.XValue) types.XValue {
 	return types.NewXObject(result)
 }
 
+// Properties takes an object and returns its properties
+//
+//	@(properties(object("a", 123, "b", "hello", "c", "world"))) -> [a, b, c]
+//	@(properties(null)) -> []
```
rowanseymour, line 2109: I think this should be an error

excellent/functions/builtin.go
```diff
@@ -2101,6 +2103,32 @@ func ExtractObject(env envs.Environment, args ...types.XValue) types.XValue {
...
+//	@(properties("string")) -> ERROR
+//	@(properties(10)) -> ERROR
+//
+// @function properties(object)
+func Properties(env envs.Environment, arg types.XValue) types.XValue {
```
rowanseymour, line 2114: Let's go with `keys` to match JS's `Object.keys`

rowanseymour (changes requested):

excellent/functions/builtin.go
```diff
@@ -2101,6 +2103,28 @@ func ExtractObject(env envs.Environment, args ...types.XValue) types.XValue {
 	return types.NewXObject(result)
 }
 
+// Keys takes an object and returns its properties
```
rowanseymour, line 2106: Keys returns an array containing the property keys of `object`.

excellent/functions/builtin.go
```diff
@@ -2101,6 +2103,28 @@ func ExtractObject(env envs.Environment, args ...types.XValue) types.XValue {
...
+//
+//	@(keys(object("a", 123, "b", "hello", "c", "world"))) -> [a, b, c]
+//	@(keys("string")) -> ERROR
+//	@(keys(10)) -> ERROR
+//	@(keys(null)) -> ERROR
+//
+// @function keys(object)
+func Keys(env envs.Environment, object *types.XObject) types.XValue {
+
+	if object == types.XObjectEmpty {
```
rowanseymour, line 2116: should be valid to check an empty object.. return value is just `[]`

rowanseymour (changes requested):

excellent/functions/builtin.go
norkans7: nil is transformed to `XObjectEmpty `

https://github.com/nyaruka/goflow/blob/38f42b8822dcd533045051a42d0ddaccc48f2f7e/excellent/types/object.go#L241

So that as well return []

rowanseymour (approved):

excellent/functions/builtin.go
rowanseymour: I forgot that.. well ok

PR #1128, opened 2022-09-09
Add support to get object keys

norkans7:
Same as https://github.com/nyaruka/goflow/pull/1127 except the name of the function is `keys`


norkans7:
replaced by https://github.com/nyaruka/goflow/pull/1127

PR #1136, opened 2022-10-21
Remove merging of translation text arrays

rowanseymour:
It never makes sense

### DIFF
--- a/flows/runs/run_test.go
+++ b/flows/runs/run_test.go
@@ -1,6 +1,7 @@
 package runs_test
 
 import (
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


flows/runs/run_test.go
```diff
@@ -304,3 +307,122 @@ func TestSaveResult(t *testing.T) {
...
+			description:          "contact language is valid and is flow base language, msg action only has text",
+			envLangs:             []envs.Language{"eng", "spa"},
+			contactLang:          "eng",
+			msgAction:            msgAction2,
+			expectedText:         "Hello",
+			expectedAttachments:  []utils.Attachment{},
+			expectedQuickReplies: []string{},
+		},
+		{
+			description:  "contact language is valid and translations exist, msg action only has text",
```
norkans7, line 397: description seems wrong, `msg action has all fields` I think

PR #1162, opened 2023-02-27
Sort should return error (instead of panic) if array has mixed types

rowanseymour:
Fixes https://textit.sentry.io/issues/3960098152/?project=1281372

### DIFF
--- a/excellent/functions/builtin_test.go
+++ b/excellent/functions/builtin_test.go
@@ -570,7 +570,10 @@ func TestFunctions(t *testing.T) {
 		{"round_up", dmy, []types.XValue{}, ERROR},
 
 		{"sort", dmy, []types.XValue{xa()}, xa()},
+		{"sort", dmy, []types.XValue{xa(xn("3"))}, xa(xn("3"))},
 		{"sort", dmy, []types.XValue{xa(xn("3"), xn("1"), xn("2"))}, xa(xn("1"), xn("2"), xn("3"))},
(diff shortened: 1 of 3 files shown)


excellent/functions/builtin_test.go
```diff
@@ -570,7 +570,10 @@ func TestFunctions(t *testing.T) {
 		{"round_up", dmy, []types.XValue{}, ERROR},
 
 		{"sort", dmy, []types.XValue{xa()}, xa()},
+		{"sort", dmy, []types.XValue{xa(xn("3"))}, xa(xn("3"))},
 		{"sort", dmy, []types.XValue{xa(xn("3"), xn("1"), xn("2"))}, xa(xn("1"), xn("2"), xn("3"))},
+		{"sort", dmy, []types.XValue{xa(xs("C"), xs("B"), xs("A"))}, xa(xs("A"), xs("B"), xs("C"))},
+		{"sort", dmy, []types.XValue{xa(xs("X"), xn("4"))}, ERROR}, // different types
```
norkans7, line 576: Could it be a better experience for users to make `[]types.XValue{xa(xs("3"), xn("4"))` work?
just that case for string compared to numbers? we try to convert them if that is possible

rowanseymour: I thought about that but there's no good way to determine if you're sorted strings lexigrapically or numbers numerically or dates etc... i.e. what should you try to cast all values to.

norkans7 (approved):

PR #1166, opened 2023-05-02
Fix matching location name with diacritics

norkans7:


flows/routers/cases/tests_test.go
```diff
@@ -56,6 +56,11 @@ var locationHierarchyJSON = `{
 				}
 			]
 		},
+		{
```
norkans7, line 59: I just made up that Quebec belongs to Rwanda just to test the issue

rowanseymour (approved):

PR #1168, opened 2023-05-24
Add `contains` function

rowanseymour:
Note that this is only for arrays - our approach for text  functions thus far has been to make them distinct, e.g. `text_length`

### DIFF
--- a/excellent/functions/builtin.go
+++ b/excellent/functions/builtin.go
@@ -102,12 +102,13 @@ func init() {
 		"time_from_parts": ThreeIntegerFunction(TimeFromParts),
 
 		// array functions
-		"join":    TwoArgFunction(Join),
-		"reverse": OneArrayFunction(Reverse),
-		"sort":    OneArrayFunction(Sort),
-		"sum":     OneArrayFunction(Sum),
-		"unique":  OneArrayFunction(Unique),
-		"concat":  TwoArrayFunction(Concat),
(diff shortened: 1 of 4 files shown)


excellent/functions/builtin.go
```diff
@@ -1512,6 +1513,30 @@ func TimeFromParts(env envs.Environment, hour, minute, second int) types.XValue
...
+	array, xerr := types.ToXArray(env, arg1)
+	if xerr != nil {
+		return xerr
+	}
+	if types.IsXError(value) {
+		return value
+	}
+
+	for i := 0; i < array.Count(); i++ {
+		if types.Equals(array.Get(i), value) {
```
rowanseymour, line 1532: note that this isn't using "textual equality" that we use for `x = y`.. but it's consistent with how other array functions work like `unique`. I guess we could add "true equality" as `x == y` and then say that functions like this use that.

excellent/functions/builtin.go
```diff
@@ -1631,7 +1656,7 @@ func Unique(env envs.Environment, array *types.XArray) types.XValue {
 
 		seen := false
 		for j := 0; j < len(unique); j++ {
-			if (val == nil && unique[j] == nil) || types.Equals(val, unique[j]) {
```
rowanseymour, line 1634: not needed as `types.Equals` already handles nils

PR #1184, opened 2023-09-14
OptIn assets and triggers

rowanseymour:
So that in a flow triggered by an optin or optout, you can do `@trigger.optin`, `@trigger.optin.name` etc 

And whatever action we add to solicit an optin needs to know about optins.

### DIFF
--- a/assets/optin.go
+++ b/assets/optin.go
@@ -1,0 +1,65 @@
+package assets
+
+import (
+	"fmt"
+
+	"github.com/nyaruka/gocommon/uuids"
+)
+
+// OptInUUID is the UUID of an opt in
+type OptInUUID uuids.UUID
+
+// OptIn are channel specific opt-ins
+//
+//	{
+//	  "uuid": "8925c76f-926b-4a63-a6eb-ab69e7a6b79b",
+//	  "name": "Joke Of The Day",
+//	  "channel": {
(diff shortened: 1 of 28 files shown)


assets/optin.go
```diff
@@ -0,0 +1,65 @@
...
+type OptInUUID uuids.UUID
+
+// OptIn are channel specific opt-ins
+//
+//	{
+//	  "uuid": "8925c76f-926b-4a63-a6eb-ab69e7a6b79b",
+//	  "name": "Joke Of The Day",
+//	  "channel": {
+//	     "uuid": "204e5af9-42c3-4d46-8aab-ce204dff25b4",
+//	     "name": "Facebook"
```
rowanseymour, line 19: making optins channel specific.. because we're going to associate auth keys with optins and they are obviously channel specific

assets/static/channel.go
```diff
@@ -19,7 +19,7 @@ type Channel struct {
 }
 
 // NewChannel creates a new channel
-func NewChannel(uuid assets.ChannelUUID, name string, address string, schemes []string, roles []assets.ChannelRole, parent *assets.ChannelReference) assets.Channel {
```
rowanseymour, line 22: unrelated cleanup

PR #1197, opened 2023-11-24
Rework template translation finding

rowanseymour:
So that it can work with locales that we get from WhatsApp like `en-US` using a smart matcher and so we don't have to map those on the RP side and then map them back in courier.

### DIFF
--- a/flows/template_test.go
+++ b/flows/template_test.go
@@ -1,11 +1,14 @@
-package flows
+package flows_test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/nyaruka/gocommon/i18n"
 	"github.com/nyaruka/goflow/assets"
 	"github.com/nyaruka/goflow/assets/static"
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/test"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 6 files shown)


flows/template_test.go
```diff
@@ -23,85 +26,85 @@ func TestTemplateTranslation(t *testing.T) {
...
 		},
 		{
-			"c520cbda-e118-440f-aaf6-c0485088384f",
-			nil,
-			[]i18n.Locale{"deu-DE", "spa-ES"},
+			channel1,
+			[]i18n.Locale{"deu-DE"},
 			[]string{"Chef"},
-			"",
+			"Hello Chef",
```
rowanseymour, line 78: this case is a little weird because it's just defaulting to the first translation... but it does seem better to generally return something here than creating a message without a template that probably going to fail.

norkans7:
the language from WA are like `en_GB`, `es_MX` not exactly `en-GB` and `es-MX`, so we substitute  the underscore with the hyphen somewhere?

norkans7 (approved):

rowanseymour:
Been thinking about this a bit more I think we when syncing we should store _both_ the original language code, and our current `language` (3 chars) + `country` fields on the template translation object. That way the engine can continue to work with `xxx` or `xxx-YY` format locales, but in mailroom, when queuing the message to courier, we can lookup the original language code on the translation object, and queue that.

PR #1199, opened 2023-12-01
WIP support WA template params

norkans7:


assets/static/template.go
```diff
@@ -43,16 +43,20 @@ type TemplateTranslation struct {
 	Locale_        i18n.Locale              `json:"locale"          validate:"required"`
 	Namespace_     string                   `json:"namespace"`
 	VariableCount_ int                      `json:"variable_count"`
+	Components_    []map[string]any         `json:"components"`
+	Params_        map[string][]any         `json:"params"`
```
rowanseymour, line 47: no reason not to use a struct for param

assets/static/template.go
```diff
@@ -43,16 +43,20 @@ type TemplateTranslation struct {
 	Locale_        i18n.Locale              `json:"locale"          validate:"required"`
 	Namespace_     string                   `json:"namespace"`
 	VariableCount_ int                      `json:"variable_count"`
+	Components_    []map[string]any         `json:"components"`
```
rowanseymour, line 46: we only use Content for generating a preview of the message so I guess components would be the same... couldn't these just be strings?

norkans7: I think we can keep using content for the preview here and make sure each template has that generate properly when syncing them

assets/static/template.go
norkans7: Added `TemplateParam`

assets/static/template.go
rowanseymour: Tho wouldn't we want to convert button components into quick replies etc? I kinda hate putting whatsapp specific knowledge in the engine, right in the send_msg action but I don't see any other way. Let's just try to isolate it in a single function for now that takes a template + parameters and returns a `MsgOut`.

The other way to approach this would be to put the onus on the UI to render a preview of the message based on the templating data.. but then the message still looks empty in API fetches, archives etc, so I think I prefer having the engine continue to create a preview.

assets/template.go
```diff
@@ -42,13 +42,24 @@ type Template interface {
 	Translations() []TemplateTranslation
 }
 
+// TemplateParamUUID is the UUID of a param
+type TemplateParamUUID uuids.UUID
+
+// TemplateParam is a parameter for template translation
+type TemplateParam struct {
+	Type  string            `json:"type"`
+	UUID  TemplateParamUUID `json:"uuid"`
```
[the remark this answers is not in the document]
rowanseymour: This is the template _asset_ which isn't localized and doesn't have values. Shouldn't this just be `type` ?

And then we have a different struct on `templating` on `send_msg` actions.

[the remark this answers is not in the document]
norkans7: Yeah, I was thinking on where the UUID should be generated and I think that should be on the UI in the flow editor when the template is selected.

I will look into that again

flows/actions/send_msg.go
```diff
@@ -140,3 +128,53 @@ func (a *SendMsgAction) Execute(run flows.Run, step flows.Step, logModifier flow
...
+				paramValue = evaluatedQuickReplies[qrIndex]
+				qrIndex++
+			} else if templateParam.Type() != "text" {
+				paramValue = ""
+				for _, att := range evaluatedAttachments {
+					attType := strings.Split(att.ContentType(), "/")[0]
+					if templateParam.Type() == attType {
+						paramValue = att.URL()
+						break
+					}
```
norkans7, line 162: Use first attachment matching the type we expect from the template params

flows/msg.go
```diff
@@ -168,11 +168,40 @@ func (m *MsgOut) Locale() i18n.Locale { return m.Locale_ }
 // UnsendableReason returns the reason this message can't be sent (if any)
 func (m *MsgOut) UnsendableReason() UnsendableReason { return m.UnsendableReason_ }
 
+// TemplateParamUUID is the UUID of a param
+type TemplateParamUUID uuids.UUID
+
+// TemplateParam represents a single parameter
+type TemplateParam struct {
+	Type_  string            `json:"type"`
```
rowanseymour, line 176: Do we need `type` here? Wondering if each `Msg.Templating` could be stored as something like...

```go
type ComponentVariables struct {
    UUID uuids.UUID
    Variables []string
}

map[string]ComponentVariables
```

and then the localization can be of the `ComponentVariables` instead of each param.

norkans7: I think using `Variables []string` we are loosing the type for the parameter/variable and that is essential to know I think

rowanseymour: Well let's really think... do we need it? Where?

norkans7: For example when we have the expected parameters as an image, video or document attachment

https://github.com/nyaruka/goflow/pull/1199/files#diff-f146c45610389d4ec808d1c8edafab8ca680779f7da5a7393ba3e8c3b86f427aR158

norkans7:
Replaced by other PRs

PR #1204, opened 2024-01-17
Replace flows.MsgTemplating variables by params

norkans7:


flows/msg.go
```diff
@@ -168,28 +168,41 @@ func (m *MsgOut) Locale() i18n.Locale { return m.Locale_ }
 // UnsendableReason returns the reason this message can't be sent (if any)
 func (m *MsgOut) UnsendableReason() UnsendableReason { return m.UnsendableReason_ }
 
+type ComponentVariable struct {
```
rowanseymour, line 171: let's call this `TemplateParam`

rowanseymour (changes requested):

flows/msg.go
norkans7: I was trying to make a distinction between that and 
```type TemplateParam struct {
	Type_ string `json:"type"`
}```
That we'll use in the template translations

But I just renamed that to `TemplateParam`

rowanseymour: well they're definitely params now rather than variables...maybe it's ok for them to have same name as they'll be in different packages

PR #1205, opened 2024-01-17
Replace relying on variables to generate new templating

norkans7:


flows/actions/send_msg.go
```diff
@@ -122,7 +122,16 @@ func (a *SendMsgAction) Execute(run flows.Run, step flows.Step, logModifier flow
 				}
 
 				evaluatedText = translation.Substitute(evaluatedVariables)
-				templating = flows.NewMsgTemplating(a.Templating.Template, evaluatedVariables, translation.Namespace())
+
+				params := map[string][]flows.TemplateParam{}
```
norkans7, line 126: Just moved building the params here and pass that to the `flows.NewMsgTemplating` instead of passing a list of variables

flows/actions/send_msg.go
```diff
@@ -122,7 +122,18 @@ func (a *SendMsgAction) Execute(run flows.Run, step flows.Step, logModifier flow
 				}
 
 				evaluatedText = translation.Substitute(evaluatedVariables)
-				templating = flows.NewMsgTemplating(a.Templating.Template, evaluatedVariables, translation.Namespace())
+
+				params := make(map[string][]flows.TemplateParam, 1)
```
norkans7, line 126: Just moved building the params here and pass that to the flows.NewMsgTemplating instead of passing a list of variables

flows/actions/send_msg.go
```diff
@@ -122,7 +122,18 @@ func (a *SendMsgAction) Execute(run flows.Run, step flows.Step, logModifier flow
...
+
+				// TODO add support for params in other components besides body
+				if len(evaluatedVariables) > 0 {
```
norkans7, line 129: I think here is the right place since we can have access to the param types in the template translations once we support other components and parameter types

PR #1206, opened 2024-01-18
Load template translations params

norkans7:


assets/static/template.go
```diff
@@ -38,21 +38,23 @@ func (t *Template) Translations() []assets.TemplateTranslation {
...
-	Channel_       *assets.ChannelReference `json:"channel"         validate:"required"`
-	Content_       string                   `json:"content"         validate:"required"`
-	Locale_        i18n.Locale              `json:"locale"          validate:"required"`
-	Namespace_     string                   `json:"namespace"`
-	VariableCount_ int                      `json:"variable_count"`
+	Channel_       *assets.ChannelReference   `json:"channel"         validate:"required"`
+	Content_       string                     `json:"content"         validate:"required"`
+	Locale_        i18n.Locale                `json:"locale"          validate:"required"`
+	Namespace_     string                     `json:"namespace"`
+	VariableCount_ int                        `json:"variable_count"`
```
rowanseymour, line 45: Can't we get rid of variable count? Isn't it just `len(params["body"])` ?

norkans7: Yes those are the same, I have not seen where that variable is used actually, I guess probably with the flow_editor, so we can cleanup that once we are sure it is not relied on

rowanseymour: I mean get rid of it here if the engine doesn't need it

norkans7: Okay

PR #1208, opened 2024-01-18
Support more components

norkans7:


flows/actions/send_msg.go
```diff
@@ -50,11 +54,18 @@ type SendMsgAction struct {
...
+
 // Templating represents the templating that should be used if possible
 type Templating struct {
-	UUID      uuids.UUID                `json:"uuid" validate:"required,uuid4"`
-	Template  *assets.TemplateReference `json:"template" validate:"required"`
-	Variables []string                  `json:"variables" engine:"localized,evaluated"`
+	UUID      uuids.UUID                            `json:"uuid" validate:"required,uuid4"`
+	Template  *assets.TemplateReference             `json:"template" validate:"required"`
+	Variables []string                              `json:"variables" engine:"localized,evaluated"`
+	Params    map[string]TemplateComponentVariables `json:"params"`
```
rowanseymour, line 68: Since we're now talking about how the editor will store things in a flow definition.. let's hear from @ericnewcomer  if any of the following solutions make more sense than others.

Remember that localizable things in flow definitions are named properties of an object with a UUID, and every localization is an array of strings (tho often it's a single item array) because the field is a single item.

And that the editor is going to see template objects that look like:

```json
{
  "uuid": "b7b35db9-d914-4cd3-a170-46eab02531bb",
  "params": {
    "body": [{"type": "text"}, {"type": "text"}],
    "button.0": [{"type": "text"}]
  }
}
```

So first way we could represent param values is that each is a localizable single value (I think this how you first had it in code):

```json
{
  "type": "send_msg",
  "text": "Register?",
  "templating": {
    "uuid": "5e7229b1-d196-42c3-9b08-125ab80c88ea",
    "template": {
      "uuid": "bd9d983c-c0eb-4c4e-b688-3af737c410c6",
      "name": "Register"
    },
    "params": {
      "body": [
        {
          "uuid": "1067f8e2-82f0-4378-9214-0f019365ddb7",
          "value": "@contact"
        },
        {
          "uuid": "7d771d7b-aab6-4cc6-a3df-360506f759de",
          "value": "sales"
        }
      ]
      "button.0": [
        {
          "uuid": "b7b35db9-d914-4cd3-a170-46eab02531bb",
          "value": "OK"
        }
      ]
    }
  }
}
```

Or each component is the localizable object (this is how the code above has it): 

```json
{
  "type": "send_msg",
  "text": "Register?",
  "templating": {
    "uuid": "5e7229b1-d196-42c3-9b08-125ab80c88ea",
    "template": {
      "uuid": "bd9d983c-c0eb-4c4e-b688-3af737c410c6",
      "name": "Register"
    },
    "params": {
      "body": {
        "uuid": "1067f8e2-82f0-4378-9214-0f019365ddb7",
        "values": ["@contact", "sales"]
      },
      "button.0": {
        "uuid": "b7b35db9-d914-4cd3-a170-46eab02531bb",
        "values": ["OK"]
      }
    }
  }
}
```

Or make `params` the localizable object assuming no component can be called `uuid`:

```json
{
  "type": "send_msg",
  "text": "Register?",
  "templating": {
    "uuid": "5e7229b1-d196-42c3-9b08-125ab80c88ea",
    "template": {
      "uuid": "bd9d983c-c0eb-4c4e-b688-3af737c410c6",
      "name": "Register"
    },
    "params": {
      "uuid": "1067f8e2-82f0-4378-9214-0f019365ddb7",
      "body": ["@contact", "sales"],
      "button.0": ["OK"]
    }
  }
}
```

Or use `templating` as the localizable object (like how it is now with just `variables`). Probably safe as there's no component that's gonna be called `template` or `uuid` right?

```json
{
  "type": "send_msg",
  "text": "Register?",
  "templating": {
    "uuid": "5e7229b1-d196-42c3-9b08-125ab80c88ea",
    "template": {
      "uuid": "bd9d983c-c0eb-4c4e-b688-3af737c410c6",
      "name": "Register"
    },
    "body": ["@contact", "sales"],
    "button.0": ["OK"]
  }
}
```

So many ways to skin a cat as they say...

ericnewcomer:
I like option 3 (the params object with uuid), feels the most natural and easiest to work with to me.

On Fri Jan 19, 2024, 06:41 PM GMT, Rowan Seymour ***@***.***> wrote:
> @rowanseymour commented on this pull request.
------------------------------

flows/actions/send_msg.go
rowanseymour: So option 3. Note that we can also drop the UUID on `templating` once we get rid of the `variables` property because it was only there for localizing that.

norkans7: I found that I could use 
```
Params    map[string][]string       `json:"params"`
```

And use a slight modified version of option 3 
```
{
  "type": "send_msg",
  "text": "Register?",
  "templating": {
    "uuid": "5e7229b1-d196-42c3-9b08-125ab80c88ea",
    "template": {
      "uuid": "bd9d983c-c0eb-4c4e-b688-3af737c410c6",
      "name": "Register"
    },
    "params": {
      "body": ["@contact", "sales"],
      "button.0": ["OK"]
    }
  }
}
```

So we rely on `templating` uuid for translations but not have uuid in params so we can use 
```
Params    map[string][]string       `json:"params"`
```

rowanseymour: @norkans7 I don't think that works.. or at least without changes to the editor. Thus far localized things have always been direct properties of an object with a UUID. In your example `body` is not a direct property of a localizable thing.

norkans7: since the keys for params are not easy to predictI do not think we can have a uuid as a string and the others having list values

rowanseymour: You mean there could be a component called uuid? I'm not following..

norkans7: No I mean the keys are dynamic one template could have `body`, `button0`, `header`, `footer`..
another could have many buttons `body`, `button0`, `button1`, `button2`

so params can not have the uuid key mixed with those keys as its value is a string and we want the value for the other to be a list of string without knowing the exact keys name in advance

rowanseymour: Ahhhh ok now I get you - there's no way to represent that params as a simple go structure. Hmm ok 2 options...

1. Hardcoded component names.. the thing we've been trying to avoid:

```go
type Params struct {
    UUID    uuids.UUID `json:"uuid"`
    Header  []string   `json:"header,omitempty"`
    Body    []string   `json:"header,omitempty"`
    Button0 []string   `json:"button.0,omitempty"`
    Button1 []string   `json:"button.1,omitempty"`
    Button2 []string   `json:"button.2,omitempty"`
    Button3 []string   `json:"button.3,omitempty"`
    Button4 []string   `json:"button.4,omitempty"`
    Button5 []string   `json:"button.5,omitempty"`
    Button6 []string   `json:"button.6,omitempty"`
    Button7 []string   `json:"button.7,omitempty"`
    Button8 []string   `json:"button.8,omitempty"`
    Button9 []string   `json:"button.9,omitempty"`
}
```

2. Custom JSON marshaling / unmarshaling:

```go
type Params struct {
    UUID uuids.UUID `json:"uuid"`
    Values map[string][]string
}

func (p *Params) MarshalJSON() ([]byte, error) {
    m := make(map[string]any, 1 + len(p.Values))
    m["uuid"] = p.uuid
    for k, v := range p.Values {
        m[k] = v
    }
    return json.Marshal(m)
}

func (p *Params) UnmarshalJSON(d []byte) error {
    var m map[string]any
    if err := json.Unmarshal(d, &m); err != nil {
        return err
    }
    p.Values := make(map[string][]string, len(m) - 1)
    for k, v := range m {
        switch typed := v.(type) {
            case string:
                if k == "uuid" {
                     p.UUID = typed
                }
            case []string:
                p.Values[k] = typed
        }
    }
    return nil
}
```

@ericnewcomer can you make this work on the typescript side of things?

ericnewcomer: Ya, I don't see any reason why that should be a problem on the front end.

norkans7: https://github.com/nyaruka/goflow/pull/1209 has those adjustments

norkans7:
Replaced by #1209

PR #1209, opened 2024-01-24
Support more WA templates components

norkans7:
Replaces #1208

### DIFF
--- a/flows/actions/send_msg.go
+++ b/flows/actions/send_msg.go
@@ -1,6 +1,9 @@
 package actions
 
 import (
+	"strconv"
+	"strings"
+
 	"github.com/nyaruka/gocommon/i18n"
 	"github.com/nyaruka/gocommon/urns"
 	"github.com/nyaruka/gocommon/uuids"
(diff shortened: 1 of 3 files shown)


flows/actions/testdata/send_msg.json
```diff
@@ -347,6 +347,7 @@
             "text": "Hi @contact.name, who's a good boy?",
             "templating": {
                 "uuid": "9c4bf5b5-3aa4-48ec-9bb9-424a9cbc6785",
+                "params": null,
```
rowanseymour, line 350: should be omitted I think

flows/actions/send_msg.go
```diff
@@ -128,27 +174,65 @@ func (a *SendMsgAction) Execute(run flows.Run, step flows.Step, logModifier flow
...
+	evaluatedParams := make(map[string][]string)
+
+	templateTranslationParams := templateTranslation.Params()
 
-	evaluatedVariables := make([]string, len(localizedVariables))
-	for i, variable := range localizedVariables {
-		sub, err := run.EvaluateTemplate(variable)
-		if err != nil {
-			logEvent(events.NewError(err))
+	if _, ok := templateTranslationParams["body"]; ok {
```
rowanseymour, line 181: I'm confused by this... how does this work when variables is set but params isn't?

norkans7: `templateTranslationParams` in the params for WA templates that gives us a hint we should have variables for the components

```
{
  "uuid": "b7b35db9-d914-4cd3-a170-46eab02531bb",
  "params": {
    "body": [{"type": "text"}, {"type": "text"}],
    "button.0": [{"type": "text"}]
  }
}
```

so we use that and then check the old variables since the editor do not yet have support for setting params in templating

Even when the editor does below we wil overwrite the body params

PR #1211, opened 2024-01-25
Templating fixes

rowanseymour:
* Fix logic for evaluating and localizing component params
 * Include template param values in expression enumeration
 * Don't include quick replies or attachments on preview messages for now since we don't have their text content to do variable substitution on

### DIFF
--- a/flows/actions/send_msg.go
+++ b/flows/actions/send_msg.go
@@ -9,6 +9,7 @@ import (
 	"github.com/nyaruka/goflow/assets"
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/flows/events"
+	"github.com/nyaruka/goflow/utils"
 )
 
 func init() {
@@ -57,8 +58,19 @@ type TemplateParams struct {
 	Values map[string][]string
 }
 
-// LocalizationUUID gets the UUID which identifies this object for localization
-func (p *TemplateParams) LocalizationUUID() uuids.UUID { return uuids.UUID(p.UUID) }
+func (p *TemplateParams) EnumerateTemplates(localization flows.Localization, include func(i18n.Language, string)) {
+	for _, comp := range utils.SortedKeys(p.Values) {
+		for _, v := range p.Values[comp] {
+			include(i18n.NilLanguage, v)
+		}
+		for _, lang := range localization.Languages() {
+			lvals := localization.GetItemTranslation(lang, p.UUID, comp)
+			for _, v := range lvals {
+				include(lang, v)
+			}
+		}
+	}
+}
 
 func (p *TemplateParams) MarshalJSON() ([]byte, error) {
 	if p == nil {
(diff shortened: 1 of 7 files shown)


flows/actions/send_msg.go
```diff
@@ -191,52 +202,42 @@ func getTemplateMsg(action *SendMsgAction, run flows.Run, urn urns.URN, channelR
...
-			var evaluatedComponentVariables []string
-			if strings.HasPrefix(compKey, "button.") {
-				qrIndex, err := strconv.Atoi(strings.TrimPrefix(compKey, "button."))
+			for i, variable := range localizedCompParams {
+				sub, err := run.EvaluateTemplate(variable)
 				if err != nil {
 					logEvent(events.NewError(err))
 				}
-				paramValue := evaluatedQuickReplies[qrIndex]
-				evaluatedComponentVariables = []string{paramValue}
```
rowanseymour, line 206: I don't get this at all...

flows/actions/testdata/send_msg.json
```diff
@@ -419,21 +419,17 @@
...
                     ],
                     "uuid": "1067f8e2-82f0-4378-9214-0f019365ddb7"
-                },
-                "template": {
-                    "uuid": "5722e1fd-fe32-4e74-ac78-3cf41a6adb7e",
-                    "name": "affirmation"
-                },
-                "variables": [
-                    "@contact.name",
-                    "boy"
```
rowanseymour, line 435: bit weird having tests where there is both `params` and `variables`.. that shouldn't happen

flows/actions/send_msg.go
norkans7: I think users expected quick replies on the node to be used on the WA templates, so if the template has button that need variables we use the quick replies values set on the node, similarly for attachments

flows/actions/testdata/send_msg.json
norkans7: I thought we support both until the editor is updated, alternatively we need to have the flow migrations to make sure only use params

flows/actions/send_msg.go
rowanseymour: But we've never supported this because we've never supported templates with buttons..

flows/actions/testdata/send_msg.json
rowanseymour: Yes the editor should support flows with either, but there will never be a single flow that contains both.

flows/actions/send_msg.go
norkans7: I think that is what we want to add support for.
The confusion user have been having is that the expect the quick replies to be used for templates.
We have had to clarify that we do not yet support those components yet.

We could also force the user so always specify the parameter for the templates in addition to the quick replies

rowanseymour: I think when we support button components it will be more obvious to the user. But it's important that the preview message we generate in this case is similar to what will be sent via the WhatsApp template. There's no point adding quick replies to the preview message if the template doesn't use buttons.

Currently we only have `Content()` which is the body text... I think we need the button texts so we can perform `{{1}}` variable substitution on those... but let's worry about that later.

PR #1213, opened 2024-02-12
Simplify generating template message preview

rowanseymour:


flows/actions/testdata/_assets.json
```diff
@@ -379,20 +379,16 @@
...
                         },
                         "button.1": {
-                            "content": "",
-                            "params": [
-                                {
-                                    "type": "text"
-                                }
-                            ]
+                            "content": "No",
+                            "params": []
```
rowanseymour, line 391: so that we can test a component having content but no params - it should be used as a quick reply value in this case but doesn't need to be included in `msg.templating` because there are no param values to send to Whatsapp

PR #1214, opened 2024-02-13
Make `@webhook` an object with `headers` and `json` properties

rowanseymour:
This takes away recreating webhook responses after waits... so we need to resolve that before moving forward with this.


flows/runs/legacy.go
```diff
@@ -96,24 +95,3 @@ func arrayToObject(array *types.XArray) *types.XObject {
 	}
 	return types.NewXObject(properties)
 }
-
-// finds the last webhook response that was saved as extra on a result
-func lastWebhookSavedAsExtra(r *run) types.XValue {
```
rowanseymour, line 101: this is a potentially backwards compat breaking change if someone is relying on small webhook responses being available between sprints... but I think it's worth ripping off the bandaid because 1) we've always told users that `@webhook` isn't persisted between sprints and 2) it's unpredictable in that if the response goes over 10,000 chars, this isn't set.

norkans7: So they need to save whatever is needed from the webhook in a result variables so they can use those later, right?

norkans7 (approved):

flows/actions/testdata/call_resthook.json
```diff
@@ -242,18 +254,15 @@
...
         ],
         "webhook": {
-            "errors": [
-                "service unavailable"
-            ]
+            "__default__": "http://unavailable.com/",
+            "headers": {
+                "Content-Type": "text/plain"
+            },
+            "json": null
```
rowanseymour, line 265: cuz it's not json

PR #1220, opened 2024-02-27
Expression migrations

rowanseymour:
Adds the code to support expression migrations.. but doesn't yet add a new migration.

The idea is that for every flow spec version, we have a catalog of jsonpaths that tell us where expressions can be found in a flow definition.


flows/definition/migrations/specdata/templates.json
```diff
@@ -0,0 +1,91 @@
...
+            ],
+            "send_email": [
+                ".addresses[*]",
+                ".body",
+                ".subject"
+            ],
+            "send_msg": [
+                ".attachments[*]",
+                ".quick_replies[*]",
+                ".templating.variables[*]",
```
rowanseymour, line 53: this is missing template params... and I'm struggling to make that work and starting to wonder about a different format

PR #1232, opened 2024-03-19
Add component type to msg templating

norkans7:


flows/actions/testdata/send_msg.json
```diff
@@ -551,8 +553,9 @@
                             "uuid": "5722e1fd-fe32-4e74-ac78-3cf41a6adb7e",
                             "name": "affirmation"
                         },
-                        "params": {
-                            "body": [
+                        "components": [{
```
rowanseymour, line 556: you don't have auto formatting of JSON enabled ?

flows/actions/send_msg.go
```diff
@@ -156,25 +158,33 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
 			if i < len(evaluatedParams[key]) {
-				params[key][i] = flows.TemplateParam{Type: tp.Type(), Value: evaluatedParams[key][i]}
+				params[i] = flows.TemplateParam{Type: tp.Type(), Value: evaluatedParams[key][i]}
 			} else {
-				params[key][i] = flows.TemplateParam{Type: tp.Type(), Value: ""}
+				params[i] = flows.TemplateParam{Type: tp.Type(), Value: ""}
 			}
 		}
+		if len(params) > 0 {
+			components = append(components, flows.TemplateComponent{Type: comp.Type(), Name: key, Params: params})
```
rowanseymour, line 181: don't need name there

flows/actions/send_msg.go
```diff
@@ -156,25 +158,33 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
+		}
+
 	}
 
 	locale := translation.Locale()
-	templating := flows.NewMsgTemplating(a.Templating.Template, params, translation.Namespace())
```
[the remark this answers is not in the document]
norkans7: Right, I was checking on a strategy for that. Thanks

flows/actions/send_msg.go
```diff
@@ -155,26 +157,51 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
 		}
+	}
+
+	// next we cross reference with params defined in the template translation to get types
+	components := make([]flows.TemplateComponent, 0)
+
+	translationComponents := translation.Components()
+
+	compKeys := maps.Keys(translationComponents)
+	sort.Strings(compKeys)
```
rowanseymour, line 183: I think we want to...

1. Return components in the same order we're given them
2. Return the same number of components as we're given, even if there are no params

norkans7: I think `translation.Components()` is still the components as a dict so I think we'll use the same order when we use the components as a list

rowanseymour: Yeah so then this PR needs to include the switch to components as a list...

norkans7: ok

rowanseymour: And I'll hold off making any mailroom changes

assets/static/template.go
```diff
@@ -73,9 +73,9 @@ func (t *TemplateTranslation) Locale() i18n.Locale { return t.Locale_ }
 func (t *TemplateTranslation) Channel() *assets.ChannelReference { return t.Channel_ }
 
 type TemplateComponent struct {
-	Type_    string           `json:"type"`
 	Content_ string           `json:"content"`
 	Display_ string           `json:"display"`
+	Type_    string           `json:"type"`
```
rowanseymour, line 78: ?

norkans7: That was from merging  the main branch I had to fix the conflicts

flows/actions/send_msg.go
```diff
@@ -156,25 +157,45 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
+	buttonIndex := -1
+	for _, comp := range translationComponents {
 		compParams := comp.Params()
+		compType := comp.Type()
+
+		key := compType
+		if strings.HasPrefix(compType, "button/") {
+			buttonIndex += 1
+			key = fmt.Sprintf("button.%d", buttonIndex)
+		}
```
rowanseymour, line 174: I think we're still missing something... name on components https://github.com/nyaruka/rapidpro/pull/5117

So now editor take the list of actual components, and return a list of uuid+name+params components.. and here we can use `name` to match them.

rowanseymour: The alternative is that template syncing has maintain UUIDs for existing components.. but that's complicated

norkans7: component UUIDs are managed by the flow editor, I do not think the syncing is touching any of that

rowanseymour: yeah what I meant is we could have RP set them but it would have to be more complex syncing that doesn't just generate new UUIDs every time.

rowanseymour (changes requested):

rowanseymour (approved):

flows/msg.go
```diff
@@ -173,16 +173,22 @@ type TemplateParam struct {
 	Value string `json:"value"`
 }
 
+type TemplateComponent struct {
```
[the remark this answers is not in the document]
norkans7: And `TemplatingParam` too?

PR #1234, opened 2024-03-20
Fix and improve generation of template msg previews

rowanseymour:


flows/msg.go
```diff
@@ -178,16 +181,56 @@ type TemplatingComponent struct {
...
+	// replace {{?}} placeholders in component display using any remaining param values
+	for i := 0; i < numDisplayParams; i++ {
+		value := ""
+		if (numContentParams + i) < len(tc.Params) {
+			value = tc.Params[numContentParams+i].Value
+		}
+		display = strings.ReplaceAll(display, fmt.Sprintf("{{%d}}", i+1), value)
+	}
+
+	return content, display
```
rowanseymour, line 221: @norkans7 @ericnewcomer the complexity of this has me wondering if it's a good idea to put variables for the content part of a component and variables for the display part in the same list... because it probably means both courier and the editor would need the same logic to know which variables belong to which.

According to META's woeful documentation, for a URL type button, it can have a variable in the URL and a variable in the text:

Captura de pantalla 2024-03-20 a la(s) 14 18 35

But I can't find any examples of how to pass variables for both parts...

rowanseymour: One alternative would be to add `display_params` as separate field to components

norkans7: So far they can not have the variable at the same time, we have display for URL type and in that case the display cannot have variables, the variable will be in the URL only

in other words, a display cannot have variables, if we have the quick_reply type the variable will in the content itself we won't be having any display for that type

norkans7 (approved):

flows/msg.go
rowanseymour: Or just not support URL buttons with variables in the display part... I hate making this so complicated for a thing nobody might use.

norkans7: Meta already do not support that now, we can add that support when they add that too.
buttons can have maximum one variable if it is the URL type the variable is for the URL part only.

rowanseymour: Ugh so their docs are just wrong.. guess I'll strip out that complexity then

PR #1238, opened 2024-04-04
Add template translation external ID to msg templating

norkans7:
Contains #1240 and #1239

### DIFF
--- a/flows/msg.go
+++ b/flows/msg.go
@@ -174,6 +174,7 @@ func (m *MsgOut) UnsendableReason() UnsendableReason { return m.UnsendableReason
 type TemplatingParam struct {
 	Type  string `json:"type"`
 	Value string `json:"value"`
+	Name  string `json:"name"`
 }
 
 type TemplatingComponent struct {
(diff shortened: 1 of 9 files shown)


flows/msg.go
```diff
@@ -199,13 +201,14 @@ func (tc *TemplatingComponent) Preview(c assets.TemplateComponent) string {
 // MsgTemplating represents any substituted message template that should be applied when sending this message
 type MsgTemplating struct {
 	Template_   *assets.TemplateReference `json:"template"`
+	ExternalID_ string                    `json:"external_id"`
```
norkans7, line 204: We need this as for Twilio content template we need to send the template SID to the API and not the template name as we do not for WhatsApp API

rowanseymour: why can't mailroom just resolve that?

norkans7: Okay, I can move that to mailroom

norkans7:
replaced by https://github.com/nyaruka/mailroom/pull/226

PR #1239, opened 2024-04-05
Fix template param to have the correct ordering

norkans7:


flows/actions/send_msg.go
```diff
@@ -167,12 +168,16 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
 			} else {
-				params[i] = flows.TemplatingParam{Type: p.Type(), Value: ""}
+				params[i] = flows.TemplatingParam{Type: p.Type(), Name: p.Name(), Value: ""}
 			}
 		}
 
+		sort.Slice(params, func(i, j int) bool {
+			return params[i].Name < params[j].Name
+		})
+
```
norkans7, line 180: Fix current issue with parameters sent out of order

And the name field will also help us build the `ContentVariables ` we 'll send for Twilio content templates in addition to the template SID stored in the external ID addressed in #1238

PR #1240, opened 2024-04-05
Fail msg with template status not approved

norkans7:


assets/template.go
```diff
@@ -55,10 +55,22 @@ type TemplateComponent interface {
...
+// TemplateStatus is the status for the template translation
+type TemplateStatus string
+
+// different status that template translations can have
+const (
+	TemplateStatusApproved    TemplateStatus = "A"
+	TemplateStatusPending     TemplateStatus = "P"
+	TemplateStatusRejected    TemplateStatus = "R"
+	TemplateStatusUnsupported TemplateStatus = "X"
+)
```
rowanseymour, line 67: We don't use single char codes in the engine... 

Maybe this could be just `Approved() bool` since that's all that matters to the engine

norkans7: Okay I can change that

rowanseymour (approved):

PR #1241, opened 2024-04-12
Rework template assets to use variables instead of component specific param lists

rowanseymour:


flows/actions/testdata/_assets.json
```diff
@@ -333,34 +342,35 @@
                             "type": "header",
                             "name": "header",
                             "content": "",
-                            "params": [
-                                {
-                                    "type": "image"
-                                }
-                            ]
```
rowanseymour, line 340: stripping this out because we don't support anything besides text

flows/actions/send_msg.go
```diff
@@ -161,15 +162,19 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
 		paramValues := evaluatedParams[comp.Name()]
-		params := make([]flows.TemplatingParam, len(comp.Params()))
+		params := make([]flows.TemplatingParam, len(varMap))
 
-		for i, p := range comp.Params() {
+		for i, varName := range utils.SortedKeys(varMap) {
+			v := variables[varMap[varName]]
 			if i < len(paramValues) {
-				params[i] = flows.TemplatingParam{Type: p.Type(), Value: paramValues[i]}
+				params[i] = flows.TemplatingParam{Type: v.Type(), Value: paramValues[i]}
```
norkans7, line 175: Any reason to not add `varName` in the `TemplatingParam`?
That is the new data we need to be able to support the Twilio templates which triggered this refactoring.
I guess that could be another PR too if this is just for backward compatibility

PR #1246, opened 2024-04-17
13.5

rowanseymour:


flows/actions/base_test.go
```diff
@@ -140,16 +140,16 @@ func testActionType(t *testing.T, assetsJSON json.RawMessage, typeName string) {
 		// inject the action into a suitable node's actions in that flow
 		actionsPath := []string{"flows", fmt.Sprintf("[%d]", flowIndex), "nodes", "[0]", "actions"}
 		actionsJson := []byte(fmt.Sprintf("[%s]", string(tc.Action)))
-		assetsJSON = test.JSONReplace(assetsJSON, actionsPath, actionsJson)
+		testAssetsJSON := test.JSONReplace(assetsJSON, actionsPath, actionsJson)
```
rowanseymour, line 143: not declaring a new variable here meant that changes to the assets was leaking between tests

flows/definition/migrations/testdata/migrations/13.5.0.json
```diff
@@ -0,0 +1,225 @@
...
+                        ]
+                    },
+                    "5a5ccef6-8daa-4cb8-8e52-6b486361710f": {
+                        "params": [
+                            "Si"
+                        ]
+                    }
+                },
+                "kin": {
+                    "5a5ccef6-8daa-4cb8-8e52-6b486361710f": {
```
rowanseymour, line 110: here we have a translation for the second component but not the first - the final `template_variables` translation needs to borrow values for the first component from the original language

flows/actions/send_msg.go
```diff
@@ -32,19 +32,11 @@ const TypeSendMsg string = "send_msg"
 //	  "text": "Hi @contact.name, are you ready to complete today's survey?",
 //	  "attachments": [],
 //	  "all_urns": false,
-//	  "templating": {
```
rowanseymour, line 35: we could just go back to the 13.3 format of a `templating` object with a UUID and a list of variables.. but I never liked having an arbitrary UUID on a thing just so it could be localized... so I'm putting `template` and `template_variables` directly on the action object.

norkans7 (approved):

PR #1247, opened 2024-05-03
Validate the send_msg template only when not empty

norkans7:
Fix https://textit.sentry.io/issues/5293847588/

### DIFF
--- a/flows/actions/send_msg.go
+++ b/flows/actions/send_msg.go
@@ -47,7 +47,7 @@ type SendMsgAction struct {
 	createMsgAction
 
 	AllURNs           bool                      `json:"all_urns,omitempty"`
-	Template          *assets.TemplateReference `json:"template,omitempty"`
+	Template          *assets.TemplateReference `json:"template,omitempty" validate:"omitempty"`
 	TemplateVariables []string                  `json:"template_variables,omitempty" engine:"localized,evaluated"`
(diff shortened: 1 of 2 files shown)


flows/actions/testdata/send_msg.json
```diff
@@ -21,6 +21,31 @@
         },
         "read_error": "field 'topic' is not a valid message topic"
     },
+    {
```
norkans7, line 24: this tests are just to make sure the validation works but no easy way to test the validation with omitempty

flows/actions/send_msg.go
```diff
@@ -47,7 +47,7 @@ type SendMsgAction struct {
 	createMsgAction
 
 	AllURNs           bool                      `json:"all_urns,omitempty"`
-	Template          *assets.TemplateReference `json:"template,omitempty"`
+	Template          *assets.TemplateReference `json:"template,omitempty" validate:"omitempty"`
```
rowanseymour, line 50: if this was necessary... every single `send_msg` action without a template would be failing .. no?

norkans7: Yes, I think those that are failing are those with a previous version set before they migrate to the latest version

rowanseymour: And why are we only seeing this problem for simulation?

norkans7: mailroom will use the flow definition passed from the simulation request and that is the definition before the migrations I think

PR #1251, opened 2024-05-14
Tweak contactql grammar to enforce stricter limits on condition left-hand sides

rowanseymour:


antlr/ContactQL.g4
```diff
@@ -20,6 +21,8 @@ COMPARATOR: (
 		| HAS
 		| IS
 	);
+STRING: '"' (~["] | '\\"')* '"';
+NAME: (UnicodeLetter | UnicodeDigit | '_' | ':')+;
```
rowanseymour, line 25: So that we can support something like `field:num_goats`

PR #1252, opened 2024-05-14
Contact queries with explicit property types part 1

rowanseymour:
This adds support for parsing property conditions like `fields.age > 34` and `urns.tel != ""`.

Note that it doesn't normalize queries to look like that - that comes in separate push otherwise there's risk of one mailroom instance creating queries the other can't parse.


antlr/ContactQL.g4
```diff
@@ -5,6 +5,8 @@ import LexUnicode;
 // Lexer rules
 fragment HAS: [Hh][Aa][Ss];
 fragment IS: [Ii][Ss];
+fragment PROPTYPE: (UnicodeLetter)+;
+fragment PROPKEY: (UnicodeLetter | UnicodeDigit | '_')+;
```
rowanseymour, line 9: you could make this stuff stricter but it's actually easier to deal with errors higher up in the parser rather in the lexer

contactql/error.go
```diff
@@ -21,6 +21,7 @@ const (
 	ErrUnsupportedContains   = "unsupported_contains"   // `property` the property key
 	ErrUnsupportedComparison = "unsupported_comparison" // `property` the property key, `operator` one of =>, <, >=, <=
 	ErrUnsupportedSetCheck   = "unsupported_setcheck"   // `property` the property key, `operator` one of =, !=
+	ErrUnknownPropertyType   = "unknown_property_type"  // `type` the property type
```
rowanseymour, line 24: need to add support for this on RP side where we turn these into readable error messages

contactql/parser.go
```diff
@@ -185,27 +185,31 @@ func (c *Condition) validate(env envs.Environment, resolver Resolver) error {
...
-			if val != "active" && val != "blocked" && val != "stopped" && val != "archived" {
-				return NewQueryError(ErrInvalidStatus, "'%s' is not a valid contact status", c.value).withExtra("value", c.value)
-			}
-		} else if c.propKey == AttributeLanguage {
-			if c.value != "" {
-				_, err := i18n.ParseLanguage(c.value)
-				if err != nil {
-					return NewQueryError(ErrInvalidLanguage, "'%s' is not a valid language code", c.value).withExtra("value", c.value)
+		// for some text attributes, do some additional validation
+		if c.propType == PropertyTypeAttribute {
```
rowanseymour, line 191: little bit of reorg here so that if you have a field called `language`.. we don't enforce that the value is valid language code etc

PR #1254, opened 2024-05-16
Switch to use goreleaser and update actions to latest versions

norkans7:


goreleaser.yml
```diff
@@ -0,0 +1,11 @@
+builds:
+- skip: true
+
+release:
+  extra_files: 
+    - glob: ./docs.tar.gz
```
norkans7, line 6: I think this will upload the docs as that was being done previously.
We need to test that

.github/workflows/ci.yml
```diff
@@ -51,23 +51,18 @@ jobs:
...
-          draft: false
-          prerelease: false
-
-      - name: Upload docs asset
-        uses: actions/upload-release-asset@v1
+          args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish non-main release
```
rowanseymour, line 62: we don't need this... this is when you need to make a binary for a branch of something.. there's no goflow binary

rowanseymour (approved):

PR #1262, opened 2024-06-05
Include header images from template messages as attachments on preview message

rowanseymour:


flows/actions/testdata/send_msg.json
```diff
@@ -917,5 +912,102 @@
...
+            "text": "The Maine Coone is the only native American long haired breed.",
+            "attachments": [
+                "image/jpeg:http://example.com/cat1.jpg"
+            ],
+            "template": {
+                "uuid": "be68beff-1a5b-424b-815e-023cc53c1ddc",
+                "name": "cat_fact"
+            },
+            "template_variables": [
+                "http://example.com/cat2.jpg",
```
rowanseymour, line 930: @norkans7 @ericnewcomer wondering if this shouldn't be `"image/jpeg:http://example.com/cat2.jpg"` so that these image variables are more consistent with how we do msg media in general.. and then the onus is on courier to transform it like it would a regular attachment.

flows/actions/send_msg.go
```diff
@@ -150,27 +151,32 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
 	var previewQRs []string
 
 	for _, comp := range translation.Components() {
 		previewContent := comp.Content()
 		for key, index := range comp.Variables() {
-			previewContent = strings.ReplaceAll(previewContent, fmt.Sprintf("{{%s}}", key), variables[index].Value)
+			variable := variables[index]
+			if variable.Type == "text" {
+				previewContent = strings.ReplaceAll(previewContent, fmt.Sprintf("{{%s}}", key), variable.Value)
+			} else if variable.Type == "image" && variable.Value != "" {
```
norkans7, line 164: What about videos or documents? which are the other type that can be supported

rowanseymour: Yeah good point we should make this work for all header types since we're doing this work

flows/actions/testdata/send_msg.json
norkans7: I adjusted courier side for that in https://github.com/nyaruka/courier/pull/752

rowanseymour: Ok I've updated this PR to assume that variable values look like attachments

flows/actions/send_msg.go
```diff
@@ -127,14 +128,16 @@ func (a *SendMsgAction) getTemplateMsg(run flows.Run, urn urns.URN, channelRef *
...
+		// we pad out any missing variables with empty values
+		value := ""
 		if i < len(evaluatedVariables) {
-			variables[i] = &flows.TemplatingVariable{Type: v.Type(), Value: evaluatedVariables[i]}
-		} else {
-			variables[i] = &flows.TemplatingVariable{Type: v.Type(), Value: ""}
+			value = evaluatedVariables[i]
 		}
+
+		variables[i] = &flows.TemplatingVariable{Type: v.Type(), Value: value}
```
rowanseymour, line 140: I thought about having this ignore attachments that don't match the header type, e.g. ignore `video/mp4:http://...` if variable is type `image`... but the empty values will cause a send failure anyway and I think it will be more obvious to the user what the problem is if we just _try_ to send the wrong attachment type and let WhatsApp complain about it.

norkans7: I avoided the mismatch by using the attachment type in https://github.com/nyaruka/courier/pull/752/files should I keep using the parameter type instead?
Of course both approaches would fail to send if the parameter type is not the same as the one on the template

rowanseymour: I think your change is fine.. in the case that someone updates their template but doesn't update the flow using it, we'll send a param of the wrong type to whatsapp, and I assume they'll send back a fairly clear error message.

norkans7 (approved):

PR #1264, opened 2024-06-10
Update DT One requests

norkans7:


services/airtime/dtone/client.go
```diff
@@ -142,8 +148,8 @@ type Transaction struct {
 	} `json:"status"`
 }
 
-// TransactionSync see https://dvs-api-doc.dtone.com/#tag/Transactions
-func (c *Client) TransactionSync(externalID string, productID int, mobileNumber string) (*Transaction, *httpx.Trace, error) {
+// TransactionAsync see https://dvs-api-doc.dtone.com/#tag/Transactions
```
rowanseymour, line 151: This switch is a problem no? How do we know if the transaction was successful?

services/airtime/dtone/client.go
```diff
@@ -74,10 +74,16 @@ type Operator struct {
...
 	var response []*Operator
 
-	trace, err := c.request("GET", fmt.Sprintf("lookup/mobile-number/%s", tel), nil, &response)
+	payload := &struct {
+		MobileNumber string `json:"mobile_number"`
+	}{
+		MobileNumber: phoneNumber,
+	}
+
+	trace, err := c.request("POST", "lookup/mobile-number", payload, &response)
```
norkans7, line 86: GET method was depracated https://dvs-api-doc.dtone.com/#tag/Mobile-Number

services/airtime/dtone/client.go
```diff
@@ -164,7 +170,7 @@ func (c *Client) TransactionSync(externalID string, productID int, mobileNumber
 		},
 	}
 
-	trace, err := c.request("POST", "sync/transactions", payload, &response)
+	trace, err := c.request("POST", "async/transactions", payload, &response)
```
norkans7, line 173: synchronous approach was deprecated https://dvs-api-doc.dtone.com/#tag/Transactions

rowanseymour: I can see that but isn't this a problem? Doesn't this require that we poll it to know if it actually succeeded?

norkans7: I think we get 201 HTTP status for each transaction we send to DT One and it will send that in an async task

we could add the callback to actually update the transfer in case that is not successful

I understand that slight changes what success mean in the flow but most user will prefer having that instead of not being able to use that at all I guess.
so in the flow success mean we send the requests to DT One and we hade a transaction ID back successfully

rowanseymour: Since the sync version is only deprecated, can't we just keep using it for now until we figure out how a flow can do async transfers? 

I think to do it correctly, transfer_airtime actions would have to become waits, and then have mailroom listen for success or failure, and resume the flow session accordingly.

rowanseymour (approved):

PR #1300, opened 2024-12-26
Allow categories up to 64 chars, disallow empty categories

rowanseymour:


flows/routers/base.go
```diff
@@ -195,7 +195,7 @@ type baseRouterEnvelope struct {
 	Type       string            `json:"type"                  validate:"required"`
 	Wait       json.RawMessage   `json:"wait,omitempty"`
 	ResultName string            `json:"result_name,omitempty" validate:"omitempty,result_name"`
-	Categories []json.RawMessage `json:"categories,omitempty"  validate:"required,min=1,dive,result_category"`
```
rowanseymour, line 198: this is wrong.. and not sure how it's currently working since each category at this point is just `json.RawMessage`

PR #1301, opened 2025-01-08
Enforce a limit of 50 URNs per contact

rowanseymour:


flows/modifiers/urns.go
```diff
@@ -60,9 +60,16 @@ func (m *URNsModifier) Apply(eng flows.Engine, env envs.Environment, sa flows.Se
 			log(events.NewErrorf("'%s' is not valid URN", urn))
 		} else {
 			if m.Modification == URNsAppend || m.Modification == URNsSet {
-				modified = contact.AddURN(urn, nil)
+				if len(contact.URNs()) >= flows.MaxContactURNs {
```
norkans7, line 63: I guess the error message says we allow 50 so we use `> flows.MaxContactURNs` and not `>= flows.MaxContactURNs`

rowanseymour: this is checking if the modifier should be allowed to add one more URN tho

norkans7: Right, I think the unit test added should not be an error as that has exactly 50 urns, we need one more urn to have that as error and use `>` comparison

rowanseymour: the unit test is adding 50 urns but the contact already has 1

norkans7: ahh right

norkans7 (approved):

PR #1324, opened 2025-04-02
Add `call_llm` action with corresponding `llm_called` event and bump flow spec version.

rowanseymour:


flows/actions/testdata/call_llm.json
```diff
@@ -0,0 +1,108 @@
...
+            }
+        ],
+        "templates": [
+            "Categorize the following text as positive or negative",
+            "@input.text"
+        ],
+        "locals": {
+            "_llm": {
+                "status": "success",
+                "output": "You asked:\n\nCategorize the following text as positive or negative\n\nHi everybody"
```
rowanseymour, line 92: Open to suggestions here...

1. `_llm` because this is a system defined local variable and my thinking is that prevents overwriting a user defined local. You could specify the variable name on the action.. but maybe that's overcomplicated for MVP.
2. split router can have rules that match on `.output` or a `Failure` category that matches on `.status`. Maybe it could just match empty output?

flows/definition/flow.go
```diff
@@ -28,7 +28,7 @@ const (
 )
 
 // CurrentSpecVersion is the flow spec version supported by this library
-var CurrentSpecVersion = semver.MustParse("13.6.1")
+var CurrentSpecVersion = semver.MustParse("13.7.0")
```
rowanseymour, line 31: no migration required but we are technically introducing a new action so flows with that aren't backward compatible

PR #1326, opened 2025-04-03
Slice

rowanseymour:


excellent/types/array.go
```diff
@@ -44,6 +44,11 @@ func (x *XArray) Get(index int) XValue {
 	return x.values()[index]
 }
 
+// Get is called when this object is indexed
```
norkans7, line 47: update comment

excellent/functions/builtin.go
```diff
@@ -1518,23 +1519,55 @@ func TimeFromParts(env envs.Environment, hour, minute, second int) types.XValue
...
+	start, xerr := types.ToInteger(env, args[0])
+	if xerr != nil {
+		return xerr
+	}
+	if start < 0 {
+		return types.NewXErrorf("must start with a positive index")
+	}
+
+	end := array.Count()
+	if len(args) == 2 {
```
norkans7, line 1544: what if we have more than 2 items in args, I think we allow this to work with the second item and ignore the others

so I guess we use `len(args) >= 2`

excellent/types/array.go
norkans7: nvm that was removed later

excellent/functions/builtin.go
rowanseymour: The function is wrapped with `InitialArrayFunction` which takes care of enforcing the number of allowed args

PR #1335, opened 2025-04-09
Allow users to manage the whatsapp channels by using the preferred ch…

norkans7:
…annel action

### DIFF
--- a/flows/contact.go
+++ b/flows/contact.go
@@ -46,6 +46,9 @@ const (
 	MaxContactURNs = 50
 )
 
+// schemes of URNs that aren't tied to a specific channel
+var portableURNSchemes = map[string]bool{urns.Phone.Prefix: true, urns.WhatsApp.Prefix: true}
+
 // Contact represents a person who is interacting with the flow
 type Contact struct {
 	uuid       ContactUUID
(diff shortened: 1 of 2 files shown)


flows/contact.go
```diff
@@ -434,8 +434,8 @@ func (c *Contact) UpdatePreferredChannel(channel *Channel) bool {
 		otherURNs := make([]*ContactURN, 0)
 
 		for _, urn := range c.urns {
-			// tel URNs can be re-assigned, other URN schemes are considered channel specific
-			if urn.URN().Scheme() == urns.Phone.Prefix && channel.SupportsScheme(urns.Phone.Prefix) {
+			// tel or whatsapp URNs can be re-assigned, other URN schemes are considered channel specific
+			if (urn.URN().Scheme() == urns.Phone.Prefix && channel.SupportsScheme(urns.Phone.Prefix)) || (urn.URN().Scheme() == urns.WhatsApp.Prefix && channel.SupportsScheme(urns.WhatsApp.Prefix)) {
```
rowanseymour, line 438: let's do this via a map..

```golang
// schemes of URNs that aren't tied to a specific channel
var portableURNSchemes := map[string]bool{urns.Phone.Prefix: true, urns.WhatsApp.Prefix: true}

rowanseymour (changes requested):

PR #1336, opened 2025-04-09
Refactor webhook services

rowanseymour:
* clear `@webhook` in case of error.. allows webhook splits with no result
* let `@webhook` be more easily restored from a previously made call


flows/actions/testdata/call_resthook.json
```diff
@@ -118,8 +118,7 @@
                 "elapsed_ms": 0,
                 "retries": 0,
                 "status": "success",
-                "resthook": "new-registration",
-                "extraction": "valid"
```
rowanseymour, line 122: added this years ago but has never been used for anything

PR #1346, opened 2025-04-16
Make result optional on open ticket actions

rowanseymour:
but always write a local that can be used for splitting

### DIFF
--- a/flows/actions/open_ticket.go
+++ b/flows/actions/open_ticket.go
@@ -27,8 +27,7 @@ const TypeOpenTicket string = "open_ticket"
 //	    "name": "Weather"
 //	  },
 //	  "body": "@input",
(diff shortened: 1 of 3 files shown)


flows/actions/open_ticket.go
```diff
@@ -75,9 +74,17 @@ func (a *OpenTicketAction) Execute(ctx context.Context, run flows.Run, step flow
 
 	ticket := a.open(run, topic, assignee, evaluatedNote, logModifier, logEvent)
 	if ticket != nil {
-		a.saveResult(run, step, a.ResultName, string(ticket.UUID()), CategorySuccess, "", "", nil, logEvent)
+		run.Locals().Set("_new_ticket", string(ticket.UUID()))
```
rowanseymour, line 77: @ericnewcomer @norkans7 still not sure what the naming scheme should be for locals that users don't define.. maybe not too worried here because I doubt anyone is using this value

PR #1348, opened 2025-04-21
Simplify extracting dependencies during inspection

rowanseymour:
Less magic


flows/actions/testdata/send_broadcast.json
```diff
@@ -217,11 +217,6 @@
         ],
         "inspection": {
             "dependencies": [
-                {
-                    "key": "gender",
-                    "name": "",
-                    "type": "field"
-                },
```
rowanseymour, line 224: because dependency extraction from templates now happens after regular dependency extraction

PR #1481, opened 2026-01-29
Add prompt field, to use for location request if supported

norkans7:


flows/msg.go
```diff
@@ -69,17 +73,30 @@ func NewMsgIn(urn urns.URN, channel *assets.ChannelReference, text string, attac
 
 // NewMsgOut creates a new outgoing message
 func NewMsgOut(urn urns.URN, channel *assets.ChannelReference, content *MsgContent, templating *MsgTemplating, locale i18n.Locale, reason UnsendableReason) *MsgOut {
+	prompt := ""
+	quickReplies := []QuickReply{}
+	if content != nil && len(content.QuickReplies) > 0 {
```
rowanseymour: Don't think `content` can ever be null

PR #1537, opened 2026-07-17
Add template caching tied to session assets

rowanseymour:
Templates are now parsed once and cached, keyed by their source, with the cache living alongside session assets so all sessions using the same assets share parsed templates. Parse-once evaluation is ~8.5x faster than parsing per evaluation (#1533).

- `excellent.TemplateCache` - a thread-safe content-addressed cache: `Get(src)` parses on miss. Because a parsed template is a pure function of its source and immutable, entries never go stale and no invalidation/TTL/eviction is needed - the cache's lifetime is bound to that of the content it came from.
- `flows.Cache` - a container for such caches, exposed as `SessionAssets.Cache()`. For now it only holds `Templates()` but is a natural home for other parsed-thing caches later (e.g. contactql queries, which are currently re-parsed on every session assets construction).
- `engine.NewSessionAssets` takes an optional `*flows.Cache` - nil creates a private one; callers that rebuild session assets against unchanged content can pass a shared cache to extend its lifetime.
- Runs now evaluate templates through the cache. `excellent.Template` gains `EvaluateValue` (the typed counterpart of `Evaluate`, mirroring `Evaluator.TemplateValue`) to cover the `EvaluateTemplateValue` path - note that path keys the cache on the trimmed source since `TemplateValue` trims before evaluating.

Callers evaluating templates outside of flow runs (e.g. broadcast message rendering) can use `sa.Cache().Templates().Get(text).Evaluate(...)` instead of the engine's evaluator to get the same benefit.

Testing: the existing differential tests and fuzz target now also verify `EvaluateValue` against `Evaluator.TemplateValue` across the adversarial corpus and all runner fixture strings (fuzzed 186k execs clean); `TemplateCache` has a race-tested concurrency test; the full suite passes with all golden runner fixtures unchanged, confirming the rewired run evaluation is behaviorally identical.

### DIFF
--- a/cmd/flowrunner/main.go
+++ b/cmd/flowrunner/main.go
@@ -117,7 +117,7 @@ func RunFlow(eng flows.Engine, assetsPath string, flowUUID assets.FlowUUID, init
 		return nil, err
 	}
 
-	sa, err := engine.NewSessionAssets(envs.NewBuilder().Build(), source, nil)
+	sa, err := engine.NewSessionAssets(envs.NewBuilder().Build(), source, nil, nil)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing assets: %w", err)
 	}
--- a/core/call_test.go
+++ b/core/call_test.go
@@ -34,7 +34,7 @@ func TestCall(t *testing.T) {
 	}`))
 	require.NoError(t, err)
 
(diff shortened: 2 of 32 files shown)


rowanseymour:
Punting on this for now — design context and options captured in #1563 for future re-evaluation.

PR #1556, opened 2026-07-21
Namespace event error codes and add expression:too_complex

rowanseymour:
Adopts the `domain:subtype` convention already used everywhere except goflow's event codes, and adds a distinct code for budget-exceeded errors.

## Why

The mailroom web API builds error codes as `domain:code` — `urn:taken`, `query:unknown_property`, `flow:invalid`, `ai:unknown` — and temba parses them by splitting on `:`. mailroom's series/metric codes are colon-namespaced too (`text:in`, `voice:out`, `tokens:in`, …). goflow's flat `dependency_missing` / `urn_taken` / `expression` were the outliers, and `urn_taken` was actually **inconsistent** with the `urn:taken` the mailroom API already emits for the same concept.

## Changes

- `dependency_missing` → `dependency:missing`
- `urn_taken` → `urn:taken`
- adds `expression:too_complex`, emitted when an evaluation exceeds its cost budget, so it's distinguishable from ordinary expression errors (a consumer can route only these to Sentry — hitting the budget is supposed to be impossible)

To carry the distinct code through cleanly, `TemplateError` now holds the underlying `error` instead of a pre-flattened string, so `errorToEvents` recognises the fatal `ErrTooComplex` sentinel via `errors.Is` rather than string-matching the message.

## Consumer impact

- `dependency_missing` and `expression` are used nowhere in mailroom/courier/temba/floweditor.
- `urn_taken` is used by mailroom ([event.go](https://github.com/nyaruka/mailroom), [contacts.go]) **via the constant** `events.ErrorCodeURNTaken`, in real time (comparing freshly-produced events). Changing the constant's value is safe under the normal lockstep bump — the comparison moves on both sides together. **Requires a mailroom bump.**
- No consumer queries stored event JSON by these code strings.

Fixtures/snapshots updated; docgen verified in the devcontainer.

### DIFF
--- a/core/events/error.go
+++ b/core/events/error.go
@@ -14,9 +14,10 @@ func init() {
 const TypeError string = "error"
 
 const (
-	ErrorCodeDependencyMissing = "dependency_missing"
-	ErrorCodeURNTaken          = "urn_taken"
-	ErrorCodeExpression        = "expression"
+	ErrorCodeDependencyMissing    = "dependency:missing"
+	ErrorCodeURNTaken             = "urn:taken"
+	ErrorCodeExpression           = "expression"
+	ErrorCodeExpressionTooComplex = "expression:too_complex"
 )
 
 // Error events are created when an error occurs during flow execution.
--- a/core/events/testdata/TestEventMarshaling_error_dependency.snap
+++ b/core/events/testdata/TestEventMarshaling_error_dependency.snap
@@ -3,5 +3,5 @@
     "type": "error",
     "created_on": "2025-05-04T12:30:46.123456789Z",
     "text": "Missing dependency: field[key=age,name=Age]",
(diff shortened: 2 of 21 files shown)


rowanseymour:
Addressed the robustness point (4f9ea9cf): added `XError.Is` comparing the underlying native error, so `errors.Is(err, ErrTooComplex)` still recognizes the sentinel if it's ever re-wrapped in a new XError — the invariant is now explicit rather than resting on pointer identity, with a regression test covering the re-wrap case.

On the other two: keeping `expression` bare is intentional (it's the catch-all; flow-event codes don't go through the `:`-split that the mailroom API codes do), and the `TemplateError.Message`→`.Err` field change + code-string renames will be captured in the CHANGELOG when the release is cut (it's written by the release process, not by hand).

PR #1572, opened 2026-07-22
Return error when reading location hierarchy with more than 4 levels

norkans7:
## Problem

Location hierarchies support 4 levels, but `locationFromEnvelope` recursed to unbounded depth while the name-lookup tables were hardcoded to 4 levels with no bounds check on the build path (`addNameLookups`). A hierarchy JSON nested 5+ levels deep passed validation and then panicked with `index out of range [4] with length 4` inside `NewLocationHierarchy`. (`FindByName` had the level guard; the build path did not.)

## Solution

Added a `maxLocationLevels` constant (replacing the hardcoded 4s) and a depth check in both build paths (`ReadLocationHierarchy` and `UnmarshalJSON`): hierarchies deeper than 4 levels now return a clear error instead of panicking. Test added with a 5-level hierarchy asserting the error.

### DIFF
--- a/envs/location_test.go
+++ b/envs/location_test.go
@@ -112,3 +112,37 @@ func TestLocationHierarchy(t *testing.T) {
 	assert.Equal(t, gasabo, hierarchy.FindByPath("rwanda > kigali city > gasabo"))
 	assert.Equal(t, ndera, hierarchy.FindByPath("rwanda > kigali city > gasabo > ndera"))
 }
+
+func TestLocationHierarchyTooDeep(t *testing.T) {
+	env := envs.NewBuilder().Build()
+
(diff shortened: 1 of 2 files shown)


rowanseymour:
Closing after review: the out-of-range panic is real, but a location hierarchy deeper than 4 levels isn't data the engine supports — if it ever shows up, that's a signal that code changes are needed, not a condition to handle gracefully. So rather than validating depth with an extra traversal at read time, we've made the failure an explicit, self-describing panic at the existing bounds check (296907c4), keeping this PR's `maxLocationLevels` constant cleanup. Thanks — the diagnosis here was accurate and useful.

PR #1573, opened 2026-07-22
Fix 13.6.1 migration to trim trailing underscore from truncated result lookups when that matches an actual result key

norkans7:
## Problem

`Migrate13_6_1` truncated `@results.X` context lookups with a bare `Truncate(old, 64)`, which doesn't match how the migrated result keys are derived (`Snakify(TrimSpace(Truncate(name, 64)))`). When a result name has whitespace at the 64-char truncation boundary, the migrated key loses a trailing underscore that the truncated lookup keeps — so a previously-working `@results.x` reference silently breaks (context lookups are exact, case-insensitive).

A blind "trim trailing underscores" fix would be wrong too: `Snakify` preserves literal underscores, and a name with a real `_` at the boundary legitimately produces a key ending in `_`. The two cases are indistinguishable from the lookup string alone.

## Solution

Since 13.6.1 runs after 13.6.0, the flow's result names are already in their final form, so the migration now collects the flow's actual runtime result keys (from `set_run_result` actions and router `result_name`s) and, after truncating a lookup, trims trailing underscores only when the untrimmed form matches no real key and the trimmed form does. Lookups referencing results not defined in the flow behave exactly as before.

Two end-to-end migration test cases document both sides: a whitespace-boundary name where the lookup must be trimmed to match, and a literal-underscore name where it must not be.

### DIFF
--- a/flows/definition/migrations/13.x.go
+++ b/flows/definition/migrations/13.x.go
@@ -9,6 +9,7 @@ import (
 	"github.com/nyaruka/gocommon/uuids"
 	"github.com/nyaruka/goflow/excellent"
 	"github.com/nyaruka/goflow/excellent/refactor"
+	"github.com/nyaruka/goflow/utils"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


rowanseymour:
Closing after review: the bug is real and the fix is well-constructed — verified that the divergence between 13.6.0's name truncation (TrimSpace + snakify) and this migration's bare lookup truncation can leave a rewritten `@results.x` lookup that no longer resolves. But the trigger requires a result name over 64 chars with whitespace exactly at the truncation boundary, which is vanishingly rare, and changing the output of an already-shipped migration creates divergence between definitions migrated before and after the change — a worse trade than the bug itself at this point.

The limitation is now documented in the migration itself (baf8e8d0) so it's a known, deliberate behavior rather than an undiscovered bug.

PR #1613, opened 2026-08-17
Split restricted webhook domains into separate warn and block lists

rowanseymour:
## Summary

Webhook domain restrictions were a single list that only ever produced a warning event. This splits them into two lists: domains that warn (calls still go through) and domains that are blocked (the call isn't made at all), so domains can be migrated from the first to the second as operators are ready.

## Changes

**`flows/services.go`** — `WebhookService.IsRestricted(u) bool` is replaced by `Restriction(u) URLRestriction`, a new string type with values `URLRestrictionNone`, `URLRestrictionWarn` and `URLRestrictionBlock`.

**`services/webhooks/service.go`** — `NewServiceFactory` and `NewService` take `warnDomains` and `blockDomains` in place of the single `restrictedDomains`. Host matching is unchanged (case-insensitive, matches subdomains), factored into a shared `matchesDomain` helper. Blocking takes precedence when a domain appears in both lists.

**`flows/actions/call_webhook.go`** — a warn-listed URL logs the existing `url:restricted` warning and the call proceeds as before; a block-listed URL logs a new `url:restricted` error and the action returns without making the request, leaving `@webhook` cleared.

**`core/events/error.go`** — adds `ErrorCodeURLRestricted` (`url:restricted`). Error and warning codes are separate namespaces, so blocked and warned calls share the code string while differing in event type.

Test engines now configure `graph.facebook.com` as a warn domain and `api.twilio.com` as a block domain.

## Behavior examples

| URL matches | Event | Call made? | `@webhook` |
|---|---|---|---|
| no list | — | yes | call result |
| warn list | `warning` / `url:restricted` — "Webhook calls to X may be blocked in the future" | yes | call result |
| block list | `error` / `url:restricted` — "Webhook calls to X are not allowed" | no | cleared |

Because a blocked call produces no `webhook_called` event, a router branching on `@webhook.status` sees the same state as a webhook that was never called.

## Test plan

- [x] Full suite passes (`go test -p=1 ./...`)
- [x] `TestRestriction` covers warn, block, case insensitivity, subdomain matching, block-wins-over-warn, and non-matching hosts including the `notgraph.facebook.com.evil.com` suffix trap
- [x] New `call_webhook` action fixture asserts the error event, no HTTP call, and null webhook for a blocked domain; regenerating fixtures with `-update` produces no diff

### DIFF
--- a/cmd/flowrunner/main.go
+++ b/cmd/flowrunner/main.go
@@ -91,7 +91,7 @@ func createEngine() flows.Engine {
 	return engine.NewBuilder().
 		WithHTTPClient(http.DefaultClient).
 		WithWebhookLimits(256*1024, 10000).
-		WithWebhookServiceFactory(webhooks.NewServiceFactory(map[string]string{"User-Agent": "goflow-runner"}, nil)).
+		WithWebhookServiceFactory(webhooks.NewServiceFactory(map[string]string{"User-Agent": "goflow-runner"}, nil, nil)).
 		Build()
 }
 
--- a/core/events/base_test.go
+++ b/core/events/base_test.go
@@ -554,7 +554,7 @@ func TestWebhookCalledEventTrimming(t *testing.T) {
 
 	request, _ := http.NewRequest("GET", "http://temba.io/", strings.NewReader(strings.Repeat("X", 20000)))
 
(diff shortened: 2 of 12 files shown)


rowanseymour:
Closing in favor of https://github.com/nyaruka/goflow/pull/1614

